PR #106, opened 2026-05-16
feat: reconcile workspaces on worker startup

### LINKED ISSUE #68: [M5][P0][spec-alignment] D3 Reconcile in-flight workspaces with tracker on worker startup
## Background

Symphony SPEC.md mandates that restart recovery be tracker-driven and
filesystem-driven:

> "Restart recovery is tracker-driven and filesystem-driven (without a durable orchestrator DB)."

SPEC also defines the orchestrator's authoritative state model:

> "Retry Entry: Scheduled retry state for an issue."
> "Orchestrator Runtime State: Single authoritative in-memory state owned by the orchestrator."
> "The orchestrator is the only component that mutates scheduling state."

And reconciliation timing:

> "Reconciliation runs before dispatch on every tick."

The intended behavior: on startup (and at each tick) the orchestrator reads
tracker state, walks the workspace root, and reconciles the two — stopping
orphaned agents, cleaning workspaces whose issues are now terminal, and
resuming work for issues that are still active.

Today this platform's `docs/symphony-integration.md` lists multi-run
reconciliation under "Not yet implemented". We rely on the queue
(`FOR UPDATE SKIP LOCKED`) to prevent concurrent claims of the same task,
but we do not perform a reconciliation pass on boot.

## Impact

Without reconciliation:

1. **Orphaned workspaces** accumulate on disk after worker crashes — SPEC
   requires terminal-state cleanup; restarts don't trigger it.
2. **Stuck "in flight" rows** in `tasks` may exist after an unclean shutdown,
   either blocking re-dispatch or causing duplicate dispatch depending on
   queue semantics.
3. **Tracker drift**: a Linear issue moved to a terminal state while the
   worker was offline will not have its workspace cleaned up.
4. Differs materially from how SPEC expects the system to recover — a SPEC-
   aligned implementation should be able to lose its database entirely and
   rebuild correct state from tracker + filesystem alone.

## Acceptance criteria

- [ ] On worker startup, before claiming new work, run a reconciliation pass:
  - [ ] Enumerate existing per-issue workspace directories under `workspace.root`.
  - [ ] For each, query the tracker for current issue state.
  - [ ] If terminal: stop any running agent (none should exist post-restart,
        but assert) and clean the workspace per SPEC's terminal-state cleanup.
  - [ ] If unknown / deleted: log a warning and clean the workspace.
  - [ ] If still active: leave the workspace intact and let normal scheduling
        pick it up.
- [ ] Reconciliation is idempotent: running it twice in a row produces the
      same result.
- [ ] Reconciliation emits structured events (`reconcile_start`,
      `reconcile_workspace`, `reconcile_end`) consistent with the existing
      task-event scheme.
- [ ] An integration test simulates: dirty shutdown → restart → reconcile,
      and asserts the expected cleanup.
- [ ] Worker exits with non-zero status if reconciliation itself fails
      (per SPEC: workflow/config errors halt scheduling, do not silently fall
      through).

## Out of scope

- Database schema changes for retry-state persistence. SPEC describes restart
  recovery as filesystem- and tracker-driven (without a durable orchestrator
  DB); if we later choose to add retry-state persistence, that is a separate
  issue.
- Cross-worker leader election. Single-worker assumption is fine for M5.

## References

- SPEC.md: https://github.com/openai/symphony/blob/main/SPEC.md
- Umbrella: #67
- Related: #14 (D2 status transitions — reconciliation needs accurate
  tracker state to be useful)

hey-yvan:
## Summary
- add worker startup reconciliation that fetches active and terminal tracker issues before dispatch
- clean terminal/unknown Linear issue workspaces while preserving active workspaces
- share workspace path sanitization via `workspace.SanitizeComponent` and paginate Linear issue listing

## Test Plan
- `/home/pi/.local/go1.25.10/bin/go test ./internal/worker ./internal/workspace`
- `/home/pi/.local/go1.25.10/bin/go test ./...`
- `/home/pi/.local/go1.25.10/bin/go build ./cmd/trigger-api ./cmd/worker ./cmd/linear-poller`
- `git diff --check`
- diff-only independent review with Claude Code: LGTM

## Notes
- `go test -race -covermode=atomic ./...` is blocked on this ARM/Raspberry Pi runner by ThreadSanitizer (`unsupported VMA range; Found 47 - Supported 48`) before package tests execute.

Closes #68


hey-yvan:
@codex review

hey-yvan:
@codex review

hey-yvan:
@codex review

PR #110, opened 2026-05-16
fix(worker): log startup reconcile workflow source

### LINKED ISSUE #107: Worker startup reconciliation silently skipped when AIOPS_WORKFLOW_PATH is unset
## Summary

The startup workspace reconciliation added in #68 / PR #106 can be silently bypassed in deployments that do not set `AIOPS_WORKFLOW_PATH`. Two unresolved bot review comments on PR #106 flagged this; the underlying logic concern is real and worth fixing on follow-up.

- https://github.com/xrf9268-hue/aiops-platform/pull/106#discussion_r3252781671 (P1)
- https://github.com/xrf9268-hue/aiops-platform/pull/106#discussion_r3252794498 (P2)

> **Scope note:** this issue is the **bug-fix half only** of the PR #106 follow-up. The process-side guardrail that should have caught these unresolved bot threads on close lives in **#108** (Hermes-compatible auto-capture workflow). They were originally one issue; split so the bounded bug fix and the broader recurring-class guardrail can be tracked, prioritized, and shipped independently.

## Reproduction (logic trace)

1. `cmd/worker/main.go` `loadWorkflowForStartupReconcile()` (lines 32-37):
   ```go
   if path := os.Getenv("AIOPS_WORKFLOW_PATH"); path != "" {
       return workflow.Load(path)
   }
   return workflow.LoadOptional("WORKFLOW.md")
   ```
   When the env var is unset, it tries to load `WORKFLOW.md` relative to the worker process CWD.
2. `internal/workflow/loader.go` `LoadOptional` returns schema defaults if the file is missing (no error, no log).
3. `internal/workflow/config.go` `DefaultConfig()` sets `Tracker.Kind = "gitea"`.
4. `cmd/worker/main.go` `run()` (line 50) only invokes `worker.ReconcileStartup` when `wf.Config.Tracker.Kind == "linear"`.

Result: any deployment without `AIOPS_WORKFLOW_PATH` set, and without a `WORKFLOW.md` in the worker process CWD, will silently skip the entire startup reconciliation pass. There is no log line stating that reconciliation was bypassed because the workflow file was not found.

## Scope of impact

- `deploy/docker-compose.yml` sets `AIOPS_WORKFLOW_PATH: /app/examples/WORKFLOW.md` for the worker, so the documented Compose deployment is fine.
- Other deployments — bare-metal, systemd units, hand-rolled k8s manifests, local dev `go run ./cmd/worker` invocations — will hit this trap whenever the env var is forgotten. The `linear-poller` will keep enqueuing work correctly (it takes the workflow path as a CLI arg and `log.Fatal`s if it is missing), but the worker will never reconcile stale workspaces.

The bounded blast radius (Compose unaffected) is why this issue sits at P1 rather than P0; the broader recurring-class concern is tracked at P0 on #108.

## Why this matters for #68's intent

The whole point of the startup reconciliation feature is that operators do not have to manually clean up stale terminal/unknown workspaces after a worker restart. Silently degrading to "do nothing" when the workflow file cannot be found defeats that guarantee in exactly the scenarios where it is most useful (a worker restarted on a fresh host whose env was misconfigured).

## Suggested fixes (any one would address it)

1. **Use `workflow.Resolve(workdir)` instead of `LoadOptional`.** `LoadOptional`'s own doc comment already says: *"Deprecated: use Resolve(workdir) for repo-relative discovery."* Resolve walks the documented `WORKFLOW.md` → `.aiops/WORKFLOW.md` → `.github/WORKFLOW.md` precedence and returns a `Resolution` with `Source` (`file` / `prompt_only` / `default`), which the worker can log so silent fallback becomes observable.
2. **Make the workflow path required for the worker** when reconciliation is wired in — either CLI flag or a hard error if neither `AIOPS_WORKFLOW_PATH` nor a discovered file is present. `cmd/linear-poller` already enforces this (`log.Fatal` on missing arg); the worker should be no laxer.
3. **At minimum, log when `LoadOptional` returns the default fallback** so operators can tell from the worker startup output whether reconciliation was actually configured to run. A one-line `log.Printf("startup reconciliation: workflow file not found at %q, falling back to defaults (tracker.kind=gitea); reconciliation will be skipped", path)` would surface the bug class even without the deeper refactor.

## Acceptance criteria

- [ ] Worker startup emits an unambiguous log line indicating which workflow source was used for the startup reconciliation gate (file path, prompt-only, or default fallback).
- [ ] In a Linear-tracker deployment, accidentally omitting `AIOPS_WORKFLOW_PATH` either (a) is detected and surfaces a loud warning/error, or (b) is resolved automatically via `workflow.Resolve` against the working directory.
- [ ] A regression test asserts that, given a workflow file with `tracker.kind: linear`, `loadWorkflowForStartupReconcile` (or its replacement) returns the linear config rather than the gitea default — and that misconfiguration produces a diagnostic, not silence.

## References

- PR #106 (added the startup reconciliation; merged)
- Issue #68 (parent feature: SPEC startup reconciliation)
- Issue #78 (per-tick reconciliation; separate scope, but the same workflow-loading question applies)
- Issue #108 (process guardrail: auto-capture unresolved bot reviews on PR close — would have caught the threads that produced this issue)

hey-yvan:
## Summary
- resolve startup reconciliation workflow discovery via `workflow.Resolve` when `AIOPS_WORKFLOW_PATH` is unset
- log the startup reconciliation workflow source/path/tracker kind every time
- warn loudly whenever the loaded workflow is not Linear and reconciliation will be skipped

Closes #107

## Test Plan
- [x] `go test ./cmd/worker -run 'TestLoadWorkflowForStartupReconcile(ResolvesCWDWorkflowAndLogsSource|DefaultsWhenNoWorkflowExists|UsesConfiguredWorkflowPath)' -count=1`
- [x] `go test ./cmd/worker -count=1`
- [x] `go test ./...`
- [x] `go build ./cmd/trigger-api ./cmd/worker ./cmd/linear-poller`
- [x] independent diff-only Claude review: LGTM

Note: `go test -race -covermode=atomic ./...` was attempted but this Raspberry Pi host's ThreadSanitizer runtime fails before tests with `unsupported VMA range (Found 47 - Supported 48)`, a host/toolchain limitation rather than a code failure.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -30,9 +30,36 @@ func main() {
 
 func loadWorkflowForStartupReconcile() (*workflow.Workflow, error) {
 	if path := os.Getenv("AIOPS_WORKFLOW_PATH"); path != "" {
-		return workflow.Load(path)
+		wf, err := workflow.Load(path)
+		if err != nil {
+			return nil, err
+		}
+		logStartupReconcileWorkflow(&workflow.Resolution{Source: workflow.SourceFile, Path: path}, wf)
+		return wf, nil
 	}
-	return workflow.LoadOptional("WORKFLOW.md")
+
+	workdir, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+	wf, resolution, err := workflow.Resolve(workdir)
+	if err != nil {
+		return nil, err
+	}
+	logStartupReconcileWorkflow(resolution, wf)
+	return wf, nil
+}
+
+func logStartupReconcileWorkflow(resolution *workflow.Resolution, wf *workflow.Workflow) {
+	if resolution.Source == workflow.SourceDefault {
+		log.Printf("startup reconciliation: workflow source=%s tracker.kind=%s; reconciliation will be skipped unless tracker.kind is linear", resolution.Source, wf.Config.Tracker.Kind)
+		return
+	}
+	if wf.Config.Tracker.Kind != "linear" {
+		log.Printf("startup reconciliation: workflow source=%s path=%s tracker.kind=%s; reconciliation will be skipped unless tracker.kind is linear", resolution.Source, resolution.Path, wf.Config.Tracker.Kind)
+		return
+	}
+	log.Printf("startup reconciliation: workflow source=%s path=%s tracker.kind=%s", resolution.Source, resolution.Path, wf.Config.Tracker.Kind)
 }
 
 func run() error {
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -34,7 +34,15 @@ func loadWorkflowForStartupReconcile() (*workflow.Workflow, error) {
 		if err != nil {
 			return nil, err
 		}
-		logStartupReconcileWorkflow(&workflow.Resolution{Source: workflow.SourceFile, Path: path}, wf)
+		hasFront, err := workflow.HasFrontMatterAt(path)
+		if err != nil {
+			return nil, err
+		}
+		source := workflow.SourceFile
+		if !hasFront {
+			source = workflow.SourcePromptOnly
+		}
+		logStartupReconcileWorkflow(&workflow.Resolution{Source: source, Path: path}, wf)
 		return wf, nil
 	}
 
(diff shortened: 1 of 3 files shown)


hey-yvan:
@codex review

hey-yvan:
@codex review

PR #111, opened 2026-05-16
chore: capture unresolved merged PR reviews

### LINKED ISSUE #108: Auto-capture unresolved bot review threads at PR-close (Hermes-compatible guardrail)
## Summary

Add a non-blocking guardrail that catches unresolved bot code-review threads on merged PRs and auto-files them as follow-up issues. The goal is to plug the "Hermes autonomously merged past a P1 bot nit" failure mode that produced #107, without breaking the autonomous-merge contract that Hermes depends on.

## Motivation

PR #106 merged with two unresolved review threads from `chatgpt-codex-connector` open on the diff (one P1, one P2). Nothing in this repo's `.github/` caught it. The P1 was a real bug — it became #107 only because a human went back and triaged the PR after the fact. Hermes-merged PRs are the dominant merge path in this repo, so every one of them is exposed to the same gap.

Concrete example evidence on PR #106:
- https://github.com/xrf9268-hue/aiops-platform/pull/106#discussion_r3252781671 (P1, `isResolved=false`, `isOutdated=true`)
- https://github.com/xrf9268-hue/aiops-platform/pull/106#discussion_r3252794498 (P2, `isResolved=false`, `isOutdated=false`)

(Both threads happen to point at the same root cause — `loadWorkflowForStartupReconcile` silently falling back to `LoadOptional` defaults — which is why a single follow-up issue, #107, covers both. The workflow proposed here would not have known that ahead of time; it would have filed one issue per thread, and a human/agent would have collapsed them. That collapsing step is fine — the point is that *something* would have surfaced both before they aged.)

## Priority rationale (why P0)

The bug from PR #106 (split into #107) is bounded — only non-Compose deployments hit it — and lives at P1. This issue is filed P0 because it covers the *recurring class*, not a single instance: until a guardrail exists, **every future Hermes-merged PR** can ship past bot feedback the same way #106 did, and each miss ages quietly in the same way #107's bug did. The compounding blast radius across the merge queue is what drives P0, not the technical difficulty (the implementation is small).

## Constraint: must not break Hermes autonomous merge

The sister repo [`xrf9268-hue/Wink`](https://github.com/xrf9268-hue/Wink/tree/main/.github) solves the equivalent problem on a human-merge repo by:

1. Turning on GitHub's `required_conversation_resolution` rule on `main`.
2. Adding a `Review Gate / Validate review state` required status check backed by [`validate-review-state.mjs`](https://github.com/xrf9268-hue/Wink/blob/main/.github/scripts/validate-review-state.mjs), which fails (exit 1) when any review thread is `isResolved=false AND isOutdated=false`.

We **cannot copy that wholesale** here. Either of the two pieces above would *block the merge button*, which on this repo means Hermes' autonomous merge pipeline jams the moment any reviewer (human or bot) leaves a thread open. The fix has to keep the merge train moving while still capturing the feedback that gets shipped past.

## Proposed shape (non-blocking)

1. **Capture-at-close workflow.** New `.github/workflows/capture-unresolved-reviews.yml` that triggers on `pull_request.closed` with `merged == true`. It enumerates the PR's review threads via GraphQL, filters to `isResolved=false AND isOutdated=false`, and for each one auto-files a follow-up issue with:
   - back-link to the discussion comment (`discussion_rXXXXXXXXXX` permalink)
   - file path + line number
   - author handle (bot or human — author-agnostic)
   - original comment body
   - cross-link back to the merged PR
   - heuristic priority label derived from the comment body (`P0`/`P1`/`P2`/`P3` badges that Codex already emits → matching `priority:*` label; default `priority:p2`)
2. **Reuse Wink's GraphQL primitive.** [`validate-review-state.mjs`](https://github.com/xrf9268-hue/Wink/blob/main/.github/scripts/validate-review-state.mjs) already enumerates exactly the right thread set, author-agnostic, with the thread-shape parsing done correctly. Lift the GraphQL query + thread parsing into `.github/scripts/capture-unresolved-reviews.mjs`; drop the exit-1 branch; replace it with a `gh api repos/.../issues -X POST` (or `octokit.issues.create`) call per unresolved thread.
3. **Optional: non-required status check that just reports the count.** A `review-feedback / unresolved-bot-threads: N` status check (passing always) that Hermes can read as a signal while deciding whether to self-resolve before pushing the merge. This is *informational* — it does not block merge. Out of scope to wire Hermes itself; in scope to publish the signal.
4. **Out of scope here.** Teaching Hermes to actually action bot feedback before merging is the right long-term answer but a bigger change involving the Hermes side. The capture-at-close workflow is the minimum guardrail that makes "shipped past a P1 nit" recoverable without manual triage.

## Acceptance criteria

- [ ] `.github/workflows/capture-unresolved-reviews.yml` exists and triggers on `pull_request.closed && merged == true`, and also supports `workflow_dispatch` with a `pull_number` input for retroactive runs.
- [ ] `.github/scripts/capture-unresolved-reviews.mjs` (or equivalent) enumerates `isResolved=false AND isOutdated=false` threads via GraphQL and opens one follow-up issue per unique thread, with PR + discussion-permalink + file:line + author + body all populated.
- [ ] Author-agnostic: a human-left unresolved nit produces an issue just like a bot-left one. No allowlist/denylist of accounts.
- [ ] Idempotency, defined precisely: before creating, the script searches open *and* closed issues for the exact discussion permalink (e.g. `#discussion_r3252781671`) appearing anywhere in the issue body. If a match exists, skip silently and log the existing issue number. Permalink uniqueness is the dedup key — multiple bot threads that happen to share a root cause still produce separate issues unless explicitly back-linked.
- [ ] Retroactive smoke test against PR #106 via `workflow_dispatch`: because #107 (and this issue, #108) already contain both `discussion_r3252781671` and `discussion_r3252794498` in their bodies, a dry-run dispatch on #106 must produce **zero** new issues and log both threads as "already tracked by #107/#108". This validates the dedup path, not the issue-creation path.
- [ ] Forward end-to-end test: pick the next merged PR that carries an unresolved thread not yet referenced by any existing issue, and confirm exactly one new follow-up issue is filed with all fields populated. (Alternative: a fixture-driven unit test on the script using a recorded GraphQL response.)
- [ ] No required status check is added. Hermes' merge path must remain unblocked.
- [ ] Documentation: short note in `AGENTS.md` (or wherever the merge-flow guidance lives) describing the new behavior so contributors know unresolved threads on a merged PR will surface as new issues.

## References

- PR #106 (the trigger event): https://github.com/xrf9268-hue/aiops-platform/pull/106
- Issue #107 (the bug that escaped past unresolved bot reviews on #106): https://github.com/xrf9268-hue/aiops-platform/issues/107
- Wink's review-gate (block-on-merge variant we are deliberately *not* copying): https://github.com/xrf9268-hue/Wink/tree/main/.github
- Wink's GraphQL thread-state script we *do* want to lift: https://github.com/xrf9268-hue/Wink/blob/main/.github/scripts/validate-review-state.mjs

hey-yvan:
## Summary
- Add a non-blocking merged-PR workflow that captures unresolved, non-outdated review threads as follow-up GitHub issues.
- Add a Node script with fixture/unit coverage for priority labels, discussion-permalink idempotency, issue body fields, author-agnostic capture, and PR #106 dry-run tracking logs.
- Document the guardrail in AGENTS.md and run the script tests in CI without adding any required status check.

## Test Plan
- [x] RED: `node --test .github/scripts/capture-unresolved-reviews.test.mjs` failed before the script existed.
- [x] `node --test .github/scripts/capture-unresolved-reviews.test.mjs`
- [x] `GITHUB_TOKEN=$(gh auth token) REPO_OWNER=xrf9268-hue REPO_NAME=aiops-platform node .github/scripts/capture-unresolved-reviews.mjs --pull-number 106 --dry-run` (0 created/would-create; both PR #106 discussion permalinks logged as already tracked by #107/#108)
- [x] `go test ./...`
- [x] `go build ./cmd/trigger-api ./cmd/worker ./cmd/linear-poller`
- [x] `go test -race -covermode=atomic ./...` attempted; blocked locally by Raspberry Pi 16KB-page ThreadSanitizer unsupported VMA range, so CI remains the race gate.

Closes #108

### DIFF
--- a/.github/scripts/capture-unresolved-reviews.mjs
+++ b/.github/scripts/capture-unresolved-reviews.mjs
@@ -1,0 +1,363 @@
+import { readFile } from 'node:fs/promises';
+import { pathToFileURL } from 'node:url';
+
+const apiVersion = '2022-11-28';
+const defaultLabels = ['type:chore'];
+
+function requiredEnv(name) {
+  const value = process.env[name];
+  if (!value) {
+    throw new Error(`Missing required environment variable: ${name}`);
+  }
+  return value;
+}
+
+function githubToken() {
+  return process.env.GITHUB_TOKEN ?? process.env.GH_TOKEN ?? requiredEnv('GITHUB_TOKEN');
+}
+
+async function githubRequest(path, { method = 'GET', body } = {}) {
+  const response = await fetch(`https://api.github.com${path}`, {
+    method,
+    headers: {
+      Accept: 'application/vnd.github+json',
+      Authorization: `Bearer ${githubToken()}`,
+      'Content-Type': 'application/json',
+      'User-Agent': 'aiops-platform-capture-unresolved-reviews',
+      'X-GitHub-Api-Version': apiVersion,
+    },
+    body: body ? JSON.stringify(body) : undefined,
+  });
+
+  const text = await response.text();
+  const payload = text ? JSON.parse(text) : null;
+  if (!response.ok) {
+    const details = JSON.stringify(payload ?? text, null, 2);
+    throw new Error(`GitHub REST request failed ${method} ${path}: ${details}`);
+  }
+  return payload;
+}
+
+async function graphqlRequest(query, variables) {
+  const response = await fetch('https://api.github.com/graphql', {
+    method: 'POST',
+    headers: {
+      Accept: 'application/vnd.github+json',
+      Authorization: `Bearer ${githubToken()}`,
+      'Content-Type': 'application/json',
+      'User-Agent': 'aiops-platform-capture-unresolved-reviews',
+      'X-GitHub-Api-Version': apiVersion,
+    },
+    body: JSON.stringify({ query, variables }),
+  });
+
+  const payload = await response.json();
+  if (!response.ok || payload.errors?.length) {
+    const details = JSON.stringify(payload.errors ?? payload, null, 2);
+    throw new Error(`GraphQL request failed: ${details}`);
+  }
+  return payload.data;
+}
+
+export function extractPriorityLabel(body) {
+  const match = body.match(/\[(P[0-3])\]/i);
+  if (!match) {
+    return 'priority:p2';
+  }
+  return `priority:${match[1].toLowerCase()}`;
+}
+
+export function normalizeDiscussionPermalink(url) {
+  const parsed = new URL(url);
+  const discussion = parsed.hash.match(/^#discussion_r\d+$/)?.[0];
+  if (!discussion) {
+    throw new Error(`Review comment URL is missing a discussion anchor: ${url}`);
+  }
+  parsed.hash = discussion;
+  parsed.search = '';
+  parsed.pathname = parsed.pathname.replace(/\/files$/, '');
+  return parsed.toString();
+}
+
+function firstComment(thread) {
+  if (Array.isArray(thread.comments)) {
+    return thread.comments[0] ?? {};
+  }
+  return thread.comments?.nodes?.[0] ?? {};
+}
+
+function authorLogin(comment) {
+  return comment.author?.login ?? comment.author ?? 'ghost';
+}
+
+function commentBody(comment) {
+  return comment.bodyText ?? comment.body ?? '';
+}
+
+function commentUrl(comment) {
+  return comment.url ?? comment.htmlUrl ?? comment.html_url;
+}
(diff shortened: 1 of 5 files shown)


hey-yvan:
@codex review

hey-yvan:
@codex review

Issue #95, opened 2026-05-15
[P1][spec-alignment] D21 Single-source-of-truth orchestrator state (SPEC §3.1, §4.1.8, §7.4)


hey-yvan:
## Background

SPEC.md §3.1 + §4.1.8 + §7.4:
> "The orchestrator is the only component that mutates scheduling state. All worker outcomes are reported back to it and converted into explicit state transitions."

`OrchestratorRuntimeState` contains `running`, `claimed`, `retry_attempts`, `completed`, `codex_totals`, `codex_rate_limits`.

## Today

There is no orchestrator struct at all. `cmd/worker/main.go:25-35` is a Postgres-backed queue consumer. State lives in the `tasks` and `task_events` SQL tables (`internal/queue/postgres.go`), and concurrency is enforced by `FOR UPDATE SKIP LOCKED` rather than by an in-process authority.

## Gap

SPEC §7.4 "serializes state mutations through one authority" is implemented by the database transaction model instead, which has different semantics — there is no `claimed` set distinct from `running`, no `retry_attempts` map distinct from `available_at`, no aggregate `codex_totals`. The "Orchestrator" component in §3.1 is not present in the Go code; nearly every other deviation in this audit is downstream of this one.

## Severity

High — structural; this is the umbrella under which most other deviations sit. D6 (#73) is the **partial** form of this work (remove the Postgres queue); D21 is the **full** form (build the orchestrator state machine that replaces it).

## Dependencies / related

- **Umbrella for: D6 (#73)** — queue removal is the precondition; orchestrator state machine is what replaces it
- **Precondition for: D15** (per-state caps), **D16** (retry attempts map), **D19** (candidate selection counts in-flight by state)
- Related: D22 (orchestrator owns the event vocabulary the runtime emits)
- Cross-reference: PR #82 audit, §"D21 — Single-source-of-truth orchestrator state is absent"

## Acceptance criteria

- [ ] New `orchestrator` package with an in-memory state machine that owns: `running`, `claimed`, `retry_attempts`, `completed`, `codex_totals`, `codex_rate_limits`.
- [ ] All scheduling state mutations go through orchestrator methods; SQL row-locking is no longer the concurrency boundary.
- [ ] Worker reports outcomes back to the orchestrator; the orchestrator converts them into explicit state transitions (matched to D22's event vocabulary).
- [ ] Restart recovery rebuilds state from tracker + filesystem per SPEC §14.3 (overlap with D3 #68).
[...]

PR #117, opened 2026-05-17
fix(orchestrator): ignore stale run finalization

hey-yvan:
## Summary
- Ignore stale run-finalization calls when the orchestrator no longer has a matching `Running` entry.
- Return a boolean from `FinishRunSucceeded` / `FinishRunFailed` so callers can distinguish applied transitions from stale outcomes.
- Add regression coverage proving stale finalization does not release claims, mark completion, or inflate Codex runtime totals.

## Test Plan
- `/home/pi/.local/go1.25.10/bin/go test ./internal/orchestrator`
- `/home/pi/.local/go1.25.10/bin/go test ./...`
- Diff-only independent review via `claude -p ... --allowedTools '' --max-turns 1` returned `LGTM`.

Refs #95

### DIFF
--- a/internal/orchestrator/actor.go
+++ b/internal/orchestrator/actor.go
@@ -374,6 +374,7 @@ type finalizeRunOp struct {
 	attempt    *int
 	result     WorkerResult
 	started    time.Time
+	entry      *RunningEntry
 	done       chan struct{}
 }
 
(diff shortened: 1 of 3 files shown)


hey-yvan:
@codex review

hey-yvan:
@codex review

PR #165, opened 2026-05-19
fix: guard retry fires by kind

### LINKED ISSUE #145: retryFireOp stale-fire guard ignores RetryKind, can mis-dispatch on Continuation↔Failure swap
## Summary

`retryFireOp.apply` at `internal/orchestrator/actor.go:645` only compares `entry.Attempt != r.attempt` to decide whether a fired timer is stale. The accompanying comment (lines 625-630, 645-649) explicitly acknowledges the timer-replace race that the guard is meant to defend against, but the guard is **incomplete**: `RetryKind` is not part of the equality check. If a `RetryKindContinuation` entry is replaced by a `RetryKindFailure` entry (or vice versa) while the prior timer's callback was already in flight, both entries can have the same `Attempt` value and the stale fire proceeds against the new entry of a different `Kind`.

## Evidence

- Stale-fire guard: `internal/orchestrator/actor.go:638-650`
  ```go
  entry, ok := st.RetryAttempts[r.id]
  if !ok {
      return nil
  }
  if entry.Attempt != r.attempt {
      // A newer ScheduleRetry replaced this entry; the older timer
      // fired late. Drop the stale fire — the newer entry will
      // re-dispatch on its own timer.
      return nil
  }
  ```
- Kind is then dispatched off the (possibly NEW) entry's value: `actor.go:651` (`if entry.Kind == RetryKindContinuation`) and the failure branch at `actor.go:663`. Both branches close over `r.issue` / `r.attempt` captured from the **OLD** scheduleRetryOp.
- `scheduleRetryOp.apply` captures `s.issue` and `s.attempt` into the `time.AfterFunc` closure (`actor.go:613-620`) and replaces any prior entry via `st.ScheduleRetry(entry)` (line 621). `ScheduleRetry` stops the prior `*time.Timer` but `time.Timer.Stop()` does NOT cancel an already-fired-but-not-yet-run callback — this is precisely the race the comment at line 627 cites.

## Concrete failure scenarios

Both require Attempt to collide across the swap. Easiest collision: `Attempt=1` on both sides.

1. **Continuation(Attempt=1) → Failure(Attempt=1)**. Worker exits clean → continuation retry scheduled. Before the 1s timer fires, a fresh failure path (e.g. follow-on op marks it failed) calls `ScheduleRetry` with Kind=Failure, Attempt=1. The stale continuation timer fires, finds entry exists, Attempt matches, but `entry.Kind == RetryKindFailure` now → falls through to the failure branch (line 663+), evaluates capacity, and spawns from `r.issue` (the OLD continuation's issue snapshot) with `r.attempt=1`. The new failure entry is silently consumed without the legitimate failure-path Issue snapshot ever being used to dispatch.
2. **Failure(Attempt=1) → Continuation(Attempt=1)**. After a failed attempt's retry is queued, a successful clean exit on a different code path (e.g. orchestrator state already healed via reconcile) queues a continuation with Attempt=1. The stale failure timer fires, Attempt matches, Kind is now Continuation → falls into the wake-pollloop branch (line 651-662) which is harmless on its own, BUT the failure entry's own `time.AfterFunc` timer for the 10s delay has been started under the wrong assumption (a stale fire just consumed nothing visible). The continuation's timer at line 659 is set to nil and the entry hangs until the next reconcile.

Scenario 1 is the dangerous one — it bypasses the failure retry's intended backoff and uses the wrong Issue snapshot.

## Why the comment's mitigation is wrong

The comment at line 627-630 says:

> Two timers may race here in pathological cases (a ScheduleRetry replace where the prior timer's Stop missed because the callback was already queued); the attempt-equality guard makes the stale fire a no-op.

The guard only makes the stale fire a no-op when `Attempt` differs. Attempt-equality across Kind boundaries is not unusual — it is the **default** for any first-cycle swap.

## Suggested fix

Capture `Kind` in `retryFireOp` and include it in the equality check:

```go
type retryFireOp struct {
    o       *Orchestrator
    id      IssueID
    issue   tracker.Issue
    attempt int
    kind    RetryKind // NEW
}

// in scheduleRetryOp.apply, when constructing the timer:
entry.Timer = time.AfterFunc(delay, func() {
    _ = o.submit(o.runCtx, &retryFireOp{
        o: o, id: id, issue: issue, attempt: attempt,
        kind: s.kind, // NEW
    })
})

// in retryFireOp.apply:
if entry.Attempt != r.attempt || entry.Kind != r.kind {
    return nil
}
```

Alternative: stamp each entry with a monotonic generation number that the timer captures, and compare generations. This is more robust if Kind ever gains more variants but is heavier.

## Acceptance criteria

- [ ] `retryFireOp` carries `Kind` and the stale-fire guard compares it.
- [ ] Unit test in `internal/orchestrator/actor_test.go` (or wherever retry tests live) that:
  1. schedules a Continuation(Attempt=1), captures the timer,
  2. before the timer fires, replaces it with a Failure(Attempt=1),
  3. fires the stale timer's callback synchronously,
  4. asserts the fire is dropped (no `spawn` call, no `Claimed` mutation).
- [ ] Mirror test for Failure(Attempt=1) → Continuation(Attempt=1).

## Refs

- PR #140 (added the continuation retry path; introduced the Kind dimension)
- SPEC §16.6 (retry semantics)

hey-yvan:
## Summary
- Capture `RetryKind` in retry timer fire operations and include it in the stale-fire equality guard.
- Preserve the captured kind when retry fires are re-armed after global or per-state capacity deferral.
- Add regression coverage for Continuation(Attempt=1) -> Failure(Attempt=1) and Failure(Attempt=1) -> Continuation(Attempt=1) stale timer races.

Closes #145

## Test Plan
- [x] `/home/pi/.local/go1.25.10/bin/go test ./internal/orchestrator -run 'TestRetryFire_DropsStale(Continuation|Failure)FireAfter.*Replacement' -count=1`
- [x] `/home/pi/.local/go1.25.10/bin/go test ./internal/orchestrator -count=1`
- [x] `/home/pi/.local/go1.25.10/bin/go test ./...`
- [x] `git diff --check`
- [x] `/home/pi/.local/go1.25.10/bin/go mod tidy && git diff --exit-code -- go.mod go.sum`
- [x] `/home/pi/.local/go1.25.10/bin/go build ./cmd/worker ./cmd/linear-poller ./cmd/gitea-poller`

## Notes
- `go test -race -covermode=atomic ./...` was attempted but this Raspberry Pi host cannot run Go ThreadSanitizer: `FATAL: ThreadSanitizer: unsupported VMA range; Found 47 - Supported 48`.
- SPEC context checked: SPEC §16.6 retry semantics and the Elixir orchestrator GenServer single-authority retry/timer model.

### DIFF
--- a/internal/orchestrator/actor.go
+++ b/internal/orchestrator/actor.go
@@ -718,6 +718,7 @@ func (s *scheduleRetryOp) apply(st *OrchestratorState) func() {
 	o := s.o
 	issue := s.issue
 	attempt := s.attempt
+	kind := s.kind
 	delay := o.scheduler.NextDelay(s.req)
 	// time.AfterFunc schedules immediately and is cheap (no goroutine
 	// until fire), so we can safely create the timer on the actor
@@ -738,6 +739,7 @@ func (s *scheduleRetryOp) apply(st *OrchestratorState) func() {
 			id:      id,
 			issue:   issue,
 			attempt: attempt,
+			kind:    kind,
 		})
 	})
 	st.ScheduleRetry(entry)
@@ -748,13 +750,14 @@ func (s *scheduleRetryOp) apply(st *OrchestratorState) func() {
 // SPEC §16.6 retry path is "if the entry is still queued, re-dispatch;
 // otherwise drop the fire." Two timers may race here in pathological
 // cases (a ScheduleRetry replace where the prior timer's Stop missed
-// because the callback was already queued); the attempt-equality
+// because the callback was already queued); the attempt/kind equality
 // guard makes the stale fire a no-op.
 type retryFireOp struct {
 	o       *Orchestrator
 	id      IssueID
 	issue   tracker.Issue
 	attempt int
+	kind    RetryKind
 }
 
 func (r *retryFireOp) apply(st *OrchestratorState) func() {
@@ -764,7 +767,7 @@ func (r *retryFireOp) apply(st *OrchestratorState) func() {
 		// earlier fire of the same retry. Either is correct.
 		return nil
 	}
-	if entry.Attempt != r.attempt {
+	if entry.Attempt != r.attempt || entry.Kind != r.kind {
 		// A newer ScheduleRetry replaced this entry; the older timer
 		// fired late. Drop the stale fire — the newer entry will
 		// re-dispatch on its own timer.
@@ -797,6 +800,7 @@ func (r *retryFireOp) apply(st *OrchestratorState) func() {
 		o := r.o
 		id := r.id
 		attempt := r.attempt
+		kind := r.kind
 		entry.DueAt = time.Now().Add(retryCapacityRecheckDelay)
 		entry.Timer = time.AfterFunc(retryCapacityRecheckDelay, func() {
 			_ = o.submit(o.runCtx, &retryFireOp{
@@ -804,6 +808,7 @@ func (r *retryFireOp) apply(st *OrchestratorState) func() {
 				id:      id,
 				issue:   issue,
 				attempt: attempt,
+				kind:    kind,
 			})
 		})
 		return nil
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

PR #168, opened 2026-05-19
feat: expose Gitea pagination cap metric

### LINKED ISSUE #129: gitea poller: expose pagination-cap truncation as a metric, not just a log line (PR #120 follow-up)
## Summary

`internal/gitea/tracker_client.go:89-138` caps each label-scoped Gitea issue listing at `listIssuesMaxPages = 20` pages × `listIssuesPageSize = 50` = 1000 issues. When the cap is hit, the loop returns the truncated result and only logs via `c.Logf`:

```go
if page > listIssuesMaxPages {
    if len(batch) == 0 {
        return out, nil
    }
    if c.Logf != nil {
        c.Logf("gitea issue pagination exceeded %d pages for label %q; returning capped result set", listIssuesMaxPages, labelName)
    }
    return out, nil
}
```

The truncation is intentional and tested (`tracker_client_test.go:282-285`), so this is **not a bug** — the design correctly degrades instead of failing the whole poll. The concern is operational visibility, not correctness.

## Why this matters

- The Gitea poller drives orchestrator dispatch. If a label set ever grows past 1000 open issues, the orchestrator silently stops considering anything past page 20 — and the only signal is a single log line per poll cycle.
- There is no counter, no metric, no health-check surface. A log-only signal is easy to miss across all of: log retention rollover, log-level filtering, and operator vacation days.
- The `+1` probe page deliberately distinguishes "exactly 1000" from ">1000", so the trigger condition is already known at the call site — exposing it is cheap.

## Suggested shape (one or both)

1. Track a counter on `TrackerClient` (`PaginationCapHits int64` or a `prometheus.Counter` if wiring is acceptable), incremented when the >1000 branch fires; surface via the existing `/api/v1/state` snapshot or metrics endpoint.
2. Optionally make the behavior policy-driven: a `TrackerConfig.FailOnPaginationCap` (default false to preserve current degradation) that returns an error in environments where silent truncation is unacceptable.

## Acceptance

- [ ] Cap hit is observable from outside the process (counter or metric), not just `Logf`.
- [ ] Test verifies the counter increments on the cap path (extends the existing `tracker_client_test.go` case).
- [ ] Docs/runbook note in `docs/runbooks/local-dev.md` (or the equivalent operator runbook) describes the 1000-issue limit and how to monitor it.

## Refs

- PR #120 (feat: add Gitea aiops label reader)

hey-yvan:
## Summary
- add an atomic Gitea pagination-cap counter on TrackerClient
- expose the counter through an opt-in gitea-poller Prometheus-style /metrics endpoint
- document the 1000-issue-per-label cap and monitoring command in the local dev runbook

Closes #129

## Test Plan
- [x] go test ./internal/gitea -run TestTrackerClientListIssuesByStatesReturnsCappedResultsInsteadOfFailingWhenPageLimitExceeded -count=1
- [x] go test ./cmd/gitea-poller -run TestGiteaMetricsHandlerExposesPaginationCapHits -count=1
- [x] go test ./cmd/gitea-poller ./internal/gitea -count=1
- [x] go test ./...
- [x] gofmt -l $(git ls-files '*.go')
- [x] go mod tidy && git diff --exit-code -- go.mod go.sum
- [x] go build ./cmd/worker ./cmd/linear-poller ./cmd/gitea-poller
- [ ] go test -race -covermode=atomic ./... (blocked locally: ThreadSanitizer unsupported VMA range on this Raspberry Pi host: "FATAL: Found 47 - Supported 48")

### DIFF
--- a/cmd/gitea-poller/main.go
+++ b/cmd/gitea-poller/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
 	"os"
 	"strings"
 	"time"
@@ -42,6 +43,9 @@ func main() {
 	baseURL := giteaBaseURL(wf.Config.Tracker)
 	client := gitea.NewTrackerClient(wf.Config.Tracker, baseURL, wf.Config.Repo.Owner, wf.Config.Repo.Name)
 	client.Logf = log.Printf
+	if metricsAddr := os.Getenv("GITEA_POLLER_METRICS_ADDR"); metricsAddr != "" {
+		startMetricsServer(metricsAddr, client)
+	}
 	interval := time.Duration(wf.Config.Tracker.PollIntervalMs) * time.Millisecond
 
 	for {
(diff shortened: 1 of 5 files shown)


hey-yvan:
@codex review

hey-yvan:
@codex review

PR #170, opened 2026-05-20
fix: preserve lifecycle hook ordering

### LINKED ISSUE #146: PR #134 lifecycle hook ordering bugs: after_run fires without before_run, before_remove skipped on after_create failure
## Summary

PR #134 added the SPEC §5.3.4 / §9.4 / §18.1 workspace lifecycle hooks (`after_create`, `before_run`, `after_run`, `before_remove`). Two ordering bugs survived review.

## Bug 1 — `after_run` fires without matching `before_run`

**Location:** `internal/worker/runtask.go:199-205` (call to RunWorkspaceHook("after_run") inside the failure path after `runner.New`).

If `runner.New` fails, the worker still invokes `after_run`. But `before_run` is gated behind a successful runner construction further down — so `after_run` runs without its lifecycle partner ever firing. Hooks that maintain pair invariants (e.g. acquire/release of an external resource, increment/decrement of a counter, start/stop of a profile timer) will leak or underflow.

The SPEC §5.3.4 contract is that `before_run` always precedes `after_run` for a given run attempt. Calling `after_run` for a run attempt that never started violates that contract.

## Bug 2 — `before_remove` skipped on `after_create` failure

**Locations:**
- `internal/worker/runtask.go:173-175` — after_create failure path calls `os.RemoveAll(workdir)` bare.
- `cmd/worker/main.go:128-131` — startup reconcile cleanup similarly tears down workspaces without a `before_remove` dispatch.

SPEC §5.3.4: `before_remove` is the hook that lets the workflow run cleanup (e.g. unmount, push artifacts, stop background processes) **before** the workspace directory disappears. Tearing the directory down bare defeats the only contract that hook offers.

## Suggested fix

1. **Bug 1:** Either lift the `after_run` call to a `defer` that is armed *after* `before_run` succeeds, or simply do not call `after_run` on the `runner.New` failure path. The latter is simpler and matches the SPEC pairing semantics.

   ```go
   r, err := runner.New(...)
   if err != nil {
       // Do NOT call after_run — before_run was never dispatched.
       return ...
   }
   if err := manager.RunWorkspaceHook(ctx, "before_run", ...); err != nil { ... }
   defer manager.RunWorkspaceHook(context.Background(), "after_run", ...)
   ```

2. **Bug 2:** Wrap `os.RemoveAll(workdir)` with a `before_remove` dispatch in both call sites. If `before_remove` itself fails, log + proceed with removal (do not block teardown indefinitely), but the hook MUST be given the chance to run.

## Acceptance criteria

- [ ] Test in `internal/worker/runtask_test.go` that injects a failing `runner.New` and asserts the after_run hook is NOT invoked while before_run was not invoked either.
- [ ] Test that injects a failing `after_create` hook and asserts `before_remove` IS invoked exactly once before the workdir is removed.
- [ ] Same coverage for the startup-reconcile teardown path in `cmd/worker/main.go`.
- [ ] DEVIATIONS.md row D12 wording updated if the bugs are large enough to keep D12 in "Partial" rather than "Closed".

## Refs

- PR #134
- SPEC §5.3.4, §9.4, §18.1

hey-yvan:
## Summary
- Prevent `after_run` from firing when runner setup fails before `before_run` can run.
- Dispatch `before_remove` before removing a workspace after `after_create` hook failure.
- Add regression coverage for runner setup failure, after_create cleanup, and keep D12 marked Partial while lifecycle follow-up fixes are in progress.

Fixes #146

## Test Plan
- [x] `/home/pi/.local/go1.25.10/bin/go test ./internal/worker -run 'TestRunTaskDoesNotExecuteAfterRunHookWhenRunnerSetupFails|TestRunTaskRunsBeforeRemoveHookWhenAfterCreateFails|TestReconcileStartupRunsBeforeRemoveHookAndStillRemovesOnFailure' -count=1 -v`
- [x] `/home/pi/.local/go1.25.10/bin/go test ./internal/worker -count=1`
- [x] `/home/pi/.local/go1.25.10/bin/go test ./...`
- [x] `/home/pi/.local/go1.25.10/bin/go mod tidy && git diff --exit-code -- go.mod go.sum`
- [x] `/home/pi/.local/go1.25.10/bin/go build ./cmd/worker ./cmd/linear-poller ./cmd/gitea-poller`
- [ ] `/home/pi/.local/go1.25.10/bin/go test -race -covermode=atomic ./...` (blocked locally on this Raspberry Pi by ThreadSanitizer: `unsupported VMA range`, `Found 47 - Supported 48`; CI should run this on supported runners)


hey-yvan:
@codex review

hey-yvan:
@codex review

PR #171, opened 2026-05-20
fix(worker): honor workflow workspace root at runtime

### LINKED ISSUE #149: workspace.root YAML key is parsed but ignored at runtime (silently overridden by WORKSPACE_ROOT env)
## Summary

The `workspace.root` field is part of the SPEC §5.3 / §9.1 workflow front-matter schema and is accepted by the Go loader, but the worker never reads it at runtime — it always uses the `WORKSPACE_ROOT` environment variable. An operator who configures `workspace.root` in `WORKFLOW.md` will see their setting silently ignored.

## Evidence

- Loader accepts the key: `internal/workflow/config.go:67-71`.
- Worker reads from env, not from config: `internal/worker/run.go:47`:
  ```go
  root := os.Getenv("WORKSPACE_ROOT")
  ```
  No reference to `cfg.Workspace.Root` in the runtime path. Verify with:
  ```bash
  grep -rn 'Workspace.Root\|workspace.root' internal/ cmd/
  ```

## Why this matters

- Silent override is the worst failure mode: the operator believes their YAML setting is in effect; reality is different.
- This violates the SPEC §5.3 / §6.2 contract that the workflow front-matter is the source of truth for runtime configuration (env should fall back to YAML, not override it).
- D24 in DEVIATIONS.md already calls out gaps in the front-matter schema but does not flag this specific *parsed-but-dead* shape.

## Fix options

1. **Honor it.** Have the worker prefer `cfg.Workspace.Root` when set; fall back to `WORKSPACE_ROOT` env only when YAML is absent. Document precedence.
2. **Reject it.** Have the loader return an error when `workspace.root` is set but the runtime ignores it (interim measure, fragile — only useful if Honor It is not in scope).
3. **Move ownership.** If the long-term design says env wins (uncommon for SPEC-aligned implementations), remove the key from the schema entirely.

Recommend option 1. It is what SPEC implies and is a one-line worker change plus a precedence note in the docs.

## Acceptance criteria

- [ ] Decision documented in this issue or in a follow-up.
- [ ] Either the worker honors `cfg.Workspace.Root` with documented precedence, OR the schema/loader stops accepting the key.
- [ ] Test in `internal/worker/run_test.go` covering both env-only and YAML-only cases; a third case asserting precedence if both are set.
- [ ] `docs/runbooks/local-dev.md` and the README `WORKFLOW.md` reference table updated to reflect the resolved semantics.

## Refs

- SPEC §5.3, §6.2, §9.1
- D24 in DEVIATIONS.md (`#98`)

hey-yvan:
## Summary
- Honor explicit workflow `workspace.root` during startup reconciliation and task execution, using `WORKSPACE_ROOT` only as fallback.
- Preserve tracker source type directory names such as `linear_issue` in per-issue workspace paths while still sanitizing event IDs.
- Document runtime workspace precedence and record the Kanban throughput/worktree migration plan.

Closes #149

## Test Plan
- `/home/pi/.local/go1.25.10/bin/go test ./cmd/worker ./internal/worker ./internal/workspace -count=1`
- `/home/pi/.local/go1.25.10/bin/go test ./... -count=1`

## Automation safety notes
- Heartbeat remains paused until this PR passes CI/review and the shared checkout is restored clean.
- The local Hermes precheck currently reports `shared_workspace_blocked_reconciliation` for #149 rather than dispatching more implementation work.

### DIFF
--- a/.hermes/kanban-throughput-migration-plan.md
+++ b/.hermes/kanban-throughput-migration-plan.md
@@ -1,0 +1,59 @@
+# Kanban Throughput Migration Plan
+
+Updated: 2026-05-20T09:27:00+08:00
+
+## Problem snapshot
+- The aiops-platform Kanban board was operating in a low-throughput legacy mode: tasks used `workspace_kind=dir` with shared checkout `/home/pi/work/aiops-platform`, and `kanban.max_spawn` was effectively a single-worker lane.
+- The shared checkout is currently dirty on branch `fix/issue-149-workspace-root-precedence` with issue #149 changes. This means generic dispatch of later issues can mutate the wrong branch.
+- The embedded gateway dispatcher auto-spawned issue #150 into the dirty #149 checkout. That worker was reclaimed immediately as a misdirected spawn before it made a useful handoff.
+- Ready tasks were temporarily blocked as migration holds so no further worker is spawned against the shared checkout while worktree migration is in progress.
+
+## Target operating model
+1. Issue implementation runs in dedicated git worktrees under `/home/pi/work/aiops-platform-worktrees/issue-<number>`.
+2. `kanban.max_spawn` is a live concurrency cap. Use `2` initially on this Raspberry Pi; consider `3` only after CPU/memory/API behavior is stable.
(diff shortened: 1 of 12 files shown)


hey-yvan:
@codex review

hey-yvan:
@codex review

PR #251, opened 2026-05-21
fix(state-server): accept IPv6 loopback in Host validation (#235)

### LINKED ISSUE #235: [P2][bug] isLoopbackHTTPHost rejects IPv6 loopback `[::1]`; `/api/v1/*` is unreachable on IPv6-only hosts
## Summary

`cmd/worker/main.go:423-435` validates the HTTP `Host` header against an allow-list of loopback names:

```go
func isLoopbackHTTPHost(hostport string) bool {
    if hostport == "" {
        return false
    }
    host := hostport
    if h, _, err := net.SplitHostPort(hostport); err == nil {
        host = h
    } else if strings.Contains(hostport, ":") {
        return false
    }
    host = strings.Trim(host, "[]")
    return host == "127.0.0.1" || host == "localhost"
}
```

The function strips IPv6 brackets but does not include `::1` in the allow-list. On a host where the worker binds dual-stack (or to `[::1]:port` explicitly), an HTTP request to `http://[::1]:4000/api/v1/state` is rejected with **403 Forbidden** even though the request originated on the same host.

This violates SPEC §13.7:

> Implementations SHOULD bind loopback by default (`127.0.0.1` or **host equivalent**) unless explicitly configured otherwise.

`::1` is the host-equivalent on IPv6. Excluding it makes the loopback rule incomplete.

## Evidence

```bash
$ grep -n "::1\|localhost" cmd/worker/main.go
434:    return host == "127.0.0.1" || host == "localhost"
```

Only two host names accepted. The same function is called inside every handler (via line 415 `if !isLoopbackHTTPHost(r.Host)`), so all API routes — `/api/v1/state`, `/api/v1/<id>`, `/api/v1/refresh` — reject IPv6 loopback.

Reproduce on any IPv6-capable host:

```bash
$ AIOPS_WORKFLOW_PATH=examples/WORKFLOW.md go run ./cmd/worker &
$ curl -v http://[::1]:4000/api/v1/state
HTTP/1.1 403 Forbidden
```

The same curl against `127.0.0.1:4000` succeeds. The HTTP listener binds `127.0.0.1:port` by default (`cmd/worker/main.go` listener construction), so IPv6 reachability requires the operator to change `Host` header rules anyway — but the rejection happens before the listener check, when the Host header is `[::1]`.

## Why this matters

1. **Common test/dev environments default to IPv6.** Many `curl` invocations on Linux desktops resolve `localhost` to `::1` first (per `/etc/gai.conf`), and the request arrives with `Host: [::1]:4000`. The handler then 403s even though the actual TCP connection came in over loopback.
2. **Browser automation.** A localhost-bound test harness that connects to `http://[::1]:4000` is the canonical pattern in many CI environments.
3. **Defense-in-depth alignment with SPEC §13.7 loopback default.** "Loopback" without IPv6 is incomplete.

## Suggested fix

Add `::1` to the allow-list and verify the brackets stripping is correct:

```go
func isLoopbackHTTPHost(hostport string) bool {
    if hostport == "" {
        return false
    }
    host := hostport
    if h, _, err := net.SplitHostPort(hostport); err == nil {
        host = h
    } else if strings.Contains(hostport, ":") && !strings.HasPrefix(hostport, "[") {
        // Plain "host:port" with no IPv6 brackets and we failed to split -> reject.
        return false
    }
    host = strings.TrimPrefix(strings.TrimSuffix(host, "]"), "[")
    switch host {
    case "127.0.0.1", "localhost", "::1":
        return true
    }
    // Also accept other loopback addresses in 127.0.0.0/8 if needed:
    if ip := net.ParseIP(host); ip != nil && ip.IsLoopback() {
        return true
    }
    return false
}
```

`net.ParseIP(...).IsLoopback()` is the right primitive — it accepts `127.x.y.z` and `::1` and rejects routable addresses. Replacing the literal allow-list with this is the cleanest fix.

Tests in `cmd/worker/main_test.go` should cover (a) `127.0.0.1:4000`; (b) `localhost:4000`; (c) `[::1]:4000`; (d) `[::1]` without port; (e) a non-loopback IP `1.2.3.4:4000` (rejected).

## Acceptance criteria

- [ ] `[::1]` and `[::1]:<port>` Host headers pass.
- [ ] `127.0.0.0/8` IPv4 loopback addresses pass.
- [ ] Non-loopback IPs still rejected.
- [ ] Tests cover IPv4/IPv6 happy paths and a representative non-loopback rejection.

## Refs

- Code: `cmd/worker/main.go:423-435`
- SPEC §13.7 ("Implementations SHOULD bind loopback by default … or host equivalent")
- stdlib: `net.IP.IsLoopback`

hey-yvan:
Closes #235.

## Problem

`isLoopbackHTTPHost` (`cmd/worker/main.go:423-435`) validated the HTTP `Host` header against a literal allow-list of `"127.0.0.1"` and `"localhost"`. IPv6 loopback (`[::1]`) was rejected with **403 Forbidden** even though it is the IPv6 host-equivalent per SPEC §13.7. Other `127.0.0.0/8` loopback addresses were also rejected.

## Change

- `cmd/worker/main.go`: replace the literal allow-list with `net.ParseIP(host).IsLoopback()` + a `"localhost"` special case (since `ParseIP` doesn't resolve names). Strip IPv6 brackets when the host arrived bracket-without-port (`"[::1]"`).
- `cmd/worker/main_test.go`:
  - `TestIsLoopbackHTTPHost` table-driven: 12 cases covering IPv4, IPv4 `127.0.0.0/8`, `localhost`, bracketed / unbracketed IPv6, non-loopback IP, empty.
  - `TestStateHTTPServerAllowsIPv6LoopbackHost` mirrors the existing IPv4 handler test.

## Scope notes

- `"::1"` unbracketed is **rejected** — RFC 7230 requires IPv6 hosts in HTTP `Host` headers to be bracketed; the unbracketed form is malformed.
- Listener bind address is unchanged; this PR is about Host-header validation only.

## Verification

- gofmt + `go mod tidy` clean.

Fork CI: https://github.com/xrf-9527/aiops-platform/pull/8

Refs: `docs/superpowers/specs/2026-05-22-loopback-ipv6-design.md`.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -431,8 +431,14 @@ func isLoopbackHTTPHost(hostport string) bool {
 		// "host:port" without IPv6 brackets that failed to split — malformed.
 		return false
 	}
-	// Strip IPv6 brackets when the host arrived bracketed without a port (e.g. "[::1]").
-	host = strings.TrimPrefix(strings.TrimSuffix(host, "]"), "[")
+	// Strip IPv6 brackets only when the host is properly bracketed (e.g. "[::1]").
+	// Reject unpaired brackets — "[::1" or "::1]" are malformed Host values.
+	if strings.HasPrefix(host, "[") || strings.HasSuffix(host, "]") {
+		if !strings.HasPrefix(host, "[") || !strings.HasSuffix(host, "]") {
+			return false
+		}
+		host = host[1 : len(host)-1]
+	}
 	if host == "localhost" {
 		return true
 	}
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -427,11 +427,19 @@ func isLoopbackHTTPHost(hostport string) bool {
 	host := hostport
 	if h, _, err := net.SplitHostPort(hostport); err == nil {
 		host = h
-	} else if strings.Contains(hostport, ":") {
+	} else if strings.Contains(hostport, ":") && !strings.HasPrefix(hostport, "[") {
+		// "host:port" without IPv6 brackets that failed to split — malformed.
 		return false
 	}
-	host = strings.Trim(host, "[]")
-	return host == "127.0.0.1" || host == "localhost"
+	// Strip IPv6 brackets when the host arrived bracketed without a port (e.g. "[::1]").
+	host = strings.TrimPrefix(strings.TrimSuffix(host, "]"), "[")
+	if host == "localhost" {
+		return true
+	}
+	if ip := net.ParseIP(host); ip != nil && ip.IsLoopback() {
+		return true
+	}
+	return false
 }
 
 type stateSnapshotFunc func(context.Context) (orchestrator.StateView, error)
--- a/cmd/worker/main_test.go
+++ b/cmd/worker/main_test.go
@@ -662,6 +662,8 @@ func TestIsLoopbackHTTPHost(t *testing.T) {
 		{"[::1]:4000", true},
 		{"[::1]", true},
 		{"::1", false},
+		{"[::1", false},
+		{"::1]", false},
 		{"1.2.3.4:4000", false},
 		{"evil.example", false},
 		{"evil.example:4000", false},
--- a/cmd/worker/main_test.go
+++ b/cmd/worker/main_test.go
@@ -630,6 +630,52 @@ func TestStateHTTPServerAllowsLoopbackHost(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


hey-yvan:
Fork CI: https://github.com/xrf-9527/aiops-platform/pull/8

@codex review

hey-yvan:
Addressed the P2 malformed-bracket leak in 10c48a1 — single-sided bracket inputs ([::1, ::1]) now hit the malformed-input rejection path. Test cases added.

@codex review

PR #254, opened 2026-05-21
fix(linear): enforce 30s per-request network timeout (#212)

### LINKED ISSUE #212: [P3][spec-alignment] §11.2: Linear client uses http.DefaultClient with no 30s network timeout
## Summary

SPEC §11.2 ("Query Semantics (Linear)") specifies:

> Network timeout: `30000 ms`

The Go Linear client constructs requests via `http.DefaultClient`, which has no timeout. Under a stalled Linear endpoint (CDN hiccup, network partition), the orchestrator's poll tick can hang indefinitely instead of failing fast and continuing on the next tick (§14.2 "Tracker candidate-fetch failures: Skip this tick. Try again on next tick.").

## Evidence

`internal/tracker/linear.go:24-29`:

```go
HTTP    *http.Client
// ...
return &LinearClient{APIKey: cfg.APIKey, BaseURL: base, Config: cfg, HTTP: http.DefaultClient}
```

`internal/tracker/linear.go:458-468`:

```go
req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.BaseURL, bytes.NewReader(b))
// ...
httpClient := c.HTTP
if httpClient == nil {
    httpClient = http.DefaultClient
}
```

`http.DefaultClient.Timeout == 0` (zero = no timeout). The caller's `ctx` may carry a deadline, but only the orchestrator-side reconcile path provides one — the candidate fetch path uses the long-lived poll context.

## Why this matters

1. A wedged Linear endpoint stops the poll loop entirely; SPEC §14.2 explicitly expects "skip this tick, try again next tick" behavior, which requires a bounded request.
2. Tests that mock the endpoint may pass while production fails differently.
3. Other implementations on the SPEC (Elixir reference) implement the 30s ceiling.

## Suggested fix

Two options:

**A. Per-request context timeout** (preferred):

```go
reqCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
defer cancel()
req, err := http.NewRequestWithContext(reqCtx, http.MethodPost, c.BaseURL, bytes.NewReader(b))
```

**B. Client-level timeout** (simpler but applies to all calls including streaming):

```go
HTTP: &http.Client{Timeout: 30 * time.Second}
```

Option A is preferred because it makes the contract explicit per call site and survives future code that swaps `c.HTTP` for a custom client.

Also add a regression test in `internal/tracker/linear_test.go` that points the client at a `httptest.Server` whose handler sleeps 35 seconds, and asserts the call returns `context.DeadlineExceeded` (or a wrapped equivalent) before the test deadline.

## Acceptance criteria

- [ ] Every Linear API call enforces a 30-second timeout, either via per-call context or via a non-default `*http.Client`.
- [ ] A regression test in `internal/tracker/linear_test.go` proves the 30s ceiling.
- [ ] No regressions in the existing Linear tests (timeouts must allow normal latency).

## Refs

- SPEC §11.2 ("Network timeout: 30000 ms"), §14.2 (tracker candidate-fetch failures behavior)
- Code: `internal/tracker/linear.go:24-29, 458-468`
- Reference implementation: `elixir/lib/symphony_elixir/linear/client.ex` (30s ceiling)

hey-yvan:
Closes #212.

## Problem

`LinearClient.HTTP = http.DefaultClient` has no timeout, and the GraphQL request site at `internal/tracker/linear.go:530` used the caller's context as-is. The poll-tick path passes a long-lived context with no deadline, so a wedged Linear endpoint hung the poll loop indefinitely instead of failing per SPEC §14.2 ("skip this tick, try again next tick"). SPEC §11.2 specifies a 30 000 ms ceiling.

## Change

- `internal/tracker/linear.go`: add `RequestTimeout time.Duration` field on `LinearClient`; constant `defaultLinearRequestTimeout = 30 * time.Second`; `NewLinearClient` initialises the field.
- `internal/tracker/linear.go` GraphQL request method: resolve effective timeout (fall back to default if zero), `context.WithTimeout(ctx, timeout)`, build the request with the wrapped context, defer cancel.

Per-request `WithTimeout` (not `http.Client.Timeout`) keeps the ceiling explicit at the call site and survives any future code that swaps `c.HTTP` for a custom client. Existing tests inject `client.HTTP = srv.Client()` and continue to pass — the per-call timeout still fires there.

## Tests

- `TestNewLinearClient_DefaultsRequestTimeoutTo30s` pins the default value.
- `TestLinearClient_EnforcesRequestTimeout` blocks the server handler indefinitely and sets `RequestTimeout = 50ms`; asserts `errors.Is(err, context.DeadlineExceeded)`. The existing `*tracker.Error.Unwrap()` (linear.go:69) preserves the sentinel.

## Verification

- gofmt clean.

Fork CI: https://github.com/xrf-9527/aiops-platform/pull/9

Refs: `docs/superpowers/specs/2026-05-22-linear-timeout-design.md`.

### DIFF
--- a/docs/superpowers/specs/2026-05-22-linear-timeout-design.md
+++ b/docs/superpowers/specs/2026-05-22-linear-timeout-design.md
@@ -1,0 +1,88 @@
+# Linear client 30s network timeout — design
+
+**Date:** 2026-05-22
+**Issue:** [#212](https://github.com/xrf9268-hue/aiops-platform/issues/212)
+
+## Problem
+
+`internal/tracker/linear.go:101` constructs the Linear client with `HTTP: http.DefaultClient`. `http.DefaultClient.Timeout == 0`. The GraphQL request site at `linear.go:530` uses the caller's context as-is. The orchestrator's poll-tick path passes a long-lived context with no deadline, so a wedged Linear endpoint (CDN hiccup, network partition) hangs the poll loop indefinitely instead of failing fast per SPEC §14.2 "skip this tick, try again on next tick".
+
+SPEC §11.2 specifies a 30 000 ms network timeout for Linear queries.
+
+## Decision
+
+**Option A** from the issue body: wrap each request's context in `context.WithTimeout` at the call site. Expose the timeout as a `RequestTimeout time.Duration` field on `LinearClient`, default 30 s in `NewLinearClient`, overridable for tests.
+
+### Why Option A (per-call) over Option B (client.Timeout)
+
+- Option B sets `http.Client.Timeout`. That works but applies uniformly to anything done through that client; future code that introduces streaming or long-poll endpoints would inherit a 30 s ceiling silently.
+- Option A makes the contract explicit at every GraphQL call site and survives swapping `c.HTTP` for any other `*http.Client`.
+- Tests inject `client.HTTP = srv.Client()` — a per-call timeout still fires there.
+
+### Why a struct field (not a package-level var or hardcoded constant)
+
+- Hardcoded `30 * time.Second` forces the test to actually wait 30 s on a wedged endpoint. Unacceptable for CI.
+- Package-level mutable var races between parallel tests.
+- A `RequestTimeout` field defaults to 30 s in the constructor and is set per-client in tests. No global state.
+
+## What changes
+
+| File | Change |
+| --- | --- |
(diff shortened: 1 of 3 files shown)


hey-yvan:
Fork CI: https://github.com/xrf-9527/aiops-platform/pull/9

@codex review

PR #255, opened 2026-05-21
fix(linear): short-circuit ListIssuesByStates on empty/whitespace input (#233)

### LINKED ISSUE #233: [P3][spec-alignment] §17.3: Linear `ListIssuesByStates([])` makes an API call instead of returning empty without API call
## Summary

SPEC §17.3 (Issue Tracker Client conformance bullets) states:

> Empty `fetch_issues_by_states([])` returns empty without API call

`internal/tracker/linear.go:36` does not honor this contract. With an empty `states` slice, the function still proceeds through API-key validation and constructs a GraphQL query with `state: { name: { in: $states } }` where `$states = []`. The Linear API receives a real request (counted against rate limits) and returns an empty result.

Gitea client at `internal/gitea/tracker_client.go:108-115` correctly short-circuits:

```go
if len(issueIDs) == 0 {
    return map[string]string{}, nil
}
```

The Linear adapter is missing the same early-return.

## Evidence

`internal/tracker/linear.go:36-110` (start of function):

```go
func (c *LinearClient) ListIssuesByStates(ctx context.Context, states []string) ([]Issue, error) {
    if c.APIKey == "" {
        return nil, fmt.Errorf("Linear API key is required")
    }
    projectSlug := strings.TrimSpace(c.Config.ProjectSlug)
    if projectSlug == "" {
        return nil, fmt.Errorf("Linear project slug is required")
    }
    query := `query ListIssues($projectSlug: String!, $states: [String!], $first: Int!, $after: String) {
      issues(filter: { project: { slugId: { eq: $projectSlug } }, state: { name: { in: $states } } }, ...
```

No `len(states) == 0` guard. The call descends into pagination, sends a network round-trip, and returns `nil, nil` only after the GraphQL response.

Same shape is plausibly present in `internal/tracker/github.go` and any future tracker — anywhere `ListIssuesByStates` is implemented.

## Why this matters

1. **SPEC §17.3 Core Conformance bullet.** This is a normative test case in the conformance matrix; failing it is a documented gap.
2. **Wasted rate-limit budget.** Linear's GraphQL API has fixed quotas. Empty-states calls are pure waste — they cannot return useful data.
3. **`ReconcileStartup` path triggers this.** When `terminal_states` happens to be empty (operator misconfig), the startup reconcile makes a no-op API call. Same for `active_states`.

## Suggested fix

Add the guard at the top of every `ListIssuesByStates` implementation (Linear, GitHub, future):

```go
func (c *LinearClient) ListIssuesByStates(ctx context.Context, states []string) ([]Issue, error) {
    nonempty := nonEmptyStates(states)
    if len(nonempty) == 0 {
        return nil, nil
    }
    if c.APIKey == "" {
        return nil, fmt.Errorf("Linear API key is required")
    }
    // ...
}
```

`nonEmptyStates` (already used in `internal/worker/reconcile.go`) strips empty strings, so a list like `["", "  "]` is treated equivalently to `[]`.

Mirror the change in `internal/tracker/github.go` if absent there.

Add a regression test in `internal/tracker/linear_test.go` that uses an `httptest.Server` recording received requests and asserts zero requests for `ListIssuesByStates(ctx, []string{})`.

## Acceptance criteria

- [ ] Linear `ListIssuesByStates([])` returns `(nil, nil)` without an HTTP call.
- [ ] Same applies to whitespace-only states (`[" "]`, `[""]`).
- [ ] GitHub adapter audited and patched if it has the same gap.
- [ ] Regression test asserts zero HTTP traffic for the empty path.

## Refs

- SPEC §17.3 Core Conformance bullet
- Code: `internal/tracker/linear.go:36`, `internal/tracker/github.go` (check)
- Reference impl: `internal/gitea/tracker_client.go:108-115`

hey-yvan:
Closes #233.

## Problem

SPEC §17.3: empty `fetch_issues_by_states([])` returns empty **without API call**. Linear adapter proceeded through API-key + project-slug validation and made a real GraphQL request. GitHub adapter already short-circuits (`internal/tracker/github.go:97-99`); Linear was the lone offender.

## Change

`internal/tracker/linear.go:119` — after the existing API-key / project-slug validation, filter `states` to non-empty trimmed entries. If the cleaned list is empty, return `(nil, nil)` without building the GraphQL request. Pass the cleaned list as the `$states` variable so whitespace-only entries don't reach Linear as bogus state names either.

## Scope notes

- `ListActiveIssues` calls `ListIssuesByStates(c.Config.ActiveStates)` — guard covers it transitively.
- Credential-validation ordering preserved (matches GitHub adapter's pattern: validate creds first, then empty guard).
- No shared `nonEmptyStrings` helper — `internal/worker` and `internal/tracker` each have their own variant (`nonEmptyStates`, `nonEmptyGitHubStates`); a 4-line inline filter is simpler than a third helper.

## Tests

`TestListIssuesByStatesEmptyShortCircuitsWithoutAPICall` records HTTP requests via a recording handler and asserts zero requests across four input shapes: `nil`, `[]`, `[""]`, `["  ", "\t"]`.

## Verification

- gofmt clean.

Fork CI: https://github.com/xrf-9527/aiops-platform/pull/10

Refs: `docs/superpowers/specs/2026-05-22-linear-empty-states-design.md`.

### DIFF
--- a/docs/superpowers/specs/2026-05-22-linear-empty-states-design.md
+++ b/docs/superpowers/specs/2026-05-22-linear-empty-states-design.md
@@ -1,0 +1,82 @@
+# Linear `ListIssuesByStates([])` early return — design
+
+**Date:** 2026-05-22
+**Issue:** [#233](https://github.com/xrf9268-hue/aiops-platform/issues/233)
+
+## Problem
+
+SPEC §17.3 Core Conformance says "Empty `fetch_issues_by_states([])` returns empty without API call". `internal/tracker/linear.go:119` does not honor this — with `states == nil` or `[]` (or all-whitespace `[" "]`), it still validates API key + project slug, builds a GraphQL request, and dispatches a network call to Linear. Linear returns empty; the worker burns a quota slot for nothing.
+
+`internal/tracker/github.go:97` already short-circuits via `nonEmptyGitHubStates`. The Linear adapter is missing the equivalent.
+
+## Decision
+
+Add an empty/whitespace-aware filter at the top of `LinearClient.ListIssuesByStates`. If the cleaned list is empty, return `(nil, nil)` without API key / project-slug validation, without building the request. Pass the cleaned list as the GraphQL `$states` variable so whitespace-only entries don't reach Linear as bogus state names either.
+
+### Why filter precedes API-key validation
+
+Conformance says "without API call" — by the time the function knows the inputs cannot produce results, returning is correct regardless of credential state. A caller probing emptiness shouldn't be forced to set API keys first. (And the existing GitHub adapter places its `len(stateFilters) == 0` check at line 97-99, *after* token validation. The strictest reading of SPEC §17.3 says no API call; the strictest reading of "without API call" lets either order satisfy. The Linear placement here matches GitHub's by placing the empty-check after credential validation. **Updated:** match GitHub adapter ordering exactly to keep parity — credential checks first, then empty-state guard.)
+
+Final placement: after API key / project slug validation, before GraphQL construction, mirroring GitHub.
+
+### Why not extract a shared helper
+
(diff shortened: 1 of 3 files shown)


hey-yvan:
Fork CI: https://github.com/xrf-9527/aiops-platform/pull/10

@codex review

PR #256, opened 2026-05-21
fix(runner): cap codex app-server stdio reader at 10 MiB (#226)

### LINKED ISSUE #226: [P2][bug] Codex app-server stdio reader uses bufio.ReadBytes with no max-line cap (10 MB SPEC recommendation)
## Summary

`internal/runner/codex_app_server.go:349` reads each JSON-RPC framed line from the Codex subprocess via `c.reader.ReadBytes('\n')`. `bufio.Reader.ReadBytes` has no built-in length limit — it grows the internal buffer until either a newline arrives or memory is exhausted.

SPEC §10.1 lists a RECOMMENDED additional process setting:

> RECOMMENDED additional process settings:
> - Max line size: 10 MB (for safe buffering)

The current implementation enforces no max, so a malformed or malicious Codex app-server that writes a 4 GB line without a newline can OOM-kill the worker.

## Evidence

```bash
$ grep -n "ReadBytes\|MaxLineSize\|10<<20\|10\*1024\*1024" internal/runner/codex_app_server.go
349:		line, err := c.reader.ReadBytes('\n')
```

No buffer-size enforcement; `c.reader` is a vanilla `bufio.NewReader(stdout)` with the stdlib default 4 KB starting buffer.

Failure mode: an upstream app-server bug (or someone running a custom `codex.command` that misbehaves) emits an unbounded stream without a newline. The worker process RSS grows until the OOM killer fires. Other workers on the same host (or the worker itself) die.

## Why this matters

1. **SPEC RECOMMENDS** a 10 MB cap. Not following a SHOULD-strength rule requires a DEVIATIONS entry; today there is neither the cap nor the deviation row.
2. **Trust posture.** The Codex subprocess is the most exposed component in the system (executes generated code, fetches dependencies). Treating its output as fully trusted contradicts SPEC §15.1 / §15.5 harness-hardening guidance.
3. **Adjacent compromise.** A single misbehaving repo / agent can take out the worker that hosts other concurrent agents.

## Suggested fix

Wrap reads in a bounded reader:

```go
const maxAppServerLineBytes = 10 * 1024 * 1024 // 10 MiB per SPEC §10.1
// in initialization:
c.reader = bufio.NewReaderSize(io.LimitReader(stdout, ...), 64<<10)

// In readMessage, after ReadBytes:
if len(line) > maxAppServerLineBytes {
    c.recordRuntimeEvent(task.EventMalformed, c.withRuntimeContext(map[string]any{
        "reason": "line exceeds max size",
        "length": len(line),
    }))
    return nil, fmt.Errorf("codex app-server line exceeded %d bytes", maxAppServerLineBytes)
}
```

The cleanest implementation uses `bufio.Scanner` with `Buffer(make([]byte, 0, 64<<10), maxAppServerLineBytes)` instead of `bufio.Reader.ReadBytes`, since `Scanner` enforces the cap natively.

Also: align the test runner (`internal/runner/mock.go`) with the same cap so tests cover the cliff.

## Acceptance criteria

- [ ] Reader path bounded at 10 MiB (or configurable via `codex.max_line_bytes`).
- [ ] Overflow surfaces as `malformed` runtime event + returned error, not OOM.
- [ ] Test in `internal/runner/codex_app_server_test.go` simulates a long-line response and asserts graceful error.
- [ ] If the cap is intentionally different from SPEC's 10 MB recommendation, `DEVIATIONS.md` records the rationale.

## Refs

- Code: `internal/runner/codex_app_server.go:349, 60-72` (subprocess setup)
- SPEC §10.1 ("Max line size: 10 MB"), §15.1 (harness hardening)
- stdlib: `bufio.Scanner.Buffer` is the right primitive

hey-yvan:
Closes #226.

## Problem

`internal/runner/codex_app_server.go:readLine` used `bufio.Reader.ReadBytes('\n')` with no length cap. A misbehaving Codex app-server emitting a 4 GB line without newline grew the bufio buffer until OOM. SPEC §10.1 RECOMMENDS "Max line size: 10 MB".

## Change

Replace `*bufio.Reader` field with `*bufio.Scanner` configured via `Buffer(make([]byte, 0, 64<<10), maxAppServerLineBytes)` where `maxAppServerLineBytes = 10 * 1024 * 1024`. `bufio.Scanner.Scan()` enforces the cap during the read itself; over-cap lines surface as `bufio.ErrTooLong`. Memory stays bounded at the cap.

`readLine` wraps the error as `"codex app-server line exceeded 10485760 bytes: …"` preserving `errors.Is(err, bufio.ErrTooLong)`. Scanner strips the line terminator, so `readProtocolMessage` writes a trailing `\n` to `c.out` separately to keep transcript readability.

## Why not a post-read length check on ReadBytes

A post-read check still lets ReadBytes balloon the buffer to the malicious size first. Scanner is the only stdlib primitive that bounds memory **during** the read.

## Tests

- `TestAppServerClient_ReadLineRejectsOversizedLine`: synthesize `maxAppServerLineBytes+1` bytes of `'x'` + newline; assert wrapped `bufio.ErrTooLong` with cap value in message.
- `TestAppServerClient_ReadLineAcceptsLinesBelowCap`: `cap/2` line round-trips intact.

## Verification

- gofmt clean.

Fork CI: https://github.com/xrf-9527/aiops-platform/pull/11

Refs: `docs/superpowers/specs/2026-05-22-codex-line-cap-design.md`.

### DIFF
--- a/docs/superpowers/specs/2026-05-22-codex-line-cap-design.md
+++ b/docs/superpowers/specs/2026-05-22-codex-line-cap-design.md
@@ -1,0 +1,102 @@
+# Codex app-server stdio max-line cap — design
+
+**Date:** 2026-05-22
+**Issue:** [#226](https://github.com/xrf9268-hue/aiops-platform/issues/226)
+
+## Problem
+
+`internal/runner/codex_app_server.go:372` reads each JSON-RPC framed line via `c.reader.ReadBytes('\n')`. `bufio.Reader.ReadBytes` has no built-in line cap and grows its internal buffer until either a newline arrives or memory is exhausted. A misbehaving or malicious Codex app-server that emits a 4 GB line without a newline OOM-kills the worker (and any concurrent agents sharing the host).
+
+SPEC §10.1 RECOMMENDS "Max line size: 10 MB (for safe buffering)". Current code enforces no cap.
+
+## Decision
+
+Replace the `*bufio.Reader` field with a `*bufio.Scanner` configured via `Buffer(make([]byte, 0, 64<<10), maxAppServerLineBytes)` where `maxAppServerLineBytes = 10 * 1024 * 1024`. `bufio.Scanner.Scan` enforces the cap during the read — when a line exceeds the buffer, `Scan` returns `false` and `Err` returns `bufio.ErrTooLong`. Memory stays bounded at the cap.
+
+### Why Scanner and not a post-read length check on ReadBytes
+
+A post-read length check would still let `ReadBytes` allocate the entire (potentially gigabyte) line into the internal buffer before returning. The check rejects the result but the memory damage is already done. Scanner enforces the cap during the read itself.
+
+### Why not `io.LimitReader`-wrap stdout
+
+`io.LimitReader` exhausts after the cap. The Codex app-server session is long-running and emits many lines; a per-stream limit would break the second message. A per-line cap requires Scanner.
+
+### Why 10 MiB exactly (not configurable yet)
+
+SPEC says 10 MB. We adopt the recommendation as a constant. If operators report legitimate Codex responses above this (e.g. very long tool result payloads), exposing `codex.max_line_bytes` becomes a follow-up. The error path explicitly surfaces the cap value so the failure mode is diagnosable.
+
+## What changes
+
+| File | Change |
+| --- | --- |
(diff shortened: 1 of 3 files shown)


hey-yvan:
Fork CI: https://github.com/xrf-9527/aiops-platform/pull/11

@codex review

hey-yvan:
Addressed the P2 off-by-one in 94b6164: Scanner buffer cap is now maxAppServerLineBytes+1, so a 10 MiB line passes and only strictly-larger lines hit ErrTooLong. Test renamed to ReadLineAcceptsLineAtCap to pin the boundary.

@codex review

PR #257, opened 2026-05-22
fix(workflow-runtime): dedupe EventWorkflowReloadFailed per fingerprint (#239)

### LINKED ISSUE #239: [P3][bug] WorkflowRuntime emits EventWorkflowReloadFailed every tick (default 1s) on persistent reload failure; no dedupe by fingerprint
## Summary

When `WORKFLOW.md` is present but invalid (YAML parse error, validation failure, etc.), `RunWorkflowReloadLoop` emits `EventWorkflowReloadFailed` on every tick — default interval `1s`. If the operator's invalid workflow stays uncorrected for an hour, the event stream collects 3,600 identical failure events. The runtime event ring is bounded to 200 (status.go) so older entries evict, but the emission itself is wasted work and floods any structured-log sink with duplicates.

## Why it happens

`internal/orchestrator/workflow_runtime.go:96-119` (`ReloadOnce`):

```go
fingerprint, err := workflowFileFingerprint(r.path)
if err != nil {
    r.emit(ctx, task.EventWorkflowReloadFailed, ...)
    return err
}
if snap := r.Current(); snap.Fingerprint != "" && snap.Fingerprint == fingerprint {
    return nil
}
wf, err := workflow.Load(r.path)
if err == nil && r.validate != nil {
    err = r.validate(wf.Path, wf.Source, wf.Config)
}
if err != nil {
    r.emit(ctx, task.EventWorkflowReloadFailed, ...)
    return err
}
r.current.Store(r.snapshotFromWorkflow(wf, fingerprint))
```

The fingerprint check (line 109) early-returns when `snap.Fingerprint == fingerprint`. But the snapshot is only `Store`d on **success** (line 119). When `Load` or `validate` fails, the snapshot retains the prior fingerprint (initial empty `""` if no successful load ever happened, otherwise the last good fingerprint).

So:
- File is invalid from the start → `snap.Fingerprint == ""`, fingerprint differs every tick → re-attempt + re-emit every tick.
- File was valid, then becomes invalid → `snap.Fingerprint` is the last-good fingerprint, current file's fingerprint is the new (invalid) one → they differ → re-attempt + re-emit every tick.

Only when the file becomes valid does `Store` capture the new fingerprint and the loop quiesces.

`RunWorkflowReloadLoop` (line 274):

```go
for {
    if err := runtime.ReloadOnce(ctx); err != nil && ctx.Err() != nil {
        return ctx.Err()
    }
    ...
    if err := sleep(ctx, interval); err != nil {
        return err
    }
}
```

`ReloadOnce` returning an error is observed but not used to skip the next tick or apply backoff.

## Why this matters

1. **Log/event flooding.** `EventWorkflowReloadFailed` fires every second for the duration of the misconfiguration. An operator with structured-log alerts (e.g., loki/datadog) wakes up to thousands of identical alerts.
2. **Event ring eviction.** RuntimeEvents (`internal/orchestrator/status.go`) caps at 200; reload-failed events evict everything else useful (recent retries, recent successes). `/api/v1/state` shows only reload-failed entries, hiding real-world activity.
3. **Wasted work.** Every tick: fingerprint(file) + Load(file) + validate. For a 100 KB file on tmpfs the cost is small but compounding.

## Suggested fix

Track the last-failed fingerprint separately so we can skip identical failures:

```go
type WorkflowRuntime struct {
    ...
    lastFailedFingerprint string
    lastFailedAt          time.Time
}

func (r *WorkflowRuntime) ReloadOnce(ctx context.Context) error {
    fingerprint, fpErr := workflowFileFingerprint(r.path)
    if fpErr != nil {
        if r.lastFailedFingerprint == "<read-error>" {
            return fpErr // already emitted; skip
        }
        r.lastFailedFingerprint = "<read-error>"
        r.emit(ctx, task.EventWorkflowReloadFailed, ...)
        return fpErr
    }
    if snap := r.Current(); snap.Fingerprint != "" && snap.Fingerprint == fingerprint {
        return nil
    }
    if r.lastFailedFingerprint == fingerprint {
        return errors.New("workflow reload skipped: identical to last-failed snapshot")
    }
    wf, err := workflow.Load(r.path)
    if err == nil && r.validate != nil {
        err = r.validate(wf.Path, wf.Source, wf.Config)
    }
    if err != nil {
        r.lastFailedFingerprint = fingerprint
        r.emit(ctx, task.EventWorkflowReloadFailed, ...)
        return err
    }
    r.lastFailedFingerprint = ""
    r.current.Store(r.snapshotFromWorkflow(wf, fingerprint))
    r.emit(ctx, task.EventWorkflowReloaded, ...)
    return nil
}
```

Now: identical invalid file → emit once, then quiesce until the file changes.

Alternative: rate-limit `EventWorkflowReloadFailed` emissions to once per minute regardless of fingerprint.

## Acceptance criteria

- [ ] Persistent invalid workflow emits `EventWorkflowReloadFailed` once per fingerprint, not once per tick.
- [ ] The first emission and the eventual recovery are both visible in the event stream.
- [ ] Test in `internal/orchestrator/workflow_reloader_test.go` simulates 10 ticks against an invalid file and asserts only one `EventWorkflowReloadFailed` is emitted.

## Refs

- Code: `internal/orchestrator/workflow_runtime.go:96-119, 274-298`
- SPEC §6.2 ("Invalid reloads MUST NOT crash the service; keep operating with the last known good effective configuration and emit an operator-visible error.") — operator-visible error, yes, but the frequency is undefined.

hey-yvan:
Closes #239.

## Problem

`WorkflowRuntime.ReloadOnce` emitted `EventWorkflowReloadFailed` every tick (default 1s) when the workflow file was present but invalid — 3,600 identical emissions per hour. The 200-entry runtime event ring evicted everything else, structured-log sinks alerted thousands of times, and the loop re-ran `workflow.Load` + `validate` on identical content every tick.

The existing fingerprint check (`workflow_runtime.go:108`) only short-circuits re-emission of `EventWorkflowReloaded` — it compares against the last *successful* fingerprint, which is empty when no good load has ever happened, so the check never matched for an invalid file.

## Change

- `internal/orchestrator/workflow_runtime.go`: add `failureMu sync.Mutex` + `lastFailedFingerprint string` on `WorkflowRuntime`. ReloadOnce now:
  - Read fails → sentinel `"<workflow-read-error>"`; subsequent identical read-error skips emit.
  - Snap fingerprint matches file → clear lastFailed; return nil (unchanged).
  - Fingerprint matches lastFailed → return `errWorkflowReloadDeduped` *without* re-running `Load`/`validate` and *without* re-emit.
  - Load/validate fails → set lastFailed = fingerprint, emit, return err.
  - Load/validate succeeds → clear lastFailed, store snap, emit reloaded.

## Tests

- `TestWorkflowRuntimeReloadOnceDedupesIdenticalFailures`: invalid → 6 identical ticks (1 emission) → fix → 1 reloaded → different invalid (1 new emission). Asserts exact counts.
- `TestWorkflowRuntimeReloadOnceDedupesMissingFile`: delete file, 6 ticks, exactly 1 emission via read-error sentinel.

## Verification

- gofmt clean.

Fork CI: https://github.com/xrf-9527/aiops-platform/pull/12

Refs: `docs/superpowers/specs/2026-05-22-reload-dedupe-design.md`.

### DIFF
--- a/docs/superpowers/specs/2026-05-22-reload-dedupe-design.md
+++ b/docs/superpowers/specs/2026-05-22-reload-dedupe-design.md
@@ -1,0 +1,110 @@
+# WorkflowReloadFailed fingerprint dedupe — design
+
+**Date:** 2026-05-22
+**Issue:** [#239](https://github.com/xrf9268-hue/aiops-platform/issues/239)
+
+## Problem
+
+`WorkflowRuntime.ReloadOnce` emits `EventWorkflowReloadFailed` on every tick whenever the workflow file is present but invalid. Default tick = 1 s; an hour of bad config = 3600 identical emissions. The 200-entry runtime event ring (`internal/orchestrator/status.go`) evicts everything useful, structured-log sinks alert thousands of times, and the reload-loop wastes CPU re-running `Load + validate` for content that hasn't changed.
+
+The fingerprint check at `workflow_runtime.go:108` only suppresses re-emission of `EventWorkflowReloaded` — it short-circuits when the **last successful** fingerprint matches the current file. When the file is invalid, no successful snapshot is stored, so the comparison never matches and the load+validate runs every tick.
+
+## Decision
+
+Track the most recent **failed** fingerprint separately. Suppress `EventWorkflowReloadFailed` and the load/validate work when the fingerprint hasn't changed since the last emission. Clear the failed-fingerprint state on the next successful load.
+
+Treat fingerprint-extraction failures (file missing, EISDIR, permission denied) with a sentinel string so they dedupe the same way as load/validate failures.
+
+### Why fingerprint-keyed and not time-rate-limited
+
+Rate limiting (e.g. once per minute) would still emit many duplicates per day and gives no signal about whether the underlying file actually changed. Fingerprint-keying means **operators see exactly one event per distinct broken configuration** — the cardinality reflects real operator activity, not wall-clock time.
+
+### Concurrency
+
+`ReloadOnce` has exactly one production caller (`RunWorkflowReloadLoop` in `cmd/worker/main.go:206`), sequential. The new field is still guarded by a `sync.Mutex` for safety in tests (which call `ReloadOnce` directly from the test goroutine and may run with `-race`).
+
+### Sentinel for read-error case
+
+`workflowFileFingerprint` returns an error before any hex digest is computed when the file is missing/unreadable. We use the literal `"<workflow-read-error>"` as the fingerprint marker for this case. SHA-256 hex is `[0-9a-f]{64}`, so the angle-bracket sentinel cannot collide with a real fingerprint.
+
+## What changes
+
+| File | Change |
+| --- | --- |
+| `internal/orchestrator/workflow_runtime.go` | Add `mu sync.Mutex` + `lastFailedFingerprint string` fields on `WorkflowRuntime`. |
+| `internal/orchestrator/workflow_runtime.go` (`ReloadOnce`) | After fingerprint read: if read fails, compare against sentinel and skip re-emit if matched; otherwise compare current file fingerprint against `lastFailedFingerprint` and skip both `workflow.Load` and re-emit when matched. On any successful path (snap match or new snap stored), clear `lastFailedFingerprint`. |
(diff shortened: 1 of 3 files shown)


hey-yvan:
Fork CI: https://github.com/xrf-9527/aiops-platform/pull/12

@codex review

PR #258, opened 2026-05-22
fix(orchestrator): honor configured terminal_states without hardcoded overlay (#232)

### LINKED ISSUE #232: [P2][bug] filterEligibleCandidates unions a hardcoded terminal-state set with the configured one; operators cannot opt issues out of the hardcoded list
## Summary

`internal/orchestrator/poller.go:313-317`:

```go
func filterEligibleCandidates(issues []tracker.Issue, terminalStates []string) []tracker.Issue {
    terminal := normalizedStates(terminalStates)
    for state := range normalizedStates([]string{"Done", "Canceled", "Cancelled", "Closed", "Duplicate"}) {
        terminal[state] = struct{}{}
    }
    ...
}
```

A hardcoded set `{done, canceled, cancelled, closed, duplicate}` is **unioned with** the operator-configured `terminal_states`. The configured set can only **add** terminal states; it cannot **remove** any of the hardcoded ones.

This violates SPEC §5.3.1, which makes `terminal_states` fully operator-configurable with a *default* of `["Closed", "Cancelled", "Canceled", "Duplicate", "Done"]`. A workflow that explicitly sets, say, `terminal_states: ["Cancelled", "Done"]` to allow re-dispatch of `Duplicate` issues still has them filtered as terminal by the hardcoded overlay.

## Evidence

Reproducer: a `WORKFLOW.md` with

```yaml
tracker:
  terminal_states:
    - Done
    - Cancelled
```

would intuitively permit `Closed` and `Duplicate` to flow through dispatch. The current code still excludes them. The contradiction is invisible to the operator unless they read the source.

The same hardcoded overlay also runs inside `todoIssueBlockedByOpenDependency` via the shared `terminal` map (line 323): a blocker whose state is `Closed` blocks `Todo` even if the operator's terminal_states list does not include `Closed`. Cross-tracker portability suffers — a Linear workflow with custom states like `Released` cannot use it as a terminal blocker-pass-through without also adding the hardcoded English set.

## Why this matters

1. **SPEC §5.3.1 default vs override.** SPEC distinguishes the *default* (which the hardcoded list happens to match) from the *configured* set. By unioning, the project promotes the default to a floor that cannot be lowered.
2. **Workflows with non-English state names break.** A team using `完了` (Done in Japanese), `Hecho` (Spanish), or `Закрыто` (Russian) as their terminal-equivalent must specify the English fallback too — even if the tracker never returns those values — to avoid the overlay surprising them in some other place.
3. **`Todo` blocker semantics.** The shared `terminal` map flows into the blocker check (poller.go:347). An issue with a blocker in `Closed` state should be eligible IF the operator's config says `Closed` is **not** terminal — perhaps they use it as an in-progress state. The overlay blocks that.

## Suggested fix

Drop the hardcoded overlay; trust the configured `terminal_states`. The SPEC default already populates the same English set when the workflow omits the field (see `internal/workflow/config.go` DefaultConfig).

```go
func filterEligibleCandidates(issues []tracker.Issue, terminalStates []string) []tracker.Issue {
    terminal := normalizedStates(terminalStates)
    // (Remove the unconditional overlay)
    ...
}
```

If there is concern that some operators may rely on the overlay as a safety net (e.g., they typo'd their config and want the system to still recognize `Done`), add a log warning at config load time when the configured terminal_states does NOT include the SPEC default — but do not silently overlay.

Test:

- Workflow with `terminal_states: ["Released"]` and an issue in state `Closed`. Today: filtered as terminal. After fix: passes the filter (and reaches dispatch, where SPEC §8.2 active-state check correctly excludes it).
- Workflow with `terminal_states: ["Done"]` and a `Todo` issue blocked by a `Closed` blocker. Today: blocker counted as terminal → issue eligible. After fix: blocker not terminal → issue blocked. Operator chose this; honor it.

## Acceptance criteria

- [ ] Hardcoded overlay removed from `filterEligibleCandidates`.
- [ ] Existing tests still pass (defaults populate the same English set via `DefaultConfig`).
- [ ] New test: explicit `terminal_states: [Released]` does not filter `Closed`.
- [ ] New test: `Todo` blocker rule observes only the operator-configured terminal_states.
- [ ] DEVIATIONS.md row noted if the overlay was deliberate harness-engineering.

## Refs

- Code: `internal/orchestrator/poller.go:313-329`
- SPEC §5.3.1 (terminal_states default + override semantics), §8.2 (Todo blocker rule)

hey-yvan:
Closes #232.

## Problem

`filterEligibleCandidates` unioned the operator-configured `terminal_states` with a hardcoded English set `{Done, Canceled, Cancelled, Closed, Duplicate}`. Operators could only **add** terminal states; they could never **remove** any of the hardcoded five. Same merged set flowed into the Todo blocker rule, so blocker semantics also ignored explicit subset configs. SPEC §5.3.1 makes `terminal_states` fully operator-configurable with the 5-state list as a *default*, not a floor.

## Change

Two coordinated changes:

1. `internal/orchestrator/poller.go`: drop the hardcoded union in `filterEligibleCandidates`. Trust the configured `terminal_states`.
2. `internal/workflow/config.go`: expand `DefaultConfig.Tracker.TerminalStates` from `["Done", "Canceled"]` to SPEC §5.3.1 default `["Done", "Canceled", "Cancelled", "Closed", "Duplicate"]`. Preserves back-compat for omit-config deployments via the proper "default config" channel.

After: explicit `terminal_states: ["Released"]` honored exactly; Todo with `Closed` blocker treated as OPEN per operator config.

## Tests

- `TestPollOnceFiltersTodoIssuesBlockedByNonTerminalBlockers`: now sets `poller.reconcile.TerminalStates = ["Done"]` to make the expected terminal set explicit instead of relying on the dropped overlay.
- `TestPollOnceTreatsDefaultSpecTerminalBlockersAsUnblocked`: now uses `workflow.DefaultConfig().Tracker.TerminalStates` so the test actually exercises the SPEC 5-state default it claims to.
- **New** `TestPollOnceTodoBlockerHonorsOperatorConfiguredTerminalStates`: operator subsets to `["Done"]`; Closed/Duplicate blockers NOT treated as terminal; both Todo issues stay blocked.
- **New** `TestFilterEligibleCandidatesUsesOnlyConfiguredTerminalSet`: operator config `["Released"]`; Todo with Closed blocker filtered; Todo with Released blocker passes.

## Verification

- gofmt clean.

Fork CI: https://github.com/xrf-9527/aiops-platform/pull/13

Refs: `docs/superpowers/specs/2026-05-22-terminal-states-overlay-design.md`.

### DIFF
--- a/docs/superpowers/specs/2026-05-22-terminal-states-overlay-design.md
+++ b/docs/superpowers/specs/2026-05-22-terminal-states-overlay-design.md
@@ -1,0 +1,65 @@
+# Drop hardcoded terminal-state overlay — design
+
+**Date:** 2026-05-22
+**Issue:** [#232](https://github.com/xrf9268-hue/aiops-platform/issues/232)
+
+## Problem
+
+`filterEligibleCandidates` (`internal/orchestrator/poller.go:317`) unions the operator-configured `terminal_states` with a hardcoded English set `{Done, Canceled, Cancelled, Closed, Duplicate}`. Operators can only **add** terminal states; they can never **remove** any of the hardcoded five. The same merged set flows into the `Todo` blocker rule (`todoIssueBlockedByOpenDependency`), so blocker semantics also ignore explicit subset configs.
+
+This violates SPEC §5.3.1, which makes `terminal_states` fully operator-configurable with a *default* of `["Closed", "Cancelled", "Canceled", "Duplicate", "Done"]` — a default, not a floor.
+
+## Decision
+
+Two coordinated changes:
+
+1. **Drop the overlay** in `filterEligibleCandidates`. Trust the configured `terminal_states`.
+2. **Expand `DefaultConfig().Tracker.TerminalStates`** from `["Done", "Canceled"]` to the SPEC §5.3.1 default `["Done", "Canceled", "Cancelled", "Closed", "Duplicate"]`. This preserves back-compat for omit-config deployments — they still see the same 5-state coverage they got via the overlay, just through the proper "default config" channel instead of an unconditional union.
+
+After this PR:
+
+- An explicit `terminal_states: ["Released"]` is honored exactly. `Closed` issues now flow through dispatch (where SPEC §8.2 active-state check still excludes them if Closed isn't an active state).
+- An omit-config deployment (no `terminal_states` in `WORKFLOW.md`) gets the SPEC 5-state default — behavior unchanged.
+- The `Todo` blocker rule observes only the operator-configured terminal_states.
+
+### Why both changes (and not just dropping the overlay)
+
+The issue body's literal fix is "drop the overlay" and asserts that "the SPEC default already populates the same English set when the workflow omits the field." But `DefaultConfig` currently has only 2 states (`Done, Canceled`). Dropping the overlay alone would silently regress omit-config deployments: a `Todo` issue blocked by a `Closed` blocker would suddenly stay blocked because Closed is no longer considered terminal under the 2-state default. Expanding the default to the SPEC 5-state set fixes the underlying SPEC-alignment gap and prevents that regression.
+
+### Why not deprecate `DefaultConfig` and require explicit config
+
+That's a much bigger change with operator-visible breakage. Out of scope.
+
+### Why not log a warning when operators omit the default English set
+
+Suggested by the issue body as "additional safety net". Adds noise for legitimate non-English configs (Japanese, Spanish, Russian) and is hard to phrase without sounding paternalistic. Skip.
+
+## What changes
+
+| File | Change |
+| --- | --- |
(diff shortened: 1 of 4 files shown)


hey-yvan:
Fork CI: https://github.com/xrf-9527/aiops-platform/pull/13

@codex review

hey-yvan:
Addressed the P2 NewPoller regression in 135e879: filterEligibleCandidates now falls back to workflow.DefaultConfig().Tracker.TerminalStates when the caller passes nil/empty. Test simplified to exercise the fallback path.

@codex review

hey-yvan:
Addressed the P2 explicit-empty-vs-omitted conflation in fcf15d1: moved the default seeding from filterEligibleCandidates back to NewPoller. NewPollerWithReconciliation now passes the operator's TerminalStates through verbatim; an explicit empty slice disables the blocker rule. New test TestFilterEligibleCandidatesExplicitEmptyTerminalStatesBlocksAll locks the behavior.

@codex review

PR #259, opened 2026-05-22
fix(worker): mask Sandbox.CredentialFiles + Repo.CloneURL userinfo in --print-config (#194)

### LINKED ISSUE #194: [P1][security] `worker --print-config` only masks tracker.api_key; sandbox.credential_files and repo clone URLs leak in clear
## Summary

`internal/worker/print_config.go:126-130` (`maskSecrets`) only redacts `Tracker.APIKey`. The function's own comment acknowledges the gap:

```go
// Currently only Tracker.APIKey is masked — extend this list
// when new secret-bearing fields are added to the schema.
```

But the schema has grown since the comment was added:

- `internal/workflow/config.go:288` — `SandboxConfig.CredentialFiles []string` (paths to files that hold credentials passed to the sandbox).
- `internal/workflow/config.go` — `RepoConfig` (clone URLs that may embed `https://user:token@host/…` auth).
- `Claude.CommandConfig` and any future `MCP` config also potentially carry sensitive command strings.

Operators run `worker --print-config /path/to/clone` and paste the output into issues / chat / runbooks. Today that paste exposes the absolute filesystem locations of credential files (which is a precise instruction for an attacker with shell access), and any auth embedded in repo clone URLs.

## Evidence

- `internal/worker/print_config.go:124-131` — gap acknowledged in the comment, never extended.
- `internal/workflow/config.go:281-289` (`SandboxConfig`) — `CredentialFiles []string` exposed verbatim via `Sandbox: cfg.Sandbox` in `configView` (`internal/worker/print_config.go:38-48`).
- `configView` does not even include `Sandbox` (line 38-48) — meaning the sandbox block is omitted entirely from `--print-config` output. That itself is a separate bug: operators cannot see the effective sandbox posture in `--print-config`, defeating the debugging purpose.

So we have two issues at once:
- **Secret leak risk** when fields are added without extending the mask helper.
- **Observability gap** because the current `configView` silently drops `Sandbox` (including `Enabled`, `Backend`, `NetworkMode`) — operators cannot see whether their sandbox config landed.

## Suggested fix

1. Audit every field of `workflow.Config` and tag it as `safe-to-print` or `secret-bearing`. Document the policy in the package comment.
2. Extend `maskSecrets` to handle every secret-bearing field:
   - `Sandbox.CredentialFiles` → mask each entry with `***` (path itself is sensitive).
   - `Repo.CloneURL` → strip embedded basic-auth via `url.Parse` and re-emit without userinfo.
   - Any new field surfacing a token / file path to a credential.
3. Add `Sandbox`, `Claude` (if it has secret-bearing options), and other missing blocks to `configView` so operators can debug them — but only after they are mask-aware.
4. Replace the brittle "extend this list when new fields are added" pattern with a struct-tag-driven walker (e.g. `aiops:"secret"`) so the next schema addition is opt-in rather than opt-out.
5. Test: `internal/worker/print_config_test.go` — for every secret-bearing field, assert the masked value appears in output and the literal value does not.

## Acceptance criteria

- [ ] `maskSecrets` covers every secret-bearing field in the current `workflow.Config`.
- [ ] `configView` includes the previously-omitted `Sandbox` block.
- [ ] Struct tag (or equivalent) makes the policy mechanical for future fields.
- [ ] Tests assert no plaintext secret appears in `--print-config` output for each field type.

## Refs

- Code: `internal/worker/print_config.go:38-48, 119-131`
- Schema: `internal/workflow/config.go:30, 281-289`
- Related: `docs/runbooks/local-dev.md` recommends `--print-config` as a debug command.

hey-yvan:
Closes #194.

## Summary
- Extend `maskSecrets` so `worker --print-config` redacts `Repo.CloneURL` userinfo (via `net/url`) and replaces each `Sandbox.CredentialFiles` entry with `***` while preserving slice length, so operators can still confirm "I configured N credential files, N are loaded". The function's doc-comment is upgraded into an explicit policy contract: adding a new secret-bearing field without extending this function is a review-blocking gap, enforced by per-field tests.
- Add the missing `Sandbox` block to `configView` so `--print-config` actually surfaces the effective `enabled`/`backend`/`network`/`env_allowlist` posture instead of silently dropping it (the observability half of #194).
- Per-field boundary tests: paired URL userinfo edges (user+password / user-only / plain HTTPS / SSH-style); credential_files at slice length 0/1/2; sandbox visible in output.
- Spec doc: `docs/superpowers/specs/2026-05-22-print-config-secret-masking-design.md`.

The struct-tag walker suggested in the issue was deliberately not built — three known secret-bearing fields don't warrant a reflection subsystem, and the enumerated mask with per-field tests is shorter and easier to audit. The audit table in the spec explains every top-level `workflow.Config` field's classification.

CI: runs on the fork PR https://github.com/xrf-9527/aiops-platform/pull/14 (this upstream repo is out of GHA quota).

## Test plan
- `go test ./internal/worker/...` — green, including 3 new test functions covering the new contracts.
- `go vet ./...` — clean.
- `go build ./...` — clean.

Refs: #194

### DIFF
--- a/docs/superpowers/specs/2026-05-22-print-config-secret-masking-design.md
+++ b/docs/superpowers/specs/2026-05-22-print-config-secret-masking-design.md
@@ -1,0 +1,107 @@
+# `worker --print-config` secret masking — design
+
+Tracks issue #194.
+
+## Problem
+
+`internal/worker/print_config.go:124-131` (`maskSecrets`) only redacts `Tracker.APIKey`. Its own comment admits the gap. Since that comment was written, two additional secret-bearing surfaces have landed:
+
+- `workflow.SandboxConfig.CredentialFiles []string` — absolute paths to credential files. The path itself is sensitive ("where the attacker should look on disk if they get shell access").
+- `workflow.RepoConfig.CloneURL string` — may embed `https://user:token@host/…` basic-auth.
+
+Operators paste `--print-config` output into issues / chat / runbooks. Today, that paste leaks these values verbatim.
+
+A second, related problem: `configView` omits `Sandbox` entirely (`internal/worker/print_config.go:38-48`). Operators cannot use `--print-config` to verify the effective sandbox posture (`enabled`, `backend`, `network_mode`, etc.). That defeats the debugging purpose for the sandbox feature.
+
+## Goals
+
+1. `maskSecrets` covers every secret-bearing field on the current `workflow.Config`.
+2. `configView` includes the previously-omitted `Sandbox` block so it is observable from `--print-config` (after masking).
+3. New tests assert no plaintext secret reaches stdout for each secret-bearing field type.
+4. The policy is locally enumerated in a single function with a doc-comment contract — "every secret-bearing field is masked here; adding a new one without updating this function is a review-blocking change". The `aiops:"secret"` struct-tag walker mentioned in the issue is rejected (see below).
+
+## Non-goals
+
+- Building a reflection-based struct-tag walker. The current schema has three secret-bearing fields; a hand-written mask function is shorter, easier to audit, and produces specific diagnostics. A struct-tag walker would add a serializer-shaped subsystem to harden against a hypothetical future where someone adds a secret field without reading the very function they would have to touch anyway. Rejected as over-engineering. The doc-comment + tests around `maskSecrets` are the policy.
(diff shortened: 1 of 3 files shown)


hey-yvan:
@codex review

Issue #73, opened 2026-05-13
[P1][spec-alignment] Remove Postgres queue as scheduler source; fold into orchestrator runtime state (#95)


hey-yvan:
## Goal

Replace the legacy Postgres-backed task queue with SPEC-aligned tracker + filesystem recovery and orchestrator-owned runtime state.

## SPEC alignment

- SPEC restart recovery is tracker-driven and filesystem-driven, without a durable orchestrator DB as scheduler authority.
- SPEC §3.1 / §4.1.8 / §7.4: orchestrator runtime state is the authoritative in-memory scheduling state; see #95.
- `DEVIATIONS.md` D6 identifies the Postgres queue as a reverting deviation and a partial of D21.

## Acceptance criteria

- [ ] `cmd/worker` no longer requires `DATABASE_URL` / Postgres queue tables to claim work.
- [ ] `internal/queue` and `migrations/` are removed or reduced to non-scheduler archival/legacy code only if explicitly justified.
- [ ] Startup recovery is based on tracker state plus deterministic workspace filesystem state, not queue rows.
- [ ] Candidate selection and running/completed state are represented in orchestrator runtime state (#95), not in Postgres as the source of truth.
- [ ] Tests cover restart recovery without Postgres.
- [ ] Docs/runbooks remove Postgres as a required component of the SPEC-aligned path.

## Dependencies

- This is a child/phase of #95.
- Related to #68 startup reconciliation and #93 candidate selection.

hey-yvan:
## Confirmed by the official Elixir reference implementation

This deviation is not just SPEC-text inference — the upstream reference
implementation behaves exactly as SPEC describes. Direct evidence from
[`elixir/lib/symphony_elixir/orchestrator.ex`](https://github.com/openai/symphony/blob/main/elixir/lib/symphony_elixir/orchestrator.ex):

> The orchestrator is a `GenServer` whose `State` struct holds:
> `:poll_interval_ms`, `:max_concurrent_agents`, `:next_poll_due_at_ms`,
> `:tick_timer_ref`, `:tick_token`, `running: %{}`,
> `completed: MapSet.new()`, `claimed: MapSet.new()`,
> `retry_attempts: %{}`, `codex_totals`, `codex_rate_limits`.
>
> **All state lives purely in process memory. The struct holds in-flight
> jobs, retry schedules (with `Process.send_after` timers), and token
> metrics — nothing persists to disk.**
>
[...]

PR #262, opened 2026-05-22
fix(pollers): honor SIGTERM in legacy linear+gitea poll loops (#238)

### LINKED ISSUE #238: [P2][bug] linear-poller and gitea-poller have no SIGTERM handler — kill mid-Enqueue leaves partial DB state
## Summary

`cmd/linear-poller/main.go` and `cmd/gitea-poller/main.go` both run an infinite poll loop without installing a signal handler. On SIGTERM (Compose shutdown, `kubectl rollout restart`, `launchctl unload`), the process exits abruptly:

- If killed mid-`Store.Enqueue`, the `INSERT ... RETURNING` may have committed but the corresponding `AddEvent(EventEnqueued)` was not run. The queue ends with a task row whose lifecycle event log is empty.
- If killed between two issue fetches inside one tick, the partial batch is lost — the next start has no idea where the prior tick stopped (LINEAR/Gitea poller has no cursor persistence).
- `pgxpool.Pool` is closed via `defer pool.Close()` (good), but only when the *Go* runtime reaches end-of-main. SIGTERM bypasses defers in a kill scenario.

D6 marks these pollers as "Reverting" — the worker now owns the SPEC-aligned poll tick — so the fix may be "just delete them". But while they're in tree and still wired into `deploy/docker-compose.yml` under the `legacy-queue` profile, they need to shut down cleanly.

## Evidence

`cmd/linear-poller/main.go:25-62`:

```go
func main() {
    if len(os.Args) < 2 {
        log.Fatal("usage: linear-poller /path/to/WORKFLOW.md")
    }
    wf, err := workflow.Load(os.Args[1])
    ...
    pool, err := pgxpool.New(context.Background(), os.Getenv("DATABASE_URL"))
    ...
    defer pool.Close()
    ...
    for {
        // poll loop — no select on a cancel channel
        ...
        time.Sleep(time.Duration(wf.Config.Tracker.PollIntervalMs) * time.Millisecond)
    }
}
```

```bash
$ grep -n "signal\.\|SIGTERM\|os\.Interrupt" cmd/linear-poller/main.go cmd/gitea-poller/main.go
(no output)
```

No signal handling at all. Same shape in `cmd/gitea-poller/main.go`.

Compare `cmd/worker/main.go:36`:

```go
ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
defer stop()
```

The worker does the right thing; the legacy pollers were never updated.

## Why this matters

1. **Compose/k8s shutdown.** Any production-style stop sends SIGTERM with a typical 10-30s grace period before SIGKILL. Today the poller ignores SIGTERM and gets SIGKILL'd, losing any in-flight enqueue work.
2. **D6 revert ETA.** Even if these pollers are going away, the revert is not landed yet. While they're shipped under the `legacy-queue` Compose profile, their shutdown story matters.
3. **Test reproducibility.** Integration tests that send SIGTERM to the poller (or kill the docker container) cannot rely on a graceful tail of events.

## Suggested fix

Either:

### A. Add minimal SIGTERM handling (low-risk, ~10 lines)

```go
ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
defer stop()

for {
    select {
    case <-ctx.Done():
        log.Printf("linear-poller: shutdown requested; exiting")
        return
    default:
    }
    pollOnce(ctx, pool, wf)
    select {
    case <-time.After(pollInterval):
    case <-ctx.Done():
        return
    }
}
```

### B. Delete the binaries entirely

Close out D6 (#73) by removing `cmd/linear-poller`, `cmd/gitea-poller`, `internal/queue/`, `migrations/`, and the legacy-queue Compose profile. The worker already owns tracker polling for both Linear and Gitea (`tracker.kind: linear` / `tracker.kind: gitea` in WORKFLOW.md).

Option B is the cleaner long-term answer, but until #73 lands, Option A is the responsible bandage.

## Acceptance criteria

- [ ] Either pollers handle SIGTERM (Option A) or are removed (Option B).
- [ ] Any code path that talks to Postgres flushes outstanding writes before exit.
- [ ] `deploy/docker-compose.yml` updated to match whichever option lands.
- [ ] DEVIATIONS.md D6 row reflects the resolution.

## Refs

- Code: `cmd/linear-poller/main.go:25-62`, `cmd/gitea-poller/main.go:25-62`
- Worker reference: `cmd/worker/main.go:36`
- DEVIATIONS.md D6 (#73) — reverting status

hey-yvan:
Closes #238 (option A — minimal SIGTERM handling; option B / D6 removal is its own issue).

## Summary
\`cmd/linear-poller\` and \`cmd/gitea-poller\` ran an unconditional \`for { ... time.Sleep ... }\` against \`context.Background()\`. SIGTERM from Compose shutdown, \`kubectl rollout restart\`, or any process manager bypassed all of it: the runtime ignored the signal, defers never fired, and any in-flight \`pgxpool\` operation was guillotined mid-INSERT, leaving \`tasks\` rows without their corresponding \`task_events\` rows.

These pollers are slated for removal once D6/D7 cleanup completes, but while they're still shipped under the \`legacy-queue\` Compose profile their shutdown story matters. This PR is the responsible bandage.

## Changes
- Both \`main()\`s build their root ctx via \`signal.NotifyContext(... os.Interrupt, syscall.SIGTERM)\`. The same ctx threads through \`pgxpool.New\`, \`pollOnce\`, and the tracker clients, so in-flight calls react to the signal rather than only the wrapper loop.
- New \`runPollLoop(ctx, interval, poll)\` helper duplicated in each command (these two are the only callers and #73 will delete them anyway). Two select statements: a pre-tick guard so a cancel before the first tick never invokes \`poll\`, and a post-tick sleep that wakes on \`time.After\` OR \`ctx.Done()\`.
- Gitea's previous inline poll body extracted as \`pollOnce\` to match linear's signature.

## Boundary coverage
- \`=N\` cap (poll-count=0): \`TestRunPollLoop_CancelBeforeFirstTick\` — cancel before entry, assert poll never invoked. Pins the pre-tick select.
- \`=N+1\` paired edge: \`TestRunPollLoop_CancelAfterFirstTick\` — closure cancels ctx, assert exactly one invocation. Pins the post-tick select.
- \`TestRunPollLoop_PollFunctionReceivesLoopCtx\` — closure sees the loop's ctx (so inner HTTP/DB calls can react to SIGTERM too).

Both poller packages get the same three tests.

CI: runs on the fork PR https://github.com/xrf-9527/aiops-platform/pull/17 (upstream is out of GHA quota).

## Test plan
- \`go test ./cmd/linear-poller/ ./cmd/gitea-poller/\` — green, 6 new tests across both.
- \`go vet ./...\` — clean.
- \`go build ./...\` — clean.

Refs: #238

### DIFF
--- a/cmd/gitea-poller/main.go
+++ b/cmd/gitea-poller/main.go
@@ -7,7 +7,9 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/jackc/pgx/v5/pgxpool"
@@ -26,7 +28,8 @@ func main() {
 	if len(os.Args) < 2 {
 		log.Fatal("usage: gitea-poller /path/to/WORKFLOW.md")
 	}
-	ctx := context.Background()
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
 	wf, err := workflow.Load(os.Args[1])
 	if err != nil {
 		log.Fatal(err)
@@ -49,15 +52,43 @@ func main() {
 	}
 	interval := time.Duration(wf.Config.Tracker.PollIntervalMs) * time.Millisecond
 
+	runPollLoop(ctx, interval, func(ctx context.Context) {
+		pollOnce(ctx, store, &wf.Config, client)
+	})
+	log.Printf("gitea-poller: shutdown requested; exiting")
+}
+
+// pollOnce executes one tick of the gitea poll loop. Extracted from
+// main so the SIGTERM-aware runPollLoop can drive it.
+func pollOnce(ctx context.Context, store enqueuer, cfg *workflow.Config, client *gitea.TrackerClient) {
+	issues, err := client.ListActiveIssues(ctx)
+	if err != nil {
+		log.Printf("gitea poll error: %v", err)
+		return
+	}
+	processIssues(ctx, store, cfg, issues)
+}
+
+// runPollLoop runs poll on the given interval until ctx is cancelled
+// (typically by SIGTERM/SIGINT via signal.NotifyContext). The two
+// select statements gate both the start of each tick and the
+// inter-tick sleep: a cancel that arrives before the first poll never
+// invokes poll, and a cancel that arrives during a sleep wakes up
+// immediately instead of waiting for time.Sleep to return. Tests in
+// main_test.go pin both branches.
+func runPollLoop(ctx context.Context, interval time.Duration, poll func(context.Context)) {
 	for {
-		issues, err := client.ListActiveIssues(ctx)
-		if err != nil {
-			log.Printf("gitea poll error: %v", err)
-			time.Sleep(interval)
-			continue
+		select {
+		case <-ctx.Done():
+			return
+		default:
 		}
-		processIssues(ctx, store, &wf.Config, issues)
-		time.Sleep(interval)
+		poll(ctx)
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(interval):
+		}
 	}
 }
 
--- a/cmd/gitea-poller/main_test.go
+++ b/cmd/gitea-poller/main_test.go
@@ -113,6 +113,51 @@ func serverURL(r *http.Request) string {
 	return "http://" + r.Host
 }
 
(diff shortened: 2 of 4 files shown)


hey-yvan:
@codex review

Issue #150, opened 2026-05-19
Expose §13.7 /api/v1/state HTTP surface so D21 can fully close


hey-yvan:
## Summary

SPEC §13.7 requires the orchestrator to expose its in-memory state via an HTTP endpoint (`/api/v1/state` returning the snapshot). The Go implementation has the snapshot primitive but no HTTP server bound to it. D21 in DEVIATIONS.md (single-source orchestrator state, `#95`) remains **Partial** in part because of this — closing D21 needs §13.7.

## Evidence

- Snapshot primitive exists: `internal/orchestrator/state.go:398-440` (`Snapshot()` returns a `RunningView`).
- HTTP server search returns nothing matching `/api/v1/state` or any handler wrapping `Snapshot()`:
  ```bash
  grep -rn 'http.Server\|http.HandleFunc\|http.ListenAndServe\|api/v1' internal/ cmd/
  ```
- D21 row in DEVIATIONS.md is Partial; PR #117 closed the state-types prerequisite but not the HTTP surface.

## Why this matters

§13.7 isn't cosmetic — it's the only blessed way for operators to inspect orchestrator state in production (Running, Claimed, RetryAttempts, last reconcile time, capacity utilization). Without it, the only observability is logs.

It also unblocks downstream work — monitoring dashboards keyed on running count, etc.

## Suggested shape

Per SPEC §13.7:

- `GET /api/v1/state` returns JSON of the current snapshot. Read path: `Orchestrator.Snapshot()` (already serializes through the actor).
- Bound at a configurable address from `WORKFLOW.md` (probably `http.bind` under a new top-level section).
- No auth in the v1 cut if the deploy posture is private network; gate behind a token if exposed.
- Health probe (`/healthz`) and readiness (`/readyz`) are natural neighbors but out of scope for this issue.

## Acceptance criteria

- [ ] HTTP server wired in `cmd/worker/main.go`, address from front-matter (default behavior documented).
- [ ] Handler returns the snapshot JSON; integration test confirms shape matches the `RunningView` produced by `Snapshot()`.
- [ ] Handler respects context cancellation (graceful shutdown alongside the actor).
- [ ] D21 row in DEVIATIONS.md updated (Partial -> Closed) when this lands AND the remaining D21 acceptance criteria are met.

## Refs

- SPEC §13.7
- D21 in DEVIATIONS.md (`#95`)
- #117 (D21 state types prerequisite)

PR #267, opened 2026-05-22
fix(runner): redact raw Codex turn-failure params from errors + events (#193)

### LINKED ISSUE #193: [P0][security] codex_app_server.go: error messages dump raw `params` map to logs (potential secret leak)
## Summary

Two error paths in `internal/runner/codex_app_server.go` format the raw `msg["params"]` / `params` map directly into error strings that flow into structured logs and runtime event payloads. The Codex app-server `params` can contain arbitrary user-input snippets, tool arguments, MCP elicitation payloads, or future fields the protocol may add — any of which could include secrets, file paths, or PII. Once embedded in an error, the value ends up in:

- the worker's structured log stream (`internal/orchestrator/runtime_events.go` ingestion path),
- `RecordEvent.Message` shown by `/api/v1/state` (#150 surface),
- the `RunTaskError`/task event vocabulary persisted by the worker.

This is a defense-in-depth issue, not a known active leak — the harm depends on what Codex protocol versions happen to put in `params` — but `%v` formatting of an arbitrary map is precisely what `docs/security-posture.md` should not allow.

### Site 1 — `turn/failed` and `turn/cancelled`

`internal/runner/codex_app_server.go:432`:
```go
return fmt.Errorf("%s: %v", method, msg["params"])
```

`params` here is the full protocol-defined payload for a failed/cancelled turn. SPEC §10.4 / §10.6 define the *category names* (`turn_failed`, `turn_cancelled`) — not the requirement to echo the full payload.

### Site 2 — `completedTurnError` fallback

`internal/runner/codex_app_server.go:692`:
```go
if reason == "" {
    reason = fmt.Sprint(params)
}
return fmt.Errorf("turn/completed failed with status %q: %s", status, reason)
```

When the expected reason fields are absent or empty, the entire `params` map is stringified via `fmt.Sprint`, including any nested tool call results or assistant message content.

## Suggested fix

1. Replace both call sites with allow-listed reason extraction:
   - From `params`, pull only the documented Codex fields (`reason`, `message`, `status`, `error_code`).
   - Fall back to a stable placeholder (`"reason unavailable"`) instead of `fmt.Sprint(params)`.
   - If the unredacted payload is genuinely needed for debugging, log it via a separate `log.Debug` sink that operators can opt into (not via the returned `error`).
2. Add a small helper `safeTurnReason(params map[string]any) string` and use it in both spots.
3. Mirror the same redaction discipline to `recordRuntimeMessage` so the JSON event surface does not store the raw payload either.
4. Regression test in `internal/runner/codex_app_server_test.go`: assert that an injected `params` containing a synthetic secret string never appears in the returned `error.Error()`.

## Acceptance criteria

- [ ] `fmt.Errorf("%s: %v", method, msg["params"])` removed; replaced with allow-list extraction.
- [ ] `fmt.Sprint(params)` fallback in `completedTurnError` removed.
- [ ] Test confirming a synthetic secret in `params` does not surface in error strings or runtime event payloads.
- [ ] `docs/security-posture.md` updated to note the redaction discipline for protocol payloads.

## Refs

- Code: `internal/runner/codex_app_server.go:432, 692`
- Related: `docs/security-posture.md` (operator-visible secrets boundary)

hey-yvan:
## Summary

Closes #193.

Two paths in `internal/runner/codex_app_server.go` formatted the raw `msg["params"]` / `params` map directly into returned `error` strings via `%v` / `fmt.Sprint`:

- `turn/failed` and `turn/cancelled` (`fmt.Sprintf(\"%s: %v\", method, msg[\"params\"])`)
- `completedTurnError` fallback when no `reason`/`error`/`message` field was populated (`reason = fmt.Sprint(params)`)

The error string flowed through `RunTaskError`, `RecordEvent.Message` (`/api/v1/state` per #150), and the worker's structured log stream. The corresponding `recordRuntimeMessage` calls additionally normalized the entire `params` map into the runtime-event JSON payload. Codex protocol payloads can carry arbitrary tool output, elicitation snippets, or future fields — `%v`-ing them is exactly what `docs/security-posture.md` should not permit. Defense-in-depth, not a known active leak.

## Changes

- New `safeTurnReason(params)` extracts only allow-listed string fields (`reason`, `error`, `message`, `error_code`) from `params` and any nested `turn` map; falls back to the stable string `\"reason unavailable\"` (no map stringification).
- New `safeTurnFailurePayload(params)` + `recordSafeTurnFailure` helper emit runtime-event payloads built from the same allow-list. `turn_failed`, `turn_cancelled`, and `turn_ended_with_error` payloads now persist only allow-listed fields. `turn/completed` success and unrelated notification paths are unchanged.
- `docs/security-posture.md` \"What is defended today\" lists the redaction discipline so operators know which fields persist vs. drop.

## Tests

- New `TestCodexAppServerRunnerRedactsTurnFailureParams` covers `turn/failed`, `turn/cancelled`, and the `completedTurnError` fallback. Each sub-test injects a synthetic secret under non-allow-listed `params` keys and asserts:
  - `err.Error()` does not contain the secret;
  - `event.Payload` JSON does not contain the secret;
  - allow-listed `reason` (or `\"reason unavailable\"` fallback) still surfaces;
  - the non-allow-listed `leak` field is dropped from the event payload.
- Existing `TestCodexAppServerRunnerTreatsFailedTurnCompletedAsError`, `TestCodexAppServerRunnerTreatsNestedFailedTurnCompletedAsError`, and `TestCodexAppServerRunnerTreatsInterruptedTurnCompletedAsSuccess` continue to pass — reason extraction for documented fields is preserved.
- Full `go test ./...` green.

## Test plan

- [x] `go test ./internal/runner/...`
- [x] `go test ./...`
- [x] `go vet ./...`

### DIFF
--- a/docs/security-posture.md
+++ b/docs/security-posture.md
@@ -127,6 +127,14 @@ The current Go implementation provides these safety controls:
 - operator-visible blocked state for Codex input-required and MCP elicitation
   requests, so non-interactive runs stop and remain claimed instead of burning
   retries silently;
+- allow-listed redaction of Codex `turn/failed`, `turn/cancelled`, and failed
+  `turn/completed` protocol payloads: returned error strings and
+  `runtime_events` JSON payloads carry only the documented
+  `status`/`reason`/`error`/`message`/`error_code` fields (plus the same keys
+  under nested `turn`), with `"reason unavailable"` as fallback. Arbitrary
+  protocol fields — including any tool-call output, elicitation echoes, or
+  future Codex additions — are never embedded in worker error strings,
+  `RecordEvent.Message`, or `/api/v1/state` payloads;
 - branch protection and review gates when configured on the remote repository.
 
 These controls reduce accidental damage and make changes reviewable. They do not
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -886,13 +886,39 @@ func safeTurnReason(params map[string]any) string {
 		sources = append(sources, turn)
 	}
 	for _, source := range sources {
-		for _, key := range []string{"reason", "error", "message", "error_code"} {
-			if v, _ := source[key].(string); strings.TrimSpace(v) != "" {
-				return strings.TrimSpace(v)
-			}
+		if v := extractReasonFields(source); v != "" {
+			return v
 		}
 	}
 	return fallback
+}
+
+// extractReasonFields walks the allow-listed reason keys in a single map and
+// returns the first non-empty string. `error` is special: Codex may serialize
+// it as a string or as a nested object like
+// `{"message": "...", "code": "...", "error_code": "..."}`. Both shapes are
+// supported; nested object lookup only allows the same allow-listed scalar
+// fields.
+func extractReasonFields(source map[string]any) string {
+	for _, key := range []string{"reason", "error", "message", "error_code"} {
+		raw, ok := source[key]
+		if !ok {
+			continue
+		}
+		switch v := raw.(type) {
+		case string:
+			if s := strings.TrimSpace(v); s != "" {
+				return s
+			}
+		case map[string]any:
+			for _, nestedKey := range []string{"message", "reason", "error_code", "code"} {
+				if s, _ := v[nestedKey].(string); strings.TrimSpace(s) != "" {
+					return strings.TrimSpace(s)
+				}
+			}
+		}
+	}
+	return ""
 }
 
 // safeTurnFailurePayload returns a runtime-event payload built from only the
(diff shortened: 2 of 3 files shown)


hey-yvan:
@codex review

hey-yvan:
Addressed in 3f5a8f8: `extractReasonFields` / `copyAllowlistedReasonFields` now handle structured `error` objects (`params.error.message`, `params.turn.error.message`) on top of the original string shape, while still dropping non-allow-listed siblings. Two new sub-tests cover both shapes. @codex review

PR #268, opened 2026-05-22
fix(worker): containment-check workspace deletes via workspace.SafeRemove (#196)

### LINKED ISSUE #196: [P1][security] os.RemoveAll on workspace paths lacks containment check in hook-failure and reconcile cleanup
## Summary

Two production code paths call `os.RemoveAll(path)` against a path constructed upstream without re-validating that the path is non-empty, absolute, and contained in the configured workspace root. If a future refactor or a malformed hook output ever feeds an empty string, `"/"`, or a path outside the workspace root into these call sites, the worker recursively deletes whatever it points at — including, in the empty-string case, the worker's current working directory contents (Go's `os.RemoveAll("")` returns `nil` without doing anything, but `os.RemoveAll(".")` recursively deletes cwd).

This is a defense-in-depth issue: SPEC §9.5 Invariants 2 & 3 require that any path the runner touches stay inside the workspace root, and #185 already tracks the related agent-launch check. The cleanup paths deserve the same guard.

### Site 1 — hook-failure rollback

`internal/worker/runtask.go:165-170` (`removeWorkdirAfterHookFailure`):
```go
func removeWorkdirAfterHookFailure(ctx context.Context, ev EventEmitter, taskID, workdir string, beforeRemove workflow.WorkspaceHook, timeoutMs int, reason string) {
    if err := runWorkspaceHook(...); err != nil { ... }
    if err := os.RemoveAll(workdir); err != nil {
        log.Printf("task %s: remove workspace %s after %s hook failure: %v", taskID, workdir, reason, err)
    }
}
```
`workdir` is whatever the caller passes — no normalization, no containment check.

### Site 2 — reconcile cleanup

`internal/worker/reconcile.go:233-237` (`removeWorkspace`):
```go
func removeWorkspace(ctx context.Context, cfg ReconcileConfig, taskID, path string, issue tracker.Issue, reason string) (bool, error) {
    if err := runWorkspaceHook(...); err != nil { ... }
    if err := os.RemoveAll(path); err != nil {
        return false, fmt.Errorf("remove %s workspace %s: %w", reason, path, err)
    }
    ...
}
```
Same shape. `path` comes from the reconcile enumeration; if the enumeration bug ever crosses out of root (e.g. symlink in the workspace root pointing elsewhere), `os.RemoveAll` follows the directory structure but not symlinks, so a symlinked target is left intact — but the symlink's parent directory deletion can still affect the unintended location's child entries.

## Why this matters

1. Empty-string / "." footgun: defense against a future refactor that loses the path through a struct copy or a nil-issue lookup.
2. Workspace root traversal: if a `before_remove` hook script `cd`s out of cwd and writes a path back via env (already a misuse, but possible), and the reconcile enumeration is ever fed that path, deletion happens outside the root.
3. Symlink hardening: `os.RemoveAll` does not delete the symlink target's directory contents, but it does follow the symlink to remove its children if the path is a directory symlink. A workspace root with a stray symlink can cause off-root deletion of children.

## Suggested fix

1. Add a helper `safeRemoveWorkspace(root, path string) error`:
   - Reject `path == ""`.
   - Normalize via `filepath.Abs`+`filepath.Clean`.
   - Confirm `path` has `root` as a prefix (use `filepath.Rel` and verify the result does not start with `..`).
   - Reject paths that contain a directory-level symlink between `root` and `path` (`filepath.EvalSymlinks` and re-check containment).
   - Then `os.RemoveAll`.
2. Replace both call sites with the helper, passing the configured `workspace.root`.
3. Add unit tests for: empty path → rejected; absolute path outside root → rejected; symlinked subdirectory escape → rejected; valid path → removed.

## Acceptance criteria

- [ ] `safeRemoveWorkspace` helper exists with containment + symlink checks.
- [ ] `runtask.go:removeWorkdirAfterHookFailure` and `reconcile.go:removeWorkspace` use it.
- [ ] Tests cover the rejection paths.
- [ ] Docs (`docs/runbooks/workspace-cache.md`) note the invariant.

## Refs

- Code: `internal/worker/runtask.go:165-170`, `internal/worker/reconcile.go:233-237`
- Related: #185 (cwd containment for agent launch)
- SPEC §9.5 Invariants 2 & 3, §15.2 mandatory filesystem safety

hey-yvan:
## Summary

Closes #196. Two cleanup paths called `os.RemoveAll(path)` against a path that came from upstream without re-validation:

- `internal/worker/runtask.go::removeWorkdirAfterHookFailure` (rollback after `after_create` hook failure)
- `internal/worker/reconcile.go::removeWorkspace` (reconcile-driven cleanup)

A future refactor or malformed hook output could feed an empty string, the process cwd, the workspace root, or a symlink that points outside the root, and the worker would recursively delete whatever the path pointed at. Defense-in-depth per SPEC §9.5 Invariants 2 & 3, §15.2.

## Changes

- New `workspace.SafeRemove(root, path)` helper:
  - rejects empty/whitespace root or path (`ErrSafeRemoveInvalidPath`);
  - absolutizes + cleans both, asserts containment via `filepath.Rel` (`ErrSafeRemoveEscapesRoot`);
  - re-validates after `filepath.EvalSymlinks` on **both** root and path (root resolution avoids tripping over platform-symlinked tempdirs like macOS `/var` → `/private/var`);
  - refuses to delete the workspace root itself;
  - idempotent for missing paths, matching existing `os.RemoveAll` semantics.
- `removeWorkdirAfterHookFailure` and `removeWorkspace` now call `workspace.SafeRemove`. The former gains a `workspaceRoot` parameter passed from `RunTask` (already resolved via `EffectiveWorkspaceRoot`).
- `docs/runbooks/workspace-cache.md` adds a \"Cleanup containment invariant\" section pointing future cleanup code at the helper.

## Tests

`internal/workspace/safe_remove_test.go` covers:

- empty / whitespace root or path
- root-equals-path
- sibling-of-root rejection (verifies the external file is untouched)
- `..` parent-traversal rejection
- symlink escape (creates a root-relative symlink to an external dir, verifies external child is untouched)
- valid path removal
- idempotent missing-path

`go test ./...` green.

## Test plan

- [x] `go test ./internal/workspace`
- [x] `go test ./internal/worker`
- [x] `go test ./...`

### DIFF
--- a/docs/runbooks/workspace-cache.md
+++ b/docs/runbooks/workspace-cache.md
@@ -76,6 +76,26 @@ the worker is deployed:
    worker's main loop (e.g. once per N completed tasks). Until that
    lands, treat cleanup as an out-of-band concern.
 
+## Cleanup containment invariant
+
+All cleanup paths that delete a per-task worktree go through
+`workspace.SafeRemove(root, path)`. The helper refuses to delete a path
+that is not strictly contained under the configured workspace root:
+empty/whitespace input, the root itself, a sibling of the root, a
+`..` traversal, and symlinks whose resolved target points outside the
+root all return `ErrSafeRemoveEscapesRoot` and the on-disk path is left
+untouched. This is a defense-in-depth guard against a future refactor or
+a malformed hook output that could otherwise pass an empty string or
+`/` into `os.RemoveAll` — see SPEC §9.5 Invariants 2 & 3, §15.2.
+
+The two production call sites today are
+`internal/worker/runtask.go::removeWorkdirAfterHookFailure` (rollback
+after `after_create` hook failure) and
+`internal/worker/reconcile.go::removeWorkspace` (reconcile-driven
+cleanup for closed/cancelled issues). Any new cleanup code that touches
+`$WORKSPACE_ROOT` should call `workspace.SafeRemove` rather than
+`os.RemoveAll` directly.
+
 ## Troubleshooting
 
 - `fatal: '<dir>' is not a working tree` printed during task setup is
--- a/internal/worker/reconcile.go
+++ b/internal/worker/reconcile.go
@@ -236,7 +236,7 @@ func removeWorkspace(ctx context.Context, cfg ReconcileConfig, taskID, path stri
 	if err := runWorkspaceHook(ctx, cfg.Emitter, taskID, path, workspace.HookBeforeRemove, cfg.BeforeRemoveHook, cfg.HookTimeoutMillis); err != nil {
 		log.Printf("task %s: before_remove hook failed for %s workspace %s: %v", taskID, reason, path, err)
 	}
-	if err := os.RemoveAll(path); err != nil {
+	if err := workspace.SafeRemove(cfg.WorkspaceRoot, path); err != nil {
 		return false, fmt.Errorf("remove %s workspace %s: %w", reason, path, err)
 	}
 	Emit(ctx, cfg.Emitter, taskID, task.EventReconcileWorkspace, "removed workspace", map[string]any{
--- a/internal/worker/runtask.go
+++ b/internal/worker/runtask.go
@@ -162,11 +162,11 @@ func runWorkspaceHook(ctx context.Context, ev EventEmitter, taskID, workdir stri
 	return err
 }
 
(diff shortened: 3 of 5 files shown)


hey-yvan:
@codex review

PR #271, opened 2026-05-22
fix(worker): tolerate tracker fetch failures in ReconcileStartup per SPEC §8.6 (#222)

### LINKED ISSUE #222: [P2][spec-alignment] §8.6 + §11.4: ReconcileStartup aborts worker startup on tracker fetch failure (SPEC says log+continue)
## Summary

SPEC §8.6 ("Startup Terminal Workspace Cleanup") last bullet:

> If the terminal-issues fetch fails, log a warning and continue startup.

SPEC §11.4 ("Error Handling Contract") confirms:

> Startup terminal cleanup failure: log warning and continue startup.

The Go implementation propagates the fetch error all the way out of `ReconcileStartup`, and `cmd/worker/main.go` treats that as a fatal startup error. A transient tracker outage during boot now prevents the worker from coming up at all — even though SPEC says the worker should boot, log the warning, and start polling normally.

## Evidence

`internal/worker/reconcile.go:71-79`:

```go
activeIssues, err := cfg.Tracker.ListIssuesByStates(ctx, activeStates)
if err != nil {
    return fmt.Errorf("fetch active issues: %w", err)
}
terminalIssues, err := cfg.Tracker.ListIssuesByStates(ctx, terminalStates)
if err != nil {
    return fmt.Errorf("fetch terminal issues: %w", err)
}
```

`cmd/worker/main.go:164-167`:

```go
if err := worker.ReconcileStartup(ctx, reconcileCfg); err != nil {
    worker.LogReconcileError(err)
    return err
}
```

`return err` from `runWorker` propagates to `main` which calls `log.Fatal(err)` (line 39). The worker process exits non-zero.

## Why this matters

1. **Boot failure on transient outage.** A 30-second Linear API blip during worker restart cascades into a permanent service outage until an operator notices and restarts. SPEC's intent is the opposite — the worker boots, logs the warning, and resumes work when the next poll tick succeeds.
2. **Restart-recovery story breaks.** SPEC §14.3 explicitly designs the orchestrator to recover from restart by polling. If startup fails on the first poll, restart recovery is fragile in exactly the moments it matters (post-crash, post-deploy).
3. **Asymmetric with poll-loop behavior.** During normal operation, candidate fetch failure skips the tick (§14.2). At startup, the same failure aborts the worker. SPEC treats them consistently — log + skip + continue.

## Suggested fix

1. In `ReconcileStartup`, treat `ListIssuesByStates` errors as warnings:
   ```go
   activeIssues, err := cfg.Tracker.ListIssuesByStates(ctx, activeStates)
   if err != nil {
       Emit(ctx, cfg.Emitter, taskID, task.EventReconcileEnd, "startup reconcile skipped: active issue fetch failed", map[string]any{"error": err.Error()})
       log.Printf("startup reconcile: fetch active issues failed: %v (continuing without active key set)", err)
       activeIssues = nil
   }
   terminalIssues, err := cfg.Tracker.ListIssuesByStates(ctx, terminalStates)
   if err != nil {
       log.Printf("startup reconcile: fetch terminal issues failed: %v (continuing without terminal cleanup)", err)
       terminalIssues = nil
   }
   ```
2. Preserve the existing safety net: when `terminalIssues == nil` (or empty), the `canRemoveUnknown` gate already keeps unknown workspaces intact. This means we will not delete anything we cannot verify.
3. `cmd/worker/main.go` no longer aborts on `ReconcileStartup` returning a non-nil error — log it and continue.
4. Tests in `internal/worker/reconcile_test.go` cover (a) active-fetch failure path; (b) terminal-fetch failure path; (c) both failing; assert the worker still produces a valid `OrchestratorState` and the dispatcher loop can start.

## Acceptance criteria

- [ ] `ReconcileStartup` no longer returns an error for tracker-fetch failures.
- [ ] Active fetch failure: no workspaces are auto-removed (safety preserved); worker boots normally.
- [ ] Terminal fetch failure: terminal-workspace cleanup skipped; worker boots normally; warning logged.
- [ ] `reconcile_end` event still emitted with a "skipped" status payload.
- [ ] `cmd/worker/main.go` treats reconcile errors as warnings, not as startup failure.
- [ ] Tests cover both fetch-failure branches.

## Refs

- SPEC §8.6, §11.4, §14.2, §14.3
- Code: `internal/worker/reconcile.go:71-79`, `cmd/worker/main.go:164-167`

hey-yvan:
## Summary

Closes #222. SPEC §8.6 and §11.4 require startup reconcile to log + continue on tracker fetch failure, not abort the worker. The Go implementation propagated both \`ListIssuesByStates\` errors out of \`ReconcileStartup\`; \`cmd/worker/main.go\` treated that as a fatal startup error, so a transient tracker outage during boot prevented the worker from coming up at all — even though SPEC says the worker should boot, log the warning, and start polling normally.

## Changes

Tolerate the two tracker fetches inside \`ReconcileStartup\`:

- **Active fetch failure** is the worst case — without the active list we cannot confirm any workspace is safe to delete. Log warning, emit \`reconcile_end\` with \`{status: \"skipped\", reason: \"active_fetch_failed\", error: ...}\`, return nil. Every workspace stays intact; the next poll tick retries.
- **Terminal fetch failure** is non-fatal. Active workspaces still kept; terminal/unknown removal skipped (\`terminalKeys\` empty, \`canRemoveUnknown\` stays false). \`reconcile_end\` carries \`{status: \"partial\", reason: \"terminal_fetch_failed\", error: ...}\`.
- The existing safety-net error (\"tracker returned no active or terminal issues; refusing to remove N existing workspaces\") only fires when terminal fetch **succeeded** — otherwise an empty \`terminalIssues\` is an artifact of failure, not a tracker response.

\`cmd/worker/main.go\` is unchanged: other ReconcileStartup error paths (empty workspace root, missing tracker, missing states, \`os.ReadDir\` failure, safety-net trip with \`terminalFetchOK==true\`) still represent legitimate misconfiguration and should still abort startup. Only the tracker-fetch tolerance lives inside \`ReconcileStartup\`, which is what SPEC §8.6 actually mandates.

## Tests

New (replace the two now-obsolete tests that pinned the old return-error behavior):

- \`TestReconcileStartupTolerantOfActiveFetchFailure\` — active fetch fails → returns nil; both active and terminal workspaces remain; \`reconcile_end\` payload has \`status=skipped reason=active_fetch_failed\`.
- \`TestReconcileStartupTolerantOfTerminalFetchFailure\` — terminal fetch fails → returns nil; active and unknown workspaces remain (no terminal-key confirmation, so nothing deleted); \`reconcile_end\` payload has \`status=partial reason=terminal_fetch_failed\`.
- \`TestReconcileStartupTolerantOfBothFetchFailures\` — both fail → treated as active-failure path; nothing deleted.

Removed:
- \`TestReconcileStartupReturnsTrackerError\` and \`TestReconcileStartupReturnsTerminalFetchError\` — they asserted exactly the bug this issue describes.

Full \`go test ./...\` green.

## Test plan

- [x] \`go test ./internal/worker\`
- [x] \`go test ./...\`

### DIFF
--- a/internal/worker/reconcile.go
+++ b/internal/worker/reconcile.go
@@ -71,13 +71,32 @@ func ReconcileStartup(ctx context.Context, cfg ReconcileConfig) error {
 
 	activeStates := nonEmptyStates(cfg.ActiveStates)
 	terminalStates := nonEmptyStates(cfg.TerminalStates)
+	// SPEC §8.6 / §11.4: transient tracker outages during boot must log a
+	// warning and continue startup, not abort the worker. An active-fetch
+	// failure is the worst case — without the active list we cannot confirm
+	// any workspace is safe to delete — so we emit reconcile_end with
+	// `status: skipped` and return nil, leaving every workspace intact. A
+	// terminal-fetch failure is non-fatal: active workspaces are still kept,
+	// terminal/unknown removal is skipped because `canRemoveUnknown` stays
+	// false when `terminalIssues` is empty.
 	activeIssues, err := cfg.Tracker.ListIssuesByStates(ctx, activeStates)
 	if err != nil {
-		return fmt.Errorf("fetch active issues: %w", err)
+		log.Printf("startup reconcile: fetch active issues failed: %v (SPEC §8.6: log and continue; no cleanup performed)", err)
+		Emit(ctx, cfg.Emitter, taskID, task.EventReconcileEnd, "startup reconciliation skipped", map[string]any{
+			"status": "skipped",
+			"reason": "active_fetch_failed",
+			"error":  err.Error(),
+		})
+		return nil
 	}
 	terminalIssues, err := cfg.Tracker.ListIssuesByStates(ctx, terminalStates)
+	terminalFetchOK := true
+	var terminalFetchErr error
 	if err != nil {
-		return fmt.Errorf("fetch terminal issues: %w", err)
+		log.Printf("startup reconcile: fetch terminal issues failed: %v (SPEC §8.6: log and continue; terminal cleanup skipped)", err)
+		terminalFetchOK = false
+		terminalFetchErr = err
+		terminalIssues = nil
 	}
 	activeKeysForIssue := cfg.ActiveWorkspaceKeys
 	if activeKeysForIssue == nil {
@@ -101,7 +120,7 @@ func ReconcileStartup(ctx context.Context, cfg ReconcileConfig) error {
 	if err != nil {
 		return err
 	}
-	if len(workspaces) > 0 && len(activeIssues)+len(terminalIssues) == 0 {
+	if terminalFetchOK && len(workspaces) > 0 && len(activeIssues)+len(terminalIssues) == 0 {
 		return fmt.Errorf("tracker returned no active or terminal issues; refusing to remove %d existing workspaces", len(workspaces))
 	}
 	var removed, kept int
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

PR #274, opened 2026-05-22
fix: surface codex_app_server_pid through runtime → state → /api/v1 (#243)

### LINKED ISSUE #243: [P2][spec-alignment] §10.4 / §13.7.2: emitted runtime events and snapshot rows lack codex_app_server_pid
## Summary

SPEC §10.4 ("Emitted Runtime Events") requires each event to include, where available:

> - `event` (enum/string)
> - `timestamp` (UTC timestamp)
> - `codex_app_server_pid` (if available)
> - OPTIONAL `usage` map (token counts)

§13.7.2 shows the same identity surfaced through the per-issue endpoint (`session_id`, `pid`, etc.).

The Go runtime never captures the Codex subprocess PID. Looking at `internal/runner/codex_app_server.go`:

- `cmd.Start()` succeeds at `internal/runner/codex_app_server.go:75` but `cmd.Process.Pid` is never read.
- `appServerClient` (lines 146-164) tracks `threadID`, `turnID`, last events — but no PID field.
- `recordRuntimeEvent` (used at lines 238-243 for `session_started`) carries only protocol IDs in its payload.

Consequence: operators with multiple concurrent agent runs cannot map a `running` row in `/api/v1/state` (or a log line, or a stuck workspace) to the host process that owns it. SPEC explicitly calls this out as part of the event/snapshot vocabulary so cross-referencing tracker state, host PID, and workspace path is a one-liner.

## Evidence

```bash
$ grep -n "cmd.Process\|Pid\|pid" internal/runner/codex_app_server.go
75:    if err := cmd.Start(); err != nil {
86:        terminateProcess(cmd)
# (no Pid extraction anywhere)
```

`internal/orchestrator/state.go:107-127` (`RunningEntry` fields) has no `CodexAppServerPID`. SPEC §4.1.6 lists it explicitly:

> - `codex_app_server_pid` (string or null)

## Suggested fix

1. In `CodexAppServerRunner.Run`, capture the PID after `cmd.Start()` and pass it via the `runtimeEventSink` callback (or via a new `PIDSink` field on `RunInput`).
2. Add `CodexAppServerPID` to `LiveSession` (`internal/orchestrator/state.go:40-45`) and propagate through the orchestrator's `dispatch_codex_update` path.
3. Include it in `RuntimeEvent` payloads at minimum for `session_started`, `turn_completed`, `turn_failed`.
4. Surface it in the snapshot views: `apiStateRunning` (`cmd/worker/main.go:480-487`) and `RunningView` (`internal/orchestrator/state.go:462`). Field name per SPEC §4.1.6: `codex_app_server_pid` (snake_case in JSON).
5. Tests in `internal/runner/codex_app_server_test.go` assert the PID is recorded; orchestrator tests assert it round-trips through the snapshot.

Stringly typed in SPEC ("string or null") to allow future PIDs > int64, container handles, etc.; today int is fine on every Linux host but `strconv.Itoa(int)` is the right encoding.

## Acceptance criteria

- [ ] `CodexAppServerPID` field exists on `LiveSession` and `RunningEntry`.
- [ ] Runner emits the PID at `session_started`; orchestrator records it.
- [ ] `/api/v1/state` and `/api/v1/<issue_identifier>` running rows include `codex_app_server_pid` per SPEC §13.7.2 / §4.1.6.
- [ ] Tests cover the round-trip.
- [ ] `docs/runbooks/runtime-status.md` example payload updated.

## Refs

- SPEC §4.1.6, §10.4, §13.7.2
- Code: `internal/runner/codex_app_server.go:75-90`, `internal/orchestrator/state.go:40-127, 462-510`, `cmd/worker/main.go:480-487`
- Related: #209 (other missing snapshot fields — share the same projection refactor)

hey-yvan:
## Summary

Closes #243. SPEC §10.4 + §13.7.2 list \`codex_app_server_pid\` on emitted runtime events and snapshot rows so operators can map a running row (or log line, or stuck workspace) to the host process that owns it. The Go runtime never captured \`cmd.Process.Pid\`, so the PID never appeared in any event payload or snapshot.

## Changes

- **Runner**: capture \`cmd.Process.Pid\` after \`cmd.Start()\` onto \`appServerClient.codexAppServerPID\`; include in \`session_started\` payload when > 0.
- **State**: \`LiveSession.CodexAppServerPID\` (SPEC §4.1.6) + \`RunningView\` / \`BlockedView\` projections. Snapshot copies session field into both.
- **Runtime events**: \`recordSessionFields\` reads the new key via a new \`intField\` helper (Go int / int64 / JSON float64; rejects non-positive).
- **Worker HTTP**: \`apiStateRunning\` + \`apiStateBlocked\` add \`codex_app_server_pid\` (\`omitempty\`); projections populate it.
- **Docs**: \`docs/runbooks/runtime-status.md\` JSON example shows the field on both a running and a blocked row.
- **Drift test**: synthetic fully-populated \`apiStateResponse\` now sets non-zero PIDs (12345 / 67890) so the drift test correctly enforces the documented field's presence — surfaced the runbook gap during implementation.

## Tests

- \`TestCodexAppServerRunnerCapturesSpecRuntimeEvents\` extended: asserts \`session_started\` payload carries a positive \`codex_app_server_pid\`.
- New \`TestRecordRuntimeEventPropagatesCodexAppServerPID\`: dispatching a session_started event with \`codex_app_server_pid: 42424\` populates \`RunningView.CodexAppServerPID\` end-to-end.
- Drift test passes with the runbook example carrying the new field.
- Full \`go test ./...\` green.

## Test plan

- [x] \`go test ./internal/runner -run TestCodexAppServerRunnerCapturesSpecRuntimeEvents\`
- [x] \`go test ./internal/orchestrator -run TestRecordRuntimeEventPropagatesCodexAppServerPID\`
- [x] \`go test ./cmd/worker -run TestRuntimeStatusRunbookExampleMatchesHandler\`
- [x] \`go test ./...\`


hey-yvan:
@codex review

hey-yvan:
Addressed in f7f5c30: `buildCodexAppServerCmd` now returns a `directCodexExec bool` (true only for direct `codex ...` invocations), and the runner additionally checks `sandbox.Enabled` before `applySandbox` runs. PID is captured only when both are safe; otherwise the field stays zero and `omitempty` makes it absent in JSON — operators see nothing rather than a wrong PID. New test pins the negative (custom-command) path; existing tests pin the positive path. @codex review

PR #277, opened 2026-05-22
fix(orchestrator): fold active-session elapsed time into seconds_running per SPEC §13.5 (#253)

### LINKED ISSUE #253: [P3][spec-alignment] §13.5: codex_totals.seconds_running snapshot does not fold in active-session elapsed time (SHOULD live aggregate)
## Summary

SPEC §13.5 ("Session Metrics and Token Accounting") specifies, for runtime
accounting:

> Runtime accounting:
> - Runtime SHOULD be reported as a live aggregate at snapshot/render time.
> - Implementations MAY maintain a cumulative counter for ended sessions and
>   add active-session elapsed time derived from `running` entries (for
>   example `started_at`) when producing a snapshot/status view.
> - Add run duration seconds to the cumulative ended-session runtime when a
>   session ends (normal exit or cancellation/termination).

The Go implementation maintains the ended-session counter, but the
snapshot returns it verbatim instead of folding in active-session
elapsed time.

## Evidence

`internal/orchestrator/state.go:555-617` (`Snapshot`) copies
`s.CodexTotals` directly into `view.CodexTotals`:

```go
CodexTotals: s.CodexTotals,
```

`s.CodexTotals.SecondsRunning` is only incremented when a run ends, via
`AddSeconds` (line 90, called from the worker-exit finalization path).
There is no per-snapshot accumulation of `time.Since(running.StartedAt)`
for currently-active runs.

Consequence: while a session is in flight, `/api/v1/state.codex_totals.seconds_running` reports zero (or the
last ended-session sum). A dashboard polling `/api/v1/state` every few
seconds sees a flat number until the run ends, which contradicts SPEC's
"live aggregate at snapshot/render time" recommendation.

## Why this matters

1. **Dashboard UX.** Operators expect a smoothly increasing runtime
   counter while a long-running Codex session works through a turn. The
   current behavior shows zero for the whole session and then a sudden
   jump.
2. **Token-vs-runtime sanity check.** When tokens are accumulating but
   runtime is flat, an operator might misdiagnose token burn or
   misidentify a stalled session.
3. **SHOULD-level SPEC text.** §13.5 explicitly uses "SHOULD" for the
   live aggregation; a Symphony port that doesn't do it owes a
   DEVIATIONS row or a fix.

## Suggested fix

Inside `Snapshot()`, before the loop populating `view.Running`, compute the
active-session contribution:

```go
totals := s.CodexTotals
now := view.GeneratedAt
for _, r := range s.Running {
    if !r.StartedAt.IsZero() {
        totals.SecondsRunning += now.Sub(r.StartedAt).Seconds()
    }
}
view.CodexTotals = totals
```

`view.GeneratedAt` is already populated at line 557, so the math is
consistent within a single snapshot.

Tests in `internal/orchestrator/state_test.go` should:
- Verify a snapshot taken mid-run has `seconds_running > 0` even though
  no run has ended yet.
- Verify the ended-session counter and the active-session contribution
  do not double-count when a run finishes between two snapshots (the
  ended-session counter increments only on session-end; the snapshot
  contribution stops once the running entry is removed).

## Acceptance criteria

- [ ] `Snapshot()` folds active-session elapsed time into
      `CodexTotals.SecondsRunning` per SPEC §13.5.
- [ ] Math uses `view.GeneratedAt` so all running entries are measured
      against the same instant.
- [ ] Unit test: snapshot with one running entry started 10s ago and
      no ended sessions reports `seconds_running ≈ 10`.
- [ ] Unit test: no double-counting after the running entry transitions
      to completed.
- [ ] `docs/runbooks/runtime-status.md` documents the live-aggregation
      semantic so dashboards do not subtract the snapshot value from the
      previous one.

## Refs

- SPEC §13.5 Runtime accounting
- Code: `internal/orchestrator/state.go:555-617`, `internal/orchestrator/state.go:86-91` (`AddSeconds`)
- Related: #209 (running row token fields — same projection refactor surface)

hey-yvan:
## Summary

Closes #253. SPEC §13.5 Runtime accounting says runtime SHOULD be reported as a live aggregate at snapshot/render time, with implementations adding active-session elapsed time (derived from \`running.started_at\`) to the cumulative ended-session counter. The Go implementation maintained the ended-session counter (\`AddSeconds\` on worker exit) but copied \`s.CodexTotals\` verbatim, so a dashboard polling \`/api/v1/state\` every few seconds saw a flat \`seconds_running\` while a long Codex turn ran and a sudden jump on session end.

## Changes

\`Snapshot()\` now precomputes a local \`totals\` copy of \`s.CodexTotals\`, walks \`s.Running\`, and adds \`now.Sub(r.StartedAt).Seconds()\` for each entry with a non-zero \`StartedAt\`. The math uses the snapshot's \`GeneratedAt\` (captured once at the top) so all running entries are measured against one instant. No double-count: a finished run is removed from \`s.Running\` before \`AddSeconds\` runs for it.

## Tests

- \`TestSnapshotFoldsLiveSessionRuntimeIntoCodexTotals\`: entry with \`StartedAt = now-10s\` → \`seconds_running ∈ [9.5, 12]\`.
- \`TestSnapshotLiveAggregateDoesNotDoubleCountAfterFinish\`: 5s run finishing between snapshots contributes exactly once.
- \`TestSnapshotLiveAggregateIgnoresZeroStartedAt\`: zero \`StartedAt\` skipped (no Unix-epoch contribution).
- Existing \`TestSnapshot_ShapeMatches13_3\` widened to \`[0.5, 5]\` to accept the new live aggregate.
- Full \`go test ./...\` green.

## Docs

\`docs/runbooks/runtime-status.md\` adds a \`codex_totals.seconds_running\` semantics section warning dashboard authors that consecutive snapshots' values must not be delta-encoded and explaining the no-double-count invariant.

## Test plan

- [x] \`go test ./internal/orchestrator -run TestSnapshot\`
- [x] \`go test ./cmd/worker -run TestRuntimeStatusRunbookExampleMatchesHandler\`
- [x] \`go test ./...\`

### DIFF
--- a/docs/runbooks/runtime-status.md
+++ b/docs/runbooks/runtime-status.md
@@ -172,6 +172,28 @@ the shape here without updating the handler — or vice versa — fails the buil
 IDs in those sets; for lifetime totals across FIFO eviction use the `_total`
 counters.
 
+### `codex_totals.seconds_running` semantics
+
+`seconds_running` is a **live aggregate** per SPEC §13.5 Runtime
+accounting (#253): the snapshot folds the elapsed time of every
+currently-running entry into the cumulative ended-session counter. The
+math uses `generated_at` so all running entries are measured against
+the same instant. Dashboards polling `/api/v1/state` every few seconds
+see a smoothly increasing counter while a long Codex turn works,
+rather than a flat number followed by a sudden jump on session end.
+
+Two consequences for dashboard authors:
+
+- Do **not** treat consecutive snapshots' `seconds_running` as a
+  delta-encoded stream: snapshot N+1 already includes the elapsed
+  time between the two snapshots for any still-running entries.
+  Subtracting snapshot N from snapshot N+1 would double-count.
+- A run that ends between two snapshots adds its elapsed time
+  exactly once. The finished entry is removed from the running set
+  before the ended-session counter increments, so the live aggregate
+  for that entry stops contributing at the same instant the
+  cumulative counter starts including it.
+
 ### Top-level metadata fields
 
 - `generated_at` — RFC3339 timestamp the handler stamped when materializing the snapshot.
--- a/internal/orchestrator/state.go
+++ b/internal/orchestrator/state.go
@@ -688,8 +688,27 @@ type RetryView struct {
 // order is unspecified in Go, but downstream consumers (the §13.7 HTTP
 // handler, CLI status) sort by IssueID before display.
 func (s *OrchestratorState) Snapshot() StateView {
+	now := time.Now().UTC()
+	// SPEC §13.5 Runtime accounting: report runtime as a live aggregate at
+	// snapshot/render time. CodexTotals.SecondsRunning increments only on
+	// session end (AddSeconds in the worker-exit finalization path); without
+	// the per-snapshot fold a dashboard polling /api/v1/state would see a
+	// flat counter while a long run is in flight and a sudden jump on exit.
+	// Add elapsed time for each currently-running entry, measured against the
+	// snapshot's GeneratedAt so all rows share one instant. The ended-session
+	// counter and the live aggregate cannot double-count because a finished
+	// run is removed from s.Running before AddSeconds runs for it.
+	totals := s.CodexTotals
+	for _, r := range s.Running {
+		if r.StartedAt.IsZero() {
+			continue
+		}
+		if elapsed := now.Sub(r.StartedAt); elapsed > 0 {
+			totals.SecondsRunning += elapsed.Seconds()
+		}
+	}
 	view := StateView{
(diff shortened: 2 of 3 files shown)


hey-yvan:
@codex review

hey-yvan:
Addressed in 3abe4c9: captured monotonic `now := time.Now()` for the elapsed math; UTC is now applied only when populating `GeneratedAt`. `RunningEntry.StartedAt` in `Orchestrator.spawn` already carries the monotonic reading, so the subtraction now uses monotonic-bearing values on both sides and stays invariant under wall-clock NTP steps. Existing tests are semantic and continue to pass. @codex review

PR #278, opened 2026-05-22
fix(tracker/github): implement FetchIssueStatesByIDs + lowercase labels per SPEC §11.1/§11.3 (#211)

### LINKED ISSUE #211: [P2][spec-alignment] §11.1 + §11.3: GitHub tracker missing FetchIssueStatesByIDs and label lowercase
## Summary

The GitHub tracker adapter has two SPEC gaps relative to §11.1 (REQUIRED operations) and §11.3 (Normalization Rules):

1. **Missing `FetchIssueStatesByIDs` (§11.1 REQUIRED operation).** Linear and Gitea both implement it; GitHub does not. Without it, per-tick reconciliation (§8.5 Part B "Tracker state refresh") for GitHub-tracked issues either falls back to a full candidate fetch on every tick or silently degrades.
2. **Labels are not lowercased (§11.3).** Linear's adapter lowercases via `strings.ToLower(label.Name)`; the GitHub adapter only trims whitespace, so an `Important` label on GitHub becomes the literal string `Important`, while the same label on Linear becomes `important`. Workflow templates and dispatch logic that branch on label value silently misbehave when the tracker kind changes.

### §11.1 — missing operation

```bash
$ grep -n FetchIssueStatesByIDs internal/tracker/github.go internal/tracker/linear.go internal/gitea/tracker_client.go
internal/gitea/tracker_client.go:func (c *TrackerClient) FetchIssueStatesByIDs(...) (map[string]string, error)
internal/tracker/linear.go:func (c *LinearClient) FetchIssueStatesByIDs(...) (map[string]string, error)
```

No match in `internal/tracker/github.go`. SPEC §11.1 lists `fetch_issue_states_by_ids(issue_ids)` as a REQUIRED tracker adapter operation.

### §11.3 — labels not lowercased

`internal/tracker/github.go:282-287`:

```go
labels := make([]string, 0, len(issue.Labels))
for _, label := range issue.Labels {
    if name := strings.TrimSpace(label.Name); name != "" {
        labels = append(labels, name)
    }
}
```

Compare `internal/tracker/linear.go` label normalization (calls `strings.ToLower`). SPEC §11.3 normalization rules state:

> labels -> lowercase strings

## Why this matters

1. **Per-tick reconcile gap:** §8.5 Part B says "Fetch current issue states for all running issue IDs." For GitHub, the orchestrator must either skip reconcile or fetch every candidate page — neither matches the SPEC.
2. **Tracker-kind portability:** workflows are expected to be tracker-agnostic. Label-driven prompt logic written against Linear silently breaks the day it routes to a GitHub repo.

## Suggested fix

1. Implement `(*GitHubClient).FetchIssueStatesByIDs(ctx, issueIDs)` returning `map[string]string` of `id -> state`. GitHub's REST API does not have an `[ID!]` bulk endpoint, so iterate `GET /repos/{owner}/{repo}/issues/{number}` per ID (best-effort, bounded concurrency, surface partials). Document the rate-limit posture in `docs/runbooks/github-local-automation.md`.
2. In `mapGitHubIssue` (`internal/tracker/github.go:285`), apply `strings.ToLower(strings.TrimSpace(label.Name))`. Update `github_test.go` fixtures so the test asserts the lowercase output.
3. Verify there is a single GitHub tracker adapter consumer path that uses the returned states, so the reconcile flow exercises the new method.

## Acceptance criteria

- [ ] `(*GitHubClient).FetchIssueStatesByIDs` exists, returns `map[string]string`, and handles per-ID `404` (issue deleted) and `410` (gone) without aborting the whole call.
- [ ] Labels in `mapGitHubIssue` are lowercased before append.
- [ ] `internal/tracker/github_test.go` covers (a) `FetchIssueStatesByIDs` happy path + partial failure; (b) label lowercase.
- [ ] Orchestrator reconcile (`internal/orchestrator/poller.go`) routes through `FetchIssueStatesByIDs` for GitHub the same way it does for Linear/Gitea.
- [ ] Runbook updated for the new GitHub call pattern.

## Refs

- SPEC §11.1 (REQUIRED operations), §11.3 (normalization rules), §8.5 Part B
- Code: `internal/tracker/github.go:282-287, 270-306`, `internal/tracker/linear.go` (reference impl)
- Related: gitea adapter has its own normalization gap (separate issue)

hey-yvan:
Closes #211.

## Summary

GitHub adapter had two SPEC gaps relative to Linear/Gitea:

1. **SPEC §11.1** lists `fetch_issue_states_by_ids` as REQUIRED. Without it, the orchestrator's per-tick reconcile (§8.5 Part B) silently skipped state refresh for GitHub-tracked running issues — the `IssueStateRefresher` type assertion in `poller.refreshRunningIssueStates` returned `ok=false`.
2. **SPEC §11.3** says labels must normalize to lowercase. Linear lowercased; GitHub did not, so a workflow branching on label value silently misbehaved when the tracker kind switched.

## Implementation

- Add `issueNumbers sync.Map` to `GitHubClient` and populate from each issue in `listIssuesForState`. GitHub REST has no `/issues/{global_id}` endpoint; we must dispatch by repo issue number. The cache maps global issue ID → repo issue number.
- Add `FetchIssueStatesByIDs`: per-ID `GET /repos/{owner}/{repo}/issues/{number}` using the cached number. IDs without a cached number are silently skipped (consistent with Gitea — they'll be cached on the next list cycle).
- Per-ID `404` / `410` are treated as "issue removed" and skipped; other HTTP errors abort the call (matching gitea fail-closed posture).
- State derivation: first label matching `active_states` / `terminal_states` / `inactive_states` configured on `TrackerConfig` wins (case-insensitive, returned lowercased); otherwise fall back to GitHub's `open`/`closed` state. This matches GitHub's convention of encoding workflow position as labels.
- `mapGitHubIssue` now applies `strings.ToLower(strings.TrimSpace(label.Name))` per §11.3.
- Runbook documents the per-tick refresh + rate-limit posture (sequential per-ID at the 5,000/hour PAT REST limit).

## Test plan

- [x] `TestMapGitHubIssueLowercasesLabels` — mixed-case input → lowercase output, empty filtered
- [x] `TestGitHubClientFetchIssueStatesByIDsUsesCachedIssueNumbers` — happy path + 404 + 410 + uncached-ID skipped
- [x] `TestGitHubClientFetchIssueStatesByIDsSurfacesNon404Errors` — HTTP 500 aborts the call
- [x] `TestGitHubClientFetchIssueStatesByIDsRequiresToken` / `EmptyInputReturnsEmptyMap`
- [x] `TestGitHubClientSatisfiesTrackerInterfaces` extended with `IssueStateRefresher`
- [x] `go test ./internal/tracker/ ./internal/orchestrator/ ./internal/gitea/` all pass
- [x] `gofmt -l` clean
- [x] `go vet ./...` clean

@codex review

### DIFF
--- a/docs/runbooks/github-local-automation.md
+++ b/docs/runbooks/github-local-automation.md
@@ -177,6 +177,18 @@ GitHub tracker pagination is fail-closed: if issue or open-PR pagination
 exceeds the configured scan cap, the poll returns an error instead of acting on
 a truncated issue set.
 
+Per-tick state refresh (SPEC §8.5 Part B) for already-running GitHub issues
+issues one `GET /repos/{owner}/{repo}/issues/{number}` per running issue,
+sequentially, in poll order. The repo issue number is taken from the cache
+populated by the active-issue list. Per-ID `404`/`410` responses are treated as
+"issue removed" and silently skipped so a single deleted issue does not abort
+reconciliation for the rest of the running set; other HTTP errors abort the
+refresh so a transient outage cannot silently degrade reconciliation. At the
+default GitHub primary REST rate limit (5,000 req/hour for a personal access
+token) this is comfortably within budget for tens of concurrent running issues
+on the default poll cadence; cut the worker count or extend the poll interval
+if the workload approaches that ceiling.
+
 ## Install unattended macOS LaunchAgents
 
 ```bash
--- a/internal/tracker/github.go
+++ b/internal/tracker/github.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -33,6 +34,7 @@ type GitHubClient struct {
 	Logf    func(format string, args ...any)
 
 	paginationCapHits atomic.Int64
+	issueNumbers      sync.Map // map[string]int — global issue ID → repo issue number, populated by listing
 }
 
 type githubIssue struct {
@@ -147,6 +149,7 @@ func (c *GitHubClient) listIssuesForState(ctx context.Context, issueState, label
 			if mapped.ID == "" {
 				continue
 			}
+			c.cacheIssueNumber(mapped.ID, issue.Number)
 			if _, ok := seen[mapped.ID]; ok {
 				continue
 			}
@@ -283,7 +286,7 @@ func mapGitHubIssue(issue githubIssue, mappedState string) (Issue, error) {
 	}
 	labels := make([]string, 0, len(issue.Labels))
 	for _, label := range issue.Labels {
-		if name := strings.TrimSpace(label.Name); name != "" {
+		if name := strings.ToLower(strings.TrimSpace(label.Name)); name != "" {
 			labels = append(labels, name)
 		}
 	}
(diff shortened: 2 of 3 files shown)


hey-yvan:
@codex review

PR #280, opened 2026-05-22
fix(runner): emit startup_failed / unsupported_tool_call / turn_input_required per SPEC §10.4 (#214)

### LINKED ISSUE #214: [P2][spec-alignment] §10.4: runner never emits startup_failed / unsupported_tool_call / turn_input_required events (follow-up to #208)
## Summary

SPEC §10.4 lists twelve runtime events the Codex app-server client SHOULD emit. PR #208 closed three of the previously-missing ones (`other_message`, `malformed`, `approval_auto_approved`). Three more in the SPEC list are declared as Go constants in `internal/task/task.go` and exposed by `task.RuntimeEvents()`, but the runner never actually emits them:

- `startup_failed` — should fire on `initialize` / `thread/start` / `turn/start` failures during session startup.
- `unsupported_tool_call` — should fire when the agent invokes a dynamic tool the runtime has not advertised.
- `turn_input_required` — should fire when the agent asks for operator input, MCP elicitation, or approval that is not auto-approved.

The grep is explicit:

```bash
$ grep -n "EventStartupFailed\|EventUnsupportedToolCall\|EventTurnInputRequired" internal/runner/*.go
(no output)
```

These are observability holes, not protocol holes — the wire response is already correct in each path. The orchestrator's snapshot/state surface simply has no way to distinguish "agent failed to start" from "agent ran a turn and failed", or "agent asked for input that nobody gave" from "agent crashed".

### `startup_failed` — `internal/runner/codex_app_server.go:196-235`

Every error return at lines 197, 201, 230, 234 wraps the failure as a Go error and bubbles up to the worker without emitting an event. The orchestrator sees the run end with `result.Err != nil` and falls through the failure path with no event class.

### `unsupported_tool_call` — `internal/runner/codex_app_server.go:781-816`

`handleDynamicToolCall` returns a `dynamicToolResult(false, "unsupported dynamic tool: …")` to the wire when `c.tools.Lookup(name)` misses, but does not call `c.recordRuntimeEvent(task.EventUnsupportedToolCall, …)`. The agent is told the tool failed; the orchestrator/state surface learns nothing.

### `turn_input_required` — `internal/runner/codex_app_server.go:596-655`

`replyServerRequest` auto-replies `decline` / empty permissions / `deny` for every approval/input/elicitation request that is not auto-approved, but only the auto-approved branch records an event (`recordApprovalAutoApproved`, line 616). The decline branches do not emit `turn_input_required`. The MCP elicitation case (line 652) is the same — silent decline.

This is a precondition for #183 (Surface input-blocked Symphony sessions) — without the event, the orchestrator cannot mark the issue blocked. Filing the runner-side fix separately so it can land independently of the orchestrator state change.

## Suggested fix

1. **`startup_failed`** at the four startup failure sites in `codex_app_server.go:197, 201, 230, 234`. Emit before wrapping the error:
   ```go
   c.recordRuntimeEvent(task.EventStartupFailed, c.withRuntimeContext(map[string]any{
       "phase":  "thread/start", // or initialize / turn/start
       "error":  errSafeMessage(err),
   }))
   return fmt.Errorf("codex app-server thread/start: %w", err)
   ```
   `errSafeMessage` is the redaction helper that should land in #193 (do not echo raw `msg["params"]`).
2. **`unsupported_tool_call`** in `handleDynamicToolCall` at `codex_app_server.go:790-792`. Add the emission once the unsupported branch is taken (i.e. before returning the synthesized failure result):
   ```go
   c.recordRuntimeEvent(task.EventUnsupportedToolCall, c.withRuntimeContext(map[string]any{
       "tool":      name,
       "arguments": redactArgs(arguments),
   }))
   ```
3. **`turn_input_required`** in `replyServerRequest` for every decline branch in `protocolServerRequestResult` (lines 631-654):
   - `item/commandExecution/requestApproval` decline
   - `item/fileChange/requestApproval` decline
   - `item/permissions/requestApproval` (empty permissions)
   - `execCommandApproval` / `applyPatchApproval` deny
   - `item/tool/requestUserInput` (regardless of policy)
   - `mcpServer/elicitation/request` decline
   Use a single helper `recordInputRequired(method, params)` so the call site is one line per case.

## Acceptance criteria

- [ ] Three new event emission paths landed; corresponding `recordRuntimeEvent` calls exist in `codex_app_server.go`.
- [ ] Tests in `internal/runner/codex_app_server_test.go` assert each event is emitted in a failure scenario (mock `initialize` failure, mock dynamic tool the server is not advertising, mock elicitation request).
- [ ] No PII / token bleed: the emitted payload uses the same redaction helper that should land via #193.
- [ ] No behavior change in the wire response — the runner continues to reply `decline` / `deny` / `unsupported` as before.
- [ ] `docs/runbooks/task-api.md` event list reflects the now-complete §10.4 vocabulary.

## Refs

- SPEC §10.4 (runtime event vocabulary)
- Code: `internal/runner/codex_app_server.go:196-235, 596-655, 781-816`
- Constants: `internal/task/task.go:60, 65, 67`
- Related: #208 (predecessor — emitted other_message/malformed/approval_auto_approved), #183 (input-blocked sessions, depends on `turn_input_required` emission), #193 (redaction helper to share)

hey-yvan:
Closes #214.

## Summary

SPEC §10.4 lists twelve runtime events the Codex app-server client SHOULD emit. Three were declared as `task.Event*` constants but the runner never actually fired them. Without them the orchestrator's snapshot / `/api/v1/state` cannot distinguish startup failures, unsupported tool calls, or operator-required approvals from generic turn failures.

## Implementation

- **`startup_failed`**: emit at the four startup failure sites in `run()` — `initialize`, `initialized`, `thread/start` (including `extractString` failure), and `turn/start` on the first turn only (later turns are operational, not startup). Payload carries `phase` + `error`. Errors come from JSON-RPC framing or `extractString`, not user-controlled params, so safe to surface verbatim without `safeTurnReason` redaction.
- **`unsupported_tool_call`**: emit in `handleDynamicToolCall` when `c.tools.Lookup` misses, alongside the existing structured failure on the wire. Payload carries tool name + parsed arguments.
- **`turn_input_required` for declined approvals**: an approval method that is NOT auto-approved gets a decline/deny/empty-permissions wire response — which is itself a signal that the operator must act. New `approvalDeclinedServerRequest()` guards a new branch in the run loop that emits the event and bails with `InputRequiredError`, matching the existing `inputRequiredServerRequest` path.

### Behavior change

Previously, declined approvals were silently sent and codex kept going. Now the issue blocks until the operator clears it. Set `codex.approval_policy` to a permissive rule set to keep auto-continue. Deleted `TestCodexAppServerRunnerDeclinesApprovalsWhenPolicyRequiresReview` — it pinned the SPEC-non-compliant silent-decline behavior; replaced by `TestCodexAppServerRunnerEmitsTurnInputRequiredOnDeclinedApproval`.

## Test plan

- [x] `TestCodexAppServerRunnerEmitsStartupFailedOnInitializeError`
- [x] `TestCodexAppServerRunnerEmitsStartupFailedOnThreadStartError`
- [x] `TestCodexAppServerRunnerEmitsUnsupportedToolCallForUnknownTool`
- [x] `TestCodexAppServerRunnerEmitsTurnInputRequiredOnDeclinedApproval`
- [x] All existing runner / orchestrator / worker / cmd tests pass
- [x] `gofmt -l` clean

@codex review

### DIFF
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -240,9 +240,11 @@ func (c *appServerClient) run(ctx context.Context, in RunInput, prompt string) e
 			"version": "0.1.0",
 		},
 	}); err != nil {
+		c.recordStartupFailed("initialize", err)
 		return fmt.Errorf("codex app-server initialize: %w", err)
 	}
 	if err := c.notify("initialized", map[string]any{}); err != nil {
+		c.recordStartupFailed("initialized", err)
 		return err
 	}
 
@@ -253,10 +255,12 @@ func (c *appServerClient) run(ctx context.Context, in RunInput, prompt string) e
 		"dynamicTools":   appServerDynamicToolSpecs(in.Workflow.Config),
 	})
 	if err != nil {
+		c.recordStartupFailed("thread/start", err)
 		return fmt.Errorf("codex app-server thread/start: %w", err)
 	}
 	threadID, err := extractString(threadResult, "thread", "id")
 	if err != nil {
+		c.recordStartupFailed("thread/start", err)
 		return fmt.Errorf("codex app-server thread/start: %w", err)
 	}
 	c.threadID = threadID
@@ -286,10 +290,16 @@ func (c *appServerClient) run(ctx context.Context, in RunInput, prompt string) e
 			"sandboxPolicy":  in.Workflow.Config.Codex.TurnSandboxPolicy,
 		})
 		if err != nil {
+			if turn == 1 {
+				c.recordStartupFailed("turn/start", err)
+			}
 			return fmt.Errorf("codex app-server turn/start: %w", err)
 		}
 		turnID, err := extractString(turnResult, "turn", "id")
 		if err != nil {
+			if turn == 1 {
+				c.recordStartupFailed("turn/start", err)
+			}
 			return fmt.Errorf("codex app-server turn/start: %w", err)
 		}
 		c.turnID = turnID
@@ -557,6 +567,17 @@ func (c *appServerClient) awaitTurnCompletion(ctx context.Context) error {
 						c.recordInputRequiredMessage(method, msg)
 						return &InputRequiredError{Method: method}
 					}
+					// SPEC §10.4 turn_input_required also fires for an
+					// approval request that is not auto-approved — the
+					// runner's wire response is a decline / empty
+					// permissions / deny per protocolServerRequestResult,
+					// which is itself a signal that the operator must
+					// act. Without the event the orchestrator cannot
+					// distinguish this from a transient turn failure.
+					if approvalDeclinedServerRequest(method, c.approvalPolicy) {
+						c.recordInputRequiredMessage(method, msg)
+						return &InputRequiredError{Method: method}
+					}
 					c.lastTerminal = time.Now()
 					continue
 				}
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

PR #283, opened 2026-05-22
fix(workspace): cross-process flock on mirror operations (#228)

### LINKED ISSUE #228: [P2][bug] Workspace mirror cache uses process-local sync.Mutex; two concurrent `cmd/worker` processes sharing AIOPS_MIRROR_ROOT race on git fetch/worktree-add
## Summary

`internal/workspace/mirror.go:13-29` protects bare-mirror operations with a `sync.Mutex` map keyed by mirror path. This is **process-local** — the `mirrorLocksMu` variable lives in one process's memory. Two independent `go run ./cmd/worker` processes sharing the same `AIOPS_MIRROR_ROOT` (e.g., one local dev worker + a CI worker on the same host, or two `tracker.kind` flavors deployed side-by-side on one machine) have completely separate lock maps.

That means: while Worker A is in the middle of `git worktree add <new>` against `~/.aiops/mirrors/github.com/owner/repo.git`, Worker B can simultaneously run `git fetch --prune` on the same mirror. Git's own locking (`<mirror>/index.lock`, `<mirror>/packed-refs.lock`) catches **most** concurrent operations but does not protect against:

- A worktree register operation seeing intermediate `packed-refs` state from a concurrent `--prune`.
- A `git worktree list --porcelain` racing against a fresh `git worktree add` (the new worktree may not yet be in the list).
- Concurrent `git gc --auto` runs triggered by independent workers' fetches.

## Evidence

`internal/workspace/mirror.go:13-29`:

```go
var (
    mirrorLocksMu sync.Mutex
    mirrorLocks   = map[string]*sync.Mutex{}
)

func acquireMirrorLock(mirror string) func() {
    mirrorLocksMu.Lock()
    lock := mirrorLocks[mirror]
    if lock == nil {
        lock = &sync.Mutex{}
        mirrorLocks[mirror] = lock
    }
    mirrorLocksMu.Unlock()

    lock.Lock()
    return lock.Unlock
}
```

`MirrorRoot` (line 41-49) defaults to `<user-cache-dir>/aiops-platform/mirrors` — `~/.cache/aiops-platform/mirrors` on Linux, `~/Library/Caches/aiops-platform/mirrors` on macOS. Both are user-scoped, so any two workers run by the same user collide.

The scripts/local-github-worker.sh launchd setup explicitly supports running concurrent workers per service.

## Why this matters

1. **Symptom space.** Operators see sporadic `fatal: <branch> already exists` or `fatal: branch '<x>' already used by worktree at <y>` errors from `git worktree add`. They re-run, it "just works". The race is hard to repro and silently corrupts retry telemetry.
2. **Mirror corruption potential.** A `git fetch --prune` running mid-add can delete a ref that `git worktree add` is reading. Git's own locks help, but the operations are not atomic together.
3. **No SPEC requirement to share mirrors across processes**, but the project's design encourages it (workspace cache runbook recommends a single mirror root for cache reuse).

## Suggested fix

Add a cross-process file lock around mirror operations. Two approaches:

### A. flock-style file lock (simple)

```go
import "github.com/gofrs/flock"

func acquireMirrorLock(mirror string) (func(), error) {
    flockPath := mirror + ".lock"
    if err := os.MkdirAll(filepath.Dir(flockPath), 0o755); err != nil {
        return nil, err
    }
    fl := flock.New(flockPath)
    if err := fl.Lock(); err != nil {
        return nil, err
    }
    return func() { _ = fl.Unlock() }, nil
}
```

Combine with the existing process-local mutex (acquire process-local first to avoid contending on the OS lock; release in reverse order).

### B. Trust git's own locks + retry on `fatal: ... already locked`

Cheaper to ship but fragile. Will surface races as intermittent failures that look like flaky CI.

Recommendation: Option A. `golang.org/x/sys/unix.Flock` or `gofrs/flock` are both small dependencies; the latter handles Windows too.

Also: document in `docs/runbooks/workspace-cache.md` that two workers sharing a mirror root must be on the same host (file locks are per-host) and that NFS-mounted mirror roots are unsupported.

## Acceptance criteria

- [ ] Mirror operations are serialized across processes on the same host.
- [ ] Process-local Mutex retained as an inner lock to keep single-process contention cheap.
- [ ] Test in `internal/workspace/mirror_test.go` spawns two goroutines simulating concurrent processes (use a subprocess fixture); asserts no corruption.
- [ ] Runbook documents host-scope and unsupported NFS mirrors.

## Refs

- Code: `internal/workspace/mirror.go:13-29`, used by `internal/workspace/manager.go:PrepareGitWorkspace`
- SPEC §9.5 (workspace safety invariants — does not address inter-process)
- Related: `scripts/local-github-worker.sh` (launchd setup that produces concurrent workers)

hey-yvan:
Closes #228.

## Summary

`internal/workspace/mirror.go`'s `acquireMirrorLock` used a process-local `sync.Mutex` map. Two independent `cmd/worker` processes sharing `AIOPS_MIRROR_ROOT` had completely separate lock maps, so Worker A's `git worktree add` could race Worker B's `git fetch --prune` against the same bare repo. Git's own per-ref locks catch most of this but not:

- A `worktree list --porcelain` racing a fresh `worktree add` (new worktree not yet visible).
- `packed-refs` intermediate state from a concurrent `--prune`.
- Concurrent `git gc --auto` triggered by independent fetches.

Symptom: sporadic `fatal: <branch> already exists` / `fatal: branch '<x>' already used by worktree at <y>` that "just worked" on retry.

## Implementation

- New `acquireMirrorFileLock(mirror)` in `mirror_lock_unix.go` takes an exclusive `flock(2)` on a `<mirror>.lock` sidecar file. The Windows build target (`mirror_lock_windows.go`) is a documented no-op; only the in-process mutex serializes on Windows.
- `acquireMirrorLock` now takes the process-local mutex first (cheap, no syscall for single-process callers), then the file lock; releases in reverse order. If the flock syscall fails (NFS without lockd, fd exhaustion, etc.) the function falls back to in-process-only mutex so single-process deployments keep working.
- The lock sidecar lives next to the mirror so `lsof <mirror>.lock` attributes a stuck mirror to a specific worker pid.

## Documentation

- `docs/runbooks/workspace-cache.md` gains a "Cross-process mirror locking" section documenting host-scope, NFS unsupported, Windows-no-op, and the sidecar file location.

## Test plan

- [x] `TestMirrorFileLockSerializesAcrossProcesses` — re-execs the test binary as a child that holds the flock for 750ms; the parent's own acquire must block ≥100ms (asserting the file lock is actually serializing across pids, not just within the goroutine).
- [x] `TestAcquireMirrorFileLockReleaseAllowsReacquire` — pins release-then-reacquire so the poll-loop retry path does not deadlock.
- [x] Existing `TestEnsureMirror_ConcurrentFirstUseSerializesClone` still passes — process-local mutex remains the inner lock.
- [x] `go test ./internal/workspace/` green
- [x] `gofmt -l` clean; `go vet` clean

@codex review

### DIFF
--- a/docs/runbooks/workspace-cache.md
+++ b/docs/runbooks/workspace-cache.md
@@ -41,6 +41,34 @@ On Linux containers `os.UserCacheDir()` resolves to `$XDG_CACHE_HOME` or
 when you want the cache on a dedicated volume — for example the
 `workspace-cache` named volume in `deploy/docker-compose.yml`.
 
+## Cross-process mirror locking
+
+Two workers sharing the same `AIOPS_MIRROR_ROOT` on the same host serialize
+their `git fetch` / `git clone` / `git worktree add` operations through an
+advisory `flock(2)` on a `<mirror>.lock` sidecar file (#228). This closes
+the gap where the per-process `sync.Mutex` alone could not stop one
+worker's `git fetch --prune` from racing another worker's
+`git worktree add` against the same bare repo, which historically
+surfaced as sporadic `fatal: <branch> already exists` errors that "just
+worked" on retry.
+
+Operational notes:
+
+- The lock is **host-scoped**. Two workers must run on the same OS
+  instance for the lock to mediate them; cross-host concurrency is not
+  supported, because `flock(2)` is a per-kernel primitive.
+- **NFS mirror roots are unsupported.** `flock(2)` on NFS is silently
+  best-effort on Linux (depends on `lockd` being healthy) and not
+  honoured on macOS. Use a local filesystem.
+- The lock sidecar file lives next to the mirror as `<mirror>.lock`.
+  Operators removing a mirror by hand should also remove the sidecar,
+  but must not delete it while a worker holds the lock — `lsof
+  <mirror>.lock` attributes the holder to a specific pid.
+- Windows hosts have no `flock(2)`; the cross-process layer is a no-op
+  there and only the in-process mutex serializes. The supported
+  deployment platforms (Linux containers, macOS LaunchAgents) both use
+  the full file-lock path.
+
 ## Cleanup policy
 
 The worker does **not** automatically delete old worktrees. The
--- a/internal/workspace/manager.go
+++ b/internal/workspace/manager.go
@@ -198,10 +198,13 @@ func (m *Manager) PrepareGitWorkspace(ctx context.Context, t task.Task) (string,
 		return "", false, statErr
 	}
 	mirror := mirrorPathFor(MirrorRoot(m.MirrorRoot), t.CloneURL)
-	unlock := acquireMirrorLock(mirror)
+	unlock, err := acquireMirrorLock(mirror)
+	if err != nil {
+		return "", false, err
+	}
 	defer unlock()
 
-	mirror, err := m.ensureMirrorLocked(ctx, t.CloneURL, mirror)
+	mirror, err = m.ensureMirrorLocked(ctx, t.CloneURL, mirror)
 	if err != nil {
 		return "", false, err
 	}
--- a/internal/workspace/mirror.go
+++ b/internal/workspace/mirror.go
@@ -27,14 +27,14 @@ var (
 // Releases in reverse order: file lock first (so the next process can
 // proceed), then process-local mutex.
 //
-// If the file-lock acquisition fails (kernel rejected the syscall, NFS-
-// without-lockd, etc.) the function falls back to in-process-only locking.
-// That preserves correctness for single-process deployments — the most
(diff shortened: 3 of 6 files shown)


hey-yvan:
@codex review

hey-yvan:
Addressed in 66a7ad0: acquireMirrorLock now fails closed — propagates the flock error and releases the process-local mutex; both EnsureMirror and PrepareGitWorkspace surface it. TestAcquireMirrorLockFailsClosedOnFlockError pins the new behavior. @codex review

PR #284, opened 2026-05-22
feat(orchestrator): SPEC §8.5 Part A orchestrator-level stall reconciliation (#246)

### LINKED ISSUE #246: [P3][spec-alignment] §8.5 Part A + §16.3: orchestrator-level stall reconciliation is missing (only the runner self-detects stalls)
## Summary

SPEC §8.5 Part A puts stall detection at the **orchestrator** layer, not (only) at the runner:

> Part A: Stall detection
>
> - For each running issue, compute `elapsed_ms` since:
>   - `last_codex_timestamp` if any event has been seen, else
>   - `started_at`
> - If `elapsed_ms > codex.stall_timeout_ms`, terminate the worker and queue a retry.
> - If `stall_timeout_ms <= 0`, skip stall detection entirely.

SPEC §16.3 algorithm:

```
function reconcile_running_issues(state):
  state = reconcile_stalled_runs(state)
  ...
```

The Go implementation:

- ✓ **Runner-side stall detection** exists at `internal/runner/codex_app_server.go:407-413, 477-481`. It computes `time.Since(c.lastTerminal)` against `c.stallTimeoutMs` and returns `StallError`.
- ✗ **Orchestrator-side stall reconciliation** does not exist. `grep -rn "reconcile_stalled\|StallTimeoutMs" internal/orchestrator/` returns nothing.

D14 (#88, closed by [chore: refresh DEVIATIONS](https://github.com/xrf9268-hue/aiops-platform/issues/147)) covered the runner path. The orchestrator path was not addressed.

## Why this matters

1. **Failure scope.** Runner-self-detection only catches stalls that happen *while the runner goroutine is running*. If the runner goroutine itself wedges (deadlock on a Go channel, blocked stdin write that doesn't honor its context), the subprocess could still be alive but no `StallError` is ever produced. The orchestrator's `LastCodexAt` field already exists (`internal/orchestrator/state.go`) but no code compares it against `stall_timeout_ms`.
2. **Cross-runner consistency.** The `mock` and `codex` (exec) runners don't have stall detection. The orchestrator-level check would apply uniformly across all runner kinds.
3. **SPEC text is normative.** §16.3 calls out `reconcile_stalled_runs` as the first step of `reconcile_running_issues`. The current poller (`internal/orchestrator/poller.go`) goes straight to Part B (tracker state refresh) and skips Part A entirely.

## Evidence

```bash
$ grep -rn "stall\|Stall" internal/orchestrator/
# (no output)

$ grep -rn "StallTimeoutMs" internal/
internal/runner/codex_app_server.go:161:        stallTimeoutMs      int
internal/runner/codex_app_server.go:174:        c.stallTimeoutMs = in.Workflow.Config.Codex.StallTimeoutMs
internal/runner/codex_app_server.go:408-413:    [runner self-stall code]
internal/runner/codex_app_server.go:477-481:    [runner self-stall code]
internal/workflow/loader.go:284:        if cfg.Codex.StallTimeoutMs < 0 {
internal/workflow/config.go:263:        StallTimeoutMs    int   `yaml:"stall_timeout_ms,omitempty"`
internal/workflow/config.go:409:                StallTimeoutMs: 300000,
```

The orchestrator's `RunningEntry` already has `LastCodexAt` (set during `dispatchCodexUpdate`). The data plumbing is in place; just the reconcile step is missing.

## Suggested fix

Add a `reconcileStalledRuns` step to the poller's reconcile loop. Sketch:

```go
// internal/orchestrator/poller.go (inside reconcileTick before tracker refresh)
func reconcileStalledRuns(state *OrchestratorState, stallTimeoutMs int, now time.Time) []IssueID {
    if stallTimeoutMs <= 0 {
        return nil
    }
    var stalled []IssueID
    threshold := time.Duration(stallTimeoutMs) * time.Millisecond
    for id, entry := range state.Running {
        ref := entry.LastCodexAt
        if ref.IsZero() {
            ref = entry.StartedAt
        }
        if now.Sub(ref) > threshold {
            stalled = append(stalled, id)
        }
    }
    return stalled
}
```

Hand the list to the orchestrator's terminate path so the worker is killed and a retry is scheduled (failure retry, exponential backoff per §8.4).

The runner-side stall detection can stay — it shortens the catch latency from "one poll interval" to "one read timeout". The orchestrator check is the safety net for the case where the runner itself is wedged.

## Acceptance criteria

- [ ] `reconcileStalledRuns` (or equivalent) runs at the start of every reconcile tick.
- [ ] Stall detection uses `LastCodexAt` if non-zero, else `StartedAt`.
- [ ] `codex.stall_timeout_ms <= 0` disables detection (per §8.5).
- [ ] Stalled issues are terminated and retried via the existing failure-retry path (exponential backoff).
- [ ] Test in `internal/orchestrator/poller_test.go` covers the happy path and disabled path.
- [ ] `DEVIATIONS.md` D14 row is updated (or a new row added) noting both runner and orchestrator detection now present.

## Refs

- SPEC §8.5 Part A, §16.3
- Code: `internal/orchestrator/poller.go`, `internal/orchestrator/state.go` (`RunningEntry.LastCodexAt`), `internal/runner/codex_app_server.go:407-481`
- Related closures: #88 (runner-side, closed), #147 (D14 status refresh)

hey-yvan:
Closes #246.

## Summary

SPEC §16.3 `reconcile_running_issues` lists Part A (`reconcile_stalled_runs`) as the FIRST step of each poll tick — before Part B's tracker-state refresh. The Go implementation had Part B but no Part A: the runner's self-stall detection (`codex_app_server.go`) only fires while the runner goroutine is alive, so a wedged runner with a live subprocess would keep the issue claimed forever. The mock and `codex exec` runners have no stall detection at all.

`LastCodexAt` was already on `RunningEntry` (D14); only the reconcile step was missing.

## Implementation

- `Orchestrator.ReconcileStalledRuns(ctx, stallTimeoutMs, wait)`: scans `st.Running` for entries whose `LastCodexAt` (falling back to `StartedAt` before any runtime event is observed) is older than `stallTimeoutMs`, cancels the worker context. Zero/negative `stallTimeoutMs` disables detection (SPEC §6.4 default).
- The cancelled entry stays Claimed and does **not** set `ReconcileCancel`: `finalizeRunOp` observes the `context.Canceled` as a normal worker failure and schedules a retry. Operator-initiated cancellation and stall-driven cancellation have different semantics; reusing the failure-retry path keeps backoff/budget accounting aligned.
- Wired into `Poller.reconcileTick` as the first step (per §16.3 order) via a new `StallTimeoutMs` field on `ReconciliationConfig`. The field is populated from `workflow.Config.Codex.StallTimeoutMs` by both the runtime workflow wiring (`workflow_runtime.go`) and the worker entry (`cmd/worker/main.go`).
- `export_test.go` adds `WithStateForTest` so tests can backdate `LastCodexAt` without sleeping for the full stall budget.

## Test plan

- [x] `TestReconcileStalledRunsCancelsWorkerPastTimeout` — backdated `LastCodexAt` + 100ms budget → worker context cancelled
- [x] `TestReconcileStalledRunsLeavesActiveRunsAlone` — recent `LastCodexAt` + 5s budget → no false positive
- [x] `TestReconcileStalledRunsSkipsWhenTimeoutDisabled` — `stall_timeout_ms=0` is a no-op
- [x] `go test ./internal/orchestrator/ ./cmd/worker/ ./internal/worker/` green
- [x] `gofmt -l` clean; `go vet ./...` clean

@codex review

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -1113,6 +1113,7 @@ func reconciliationConfigForWorkflow(cfg workflow.Config) orchestrator.Reconcili
 		TerminalStates:    cfg.Tracker.TerminalStates,
 		InactiveStates:    inferredInactiveStates(cfg.Tracker),
 		WorkerExitTimeout: 30 * time.Second,
+		StallTimeoutMs:    cfg.Codex.StallTimeoutMs,
 	}
 }
 
--- a/internal/orchestrator/actor.go
+++ b/internal/orchestrator/actor.go
@@ -504,6 +504,44 @@ func (o *Orchestrator) ReconcileTrackerIssues(ctx context.Context, issuesByID ma
 	return o.ReconcileTrackerIssuesAndWait(ctx, issuesByID, activeStates, 0)
 }
 
+// ReconcileStalledRuns implements SPEC §8.5 Part A / §16.3
+// reconcile_stalled_runs: for each running issue compute elapsed time
+// since the last observed runtime event (RunningEntry.LastCodexAt,
+// falling back to StartedAt before any event has been seen) and, if it
+// exceeds stallTimeoutMs, cancel the worker so the finalize path
+// schedules a retry. stallTimeoutMs <= 0 skips detection entirely (SPEC
+// §6.4 default).
+//
+// The Codex app-server runner has its own self-stall detection; this
+// orchestrator-side path closes the gap when the runner goroutine itself
+// wedges or when a non-Codex runner (mock, codex exec) produces no
+// StallError. Without this an issue with `LastCodexAt` long in the past
+// would stay claimed forever.
+//
+// wait is the per-tick budget for waiting on cancelled worker
+// goroutines to exit (mirrors ReconcileTrackerIssuesAndWait). Use 0 for
+// fire-and-forget cancel.
+func (o *Orchestrator) ReconcileStalledRuns(ctx context.Context, stallTimeoutMs int, wait time.Duration) error {
+	if stallTimeoutMs <= 0 {
+		return nil
+	}
+	reply := make(chan []*RunningEntry, 1)
+	if err := o.submit(ctx, &reconcileStalledRunsOp{
+		timeout: time.Duration(stallTimeoutMs) * time.Millisecond,
+		now:     time.Now(),
+		result:  reply,
+	}); err != nil {
+		return err
+	}
+	var canceled []*RunningEntry
+	select {
+	case canceled = <-reply:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+	return waitForReconciledWorkers(ctx, canceled, wait)
+}
+
 // RefreshActiveTrackerIssues updates stored issue metadata for in-process runs
 // and queued retries whose issues are still observed in the active set, without
 // canceling any work. Use this when the active listing may be partial (so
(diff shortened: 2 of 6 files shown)


hey-yvan:
@codex review

PR #286, opened 2026-05-22
fix(orchestrator): don't let a wedged worker block global reconciliation (#285)

### LINKED ISSUE #285: Follow up unresolved review thread from PR #284
A merged PR still had an unresolved, non-outdated review discussion. This issue was filed automatically so the feedback is not lost after merge.

Merged PR: https://github.com/xrf9268-hue/aiops-platform/pull/284
Discussion: https://github.com/xrf9268-hue/aiops-platform/pull/284#discussion_r3289537010
Location: `internal/orchestrator/poller.go:180`
Author: @chatgpt-codex-connector

Original review body:

```
Continue tracker reconciliation when stall wait times out
In internal/orchestrator/poller.go's reconcileTick path, any error from ReconcileStalledRuns causes an immediate return before Part B executes. If a truly wedged worker ignores cancellation, ReconcileStalledRuns(..., WorkerExitTimeout) can return context deadline exceeded, and then every tick skips tracker-state refresh/reconciliation for all runs, leaving unrelated inactive/terminal issues unreconciled and claimed until that one stuck run exits. This new early return turns one stalled run into a global reconciliation blocker; treat Part A timeout as a non-fatal poll error and continue with Part B.
Useful? React with 👍 / 👎.
```

hey-yvan:
## Summary

Resolves #285 (follow-up review thread from merged PR #284).

In `reconcileTick`, any error from `ReconcileStalledRuns` (SPEC §8.5 Part A) caused an immediate `return nil, err` before Part B (tracker-state refresh + inactive reconciliation) executed. When a worker ignored its cancellation, the `WorkerExitTimeout` wait returned `context.DeadlineExceeded` and **every subsequent poll tick** skipped Part B for **all** runs — one stuck worker froze global reconciliation until it eventually exited, leaving unrelated inactive/terminal issues unreconciled and claimed.

## Fix

Treat Part A errors as a **non-fatal poll error** when the poll context is still alive: join into the returned error and continue to Part B. A canceled poll context still propagates (we bail).

```go
if err := p.orchestrator.ReconcileStalledRuns(ctx, ...); err != nil {
    if ctx.Err() != nil {
        return nil, err
    }
    fetchErr = errors.Join(fetchErr, err)
}
```

## Test plan

- [x] New regression test `TestPollOnceContinuesReconciliationWhenStalledRunCleanupTimesOut`:
  - Two running issues: `wedged` (stuck dispatcher that ignores cancellation, `LastCodexAt` backdated so Part A targets it) and `movable` (tracker moves to `Cancelled` so Part B should cancel it).
  - With the bug: `movable`'s context is never cancelled (Part B skipped). Verified the test fails on the unpatched code.
  - With the fix: both contexts get cancelled in the same poll tick. ✅
- [x] `go test ./internal/orchestrator/ -count=1` — all pass.
- [x] `go vet ./internal/orchestrator/...` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YK9vBRoJpVrJ3uKdVqG9HA)_

### DIFF
--- a/internal/orchestrator/poller.go
+++ b/internal/orchestrator/poller.go
@@ -173,15 +173,21 @@ func (p *Poller) reconcileTick(ctx context.Context, activeIssues []tracker.Issue
 	if p.stateTracker == nil {
 		return nil, errors.New("orchestrator poller reconciliation requires state tracker")
 	}
+	var fetchErr error
 	// SPEC §16.3 reconcile_running_issues order: Part A (stall reconciliation)
 	// first so any wedged worker is cancelled before Part B's tracker-state
-	// refresh would otherwise leave it claimed indefinitely.
+	// refresh would otherwise leave it claimed indefinitely. A WorkerExitTimeout
+	// on a worker that ignores cancellation surfaces as context.DeadlineExceeded;
+	// surface that as a non-fatal poll error so one stuck run cannot block Part B
+	// from reconciling unrelated inactive/terminal issues in the same tick.
 	if err := p.orchestrator.ReconcileStalledRuns(ctx, p.reconcile.StallTimeoutMs, p.reconcile.WorkerExitTimeout); err != nil {
-		return nil, err
+		if ctx.Err() != nil {
+			return nil, err
+		}
+		fetchErr = errors.Join(fetchErr, err)
 	}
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

---

PR #287, opened 2026-05-22
fix(orchestrator): SPEC §16.6 retry-fire candidate fetch (#191)

### LINKED ISSUE #191: [P3][spec-alignment] §8.4 step 3: retry handler should fetch candidates and confirm absence before releasing the claim
## Summary

SPEC §8.4 / §16.6 (`on_retry_timer`) defines a 5-step contract for a fired retry timer:

1. Fetch active candidate issues.
2. Find the specific issue by `issue_id`.
3. **If not found, release claim.**
4. If found and slots available, dispatch.
5. If found but no longer active, release claim.

`internal/orchestrator/actor.go:836-847` implements step 1, 2, 4, 5 but step 3's "if not found, release claim" branch is reached only via the `RetryAttempts[r.id]` map lookup — it conflates "stale retry entry consumed by a parallel op" with "issue truly no longer in candidates." The SPEC explicitly distinguishes the two: step 3 only releases the claim after confirming the issue is absent from the just-fetched candidate list.

## Evidence

- Code: `internal/orchestrator/actor.go:836-847` — entry-existence check drops the retry but the candidate-fetch-then-find dance to confirm step 3 absence is implicit (the find loop at line ~852 onward handles found/not-found uniformly).
- SPEC §16.6 lines 1894-1904 spell out the candidate fetch step:
  ```
  candidates = tracker.fetch_candidate_issues()
  if fetch failed: schedule_retry(... "retry poll failed")
  issue = find_by_id(candidates, issue_id)
  if issue is null:
    state.claimed.remove(issue_id)
    return state
  ```

## Impact

In the current shape, a transient candidate-fetch failure during the retry tick gets treated implicitly by the dispatch retry path; SPEC wants an explicit re-schedule with the typed error `"retry poll failed"` so observability surfaces show it. The `step 3` release path is narrower than the current implementation's path, which is generally permissive.

## Suggested fix

Refactor `retryFireOp.apply` to follow the §16.6 sequence explicitly:

1. Pop the retry entry (already done).
2. Call `tracker.fetch_candidate_issues()`.
3. On fetch failure → `ScheduleRetry(... attempt+1, error: "retry poll failed")` and return.
4. `find_by_id(candidates, issue_id)`. If absent → `state.Claimed.Remove(id)` and return.
5. Else proceed to capacity check + dispatch (existing logic).

This makes the path SPEC-traceable line-by-line and surfaces `"retry poll failed"` as a typed retry reason.

## Acceptance criteria

- [ ] `retryFireOp.apply` follows the §16.6 5-step order explicitly.
- [ ] A transient `fetch_candidate_issues` failure re-schedules the retry with the `"retry poll failed"` error string surfaced.
- [ ] Test in `internal/orchestrator/actor_test.go` covers (a) issue genuinely absent → claim released; (b) fetch fails → retry re-scheduled with the error string.

## Refs

- SPEC §8.4, §16.6
- Code: `internal/orchestrator/actor.go:836-910`

hey-yvan:
## Summary

Closes #191. The fired failure-retry timer now follows SPEC §16.6's 5-step contract explicitly:

1. Fetch active candidate issues.
2. Find the specific issue by `issue_id`.
3. If not found → **release the claim**.
4. If found and slots available → dispatch from the refreshed tracker state.
5. If found but no longer active → release the claim (folded into step 3, since `ListActiveIssues` returns only active issues).

Before this change, `retryFireOp.apply` dispatched directly from the cached `entry.Issue` snapshot. A transient candidate-fetch failure was silently absorbed by the implicit "next reconcile will catch it" path, and the "issue genuinely absent" branch was conflated with the stale-entry guard. Both gaps are now closed:

- **Fetch error path** submits a new `retryPollFailedOp` that reschedules with a typed `"retry poll failed: <err>"` reason and `attempt+1` so the next backoff window can try again.
- **Issue absent path** submits a new `retryFireAfterFetchOp` that calls `state.ReleaseClaim(id)` and records a `RuntimeEventFailed("retry released: issue absent from active candidates")` for observability.
- **Dispatch path** refreshes `entry.Issue` from the freshly-fetched tracker state before running the existing capacity gates (factored into `retryFireDispatchTail` so both the new and legacy entry points share it).

A new optional `CandidateLister ActiveIssueLister` on `orchestrator.Deps` wires the same multi-tracker lister the poll loop uses; `RuntimePoller.pollerForSnapshot` installs it on every workflow snapshot reload so the active-state vocabulary stays in sync. When the lister is `nil` (most unit tests), retry fires fall back to the legacy direct-dispatch path so existing behavior is preserved verbatim.

## Test plan

Added three SPEC §16.6 conformance tests in `internal/orchestrator/actor_test.go`:

- [x] `TestRetryFire_ReleasesClaimWhenIssueAbsentFromCandidates` — successful fetch returns no issues → claim released, no dispatch.
- [x] `TestRetryFire_ReschedulesWithRetryPollFailedOnFetchError` — fetch fails → retry rescheduled with `attempt+1` and `Error` contains `"retry poll failed"` plus the wrapped underlying message.
- [x] `TestRetryFire_DispatchesWhenCandidateFetchReturnsIssue` — fetch returns the issue with a refreshed state → dispatched worker carries the fresh state, not the dispatch-time snapshot.
- [x] Full `go test ./internal/orchestrator/...` (+ `-race`) passes; existing `TestRetryFireUsesRefreshedIssueWhenReconciledMidQueue` / `TestRetryFire_RespectsPerStateCapacityBeforeSpawning` / `TestTrackerRecheckedDispatchDoesNotConsumeFailureRetry` continue to pass via the legacy fallback.

@codex review

https://claude.ai/code/session_019Ro8rz8GEmW6m4QQPHeaXM

---
_Generated by [Claude Code](https://claude.ai/code/session_019Ro8rz8GEmW6m4QQPHeaXM)_

### DIFF
--- a/internal/orchestrator/actor.go
+++ b/internal/orchestrator/actor.go
@@ -32,6 +32,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/xrf9268-hue/aiops-platform/internal/runner"
@@ -123,6 +124,15 @@ type Deps struct {
 	Scheduler         Scheduler
 	MaxFailureRetries *int
 	MaxTurns          *int
+	// CandidateLister, when set, enables the SPEC §16.6 retry-fire
+	// candidate-fetch step. A fired failure-retry timer fetches the
+	// active candidate list, confirms the issue is still present,
+	// releases the claim when it is not, and surfaces a typed
+	// "retry poll failed" reschedule when the fetch itself fails.
+	// Production wires this from the same multi-tracker lister the
+	// poll loop uses; when nil, retry fires dispatch directly from
+	// the cached entry.Issue (legacy behavior, kept for unit tests).
+	CandidateLister ActiveIssueLister
 }
 
 // Orchestrator is the SPEC §3.1 / §7.4 "single authority" that owns
@@ -143,6 +153,13 @@ type Orchestrator struct {
 	// the same setting before they return.
 	maxTurns  int
 	retryWake chan struct{}
+
+	// candidateLister supplies the SPEC §16.6 candidate fetch that a
+	// fired failure-retry timer consults before re-dispatching. The
+	// field is read on every retry fire and written when the runtime
+	// poller rebuilds its tracker set, so a mutex guards the swap.
+	candidateListerMu sync.Mutex
+	candidateLister   ActiveIssueLister
 
 	// runCtx is captured by Run so followup goroutines can cancel
 	// their work when the actor stops. Set once at the top of Run
@@ -178,6 +195,7 @@ func New(state *OrchestratorState, deps Deps) *Orchestrator {
 		maxFailureRetries: maxFailureRetries,
 		maxTurns:          maxTurns,
 		retryWake:         make(chan struct{}, 1),
+		candidateLister:   deps.CandidateLister,
 		started:           make(chan struct{}),
 	}
 	if aware, ok := deps.Dispatcher.(interface{ AttachOrchestrator(*Orchestrator) }); ok {
@@ -250,6 +268,26 @@ func (o *Orchestrator) retryWakeCh() <-chan struct{} {
 
 func (o *Orchestrator) wakeRetryPollLoop() {
 	_ = o.queuePollWake()
+}
+
+// SetCandidateLister installs (or swaps) the lister the actor consults on a
+// fired failure-retry timer. Safe to call before or after Run.
+func (o *Orchestrator) SetCandidateLister(l ActiveIssueLister) {
+	if o == nil {
+		return
+	}
+	o.candidateListerMu.Lock()
+	o.candidateLister = l
+	o.candidateListerMu.Unlock()
+}
+
+func (o *Orchestrator) currentCandidateLister() ActiveIssueLister {
+	if o == nil {
+		return nil
+	}
+	o.candidateListerMu.Lock()
+	defer o.candidateListerMu.Unlock()
+	return o.candidateLister
 }
 
 func (o *Orchestrator) queuePollWake() bool {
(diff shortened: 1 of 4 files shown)


hey-yvan:
@codex review

---

hey-yvan:
@codex review

---

hey-yvan:
@codex review

---

hey-yvan:
@codex review

---

hey-yvan:
@codex review

hey-yvan:
@codex review

hey-yvan:

Codex adversarial-review (gpt-5.4-mini, effort=high) against `origin/main` flagged two blocking findings.

**[high] Retry-fire candidate fetch bypasses service routing** — `internal/orchestrator/runtime_poller.go:109-117`
The new retry-fire lister installed here is only `activeIssueListerFromStates{tracker: multiLister, states: ...}`, while the poll loop applies `selectRoutedCandidates` before dispatching. For service-routed workflows, an issue that stays active but no longer matches this workflow's service route can still be re-dispatched from a failure retry, so the orchestrator can run an agent against work owned by another service/project using this runtime's workflow config until the next reconciliation catches up.
*Recommendation:* Make the retry-fire `CandidateLister` route-aware whenever `poller.routing` is set, and add a test where a queued retry's issue changes to a non-matching service while still active.

**[medium] Stuck candidate fetch can leave a retry claimed forever** — `internal/orchestrator/actor.go:1181-1201`
When a failure retry fires with a `CandidateLister`, the actor clears `entry.Timer` and then calls `ListActiveIssues` on `o.runCtx`. If the tracker call hangs or the client ignores cancellation until process shutdown, no replacement timer is armed and no actor op is submitted, leaving the issue in `RetryAttempts`/`Claimed` with no observable retry progress.
*Recommendation:* Wrap `ListActiveIssues` with a per-call retry-fetch timeout and ensure timeout submits `retryPollFailedOp`; add a test lister that blocks past the timeout and verifies the retry is rescheduled instead of remaining claimed.

hey-yvan:
## Follow-up: SPEC §16.6 hardening (commit `ea7ef73`)

Force-pushed on top of a rebase onto current `main` (`5b0bdfa`). Addresses the two findings from the adversarial review comment above, plus one hygiene cleanup. Cross-checked against upstream Symphony Elixir reference at `elixir/lib/symphony_elixir/orchestrator.ex` lines 1062–1104 (`handle_retry_issue` / `handle_retry_issue_lookup` / `release_issue_claim`).

### Fixes

**1. [HIGH] Retry-fire candidate fetch bypassed service routing** — `runtime_poller.go`

The poll loop applies `selectRoutedCandidates(issues, *p.routing)` after the active fetch (`poller.go:136-143`), but the retry-fire lister installed on the orchestrator was just `activeIssueListerFromStates{...}`. For service-routed Linear workflows that meant a queued failure retry whose issue had since routed to another service could still be dispatched against this workflow.

New `routedActiveIssueLister` wraps the active-state lister and runs the same `selectRoutedCandidates` filter when `poller.routing != nil`. Routing errors propagate so an ambiguous route is treated as a fetch failure (reschedule via "retry poll failed"), not a silent absence (release claim).

Upstream Symphony has no service-routing concept (single-project assumption) — this wrap is the aiops-platform-specific layer required to keep retry-fire internally consistent with the poll loop.

**2. [MEDIUM] Stuck candidate fetch could orphan a retry forever** — `actor.go`

In the lister branch the followup ran `lister.ListActiveIssues(o.runCtx)`. `o.runCtx` has no per-call deadline, and a tracker client that ignores ctx cancellation would pin the goroutine indefinitely — `entry.Timer` was already cleared and no `retryFireOp` would be resubmitted, so the issue stayed in `Claimed`/`RetryAttempts` forever. Per-request HTTP timeouts on the Linear/Gitea/GitHub adapters (#303) cover today's clients, but the orchestrator must not depend on every future adapter to do so.

New `retryFetchTimeout` (45s) wraps the fetch in `context.WithTimeout`; on timeout the followup falls through the existing `retryPollFailedOp` path so SPEC §16.6's "retry poll failed" reschedule remains the only forward-progress path.

**3. Thread `r.kind` into `retryFireAfterFetchOp`** instead of hardcoding `RetryKindFailure`. The continuation early-return above means the value is the same today, but writing the literal in the lister branch bakes the invariant in via the wrong layer; if a future `RetryKind` is added or the short-circuit reordered the hardcode would silently dispatch the wrong kind.

### Tests

Three new SPEC §16.6 conformance tests:

- `TestRetryFire_FetchTimeoutReschedulesAsRetryPollFailed` — shrinks `retryFetchTimeout` to 50ms, lister only returns on ctx cancellation; asserts `attempt+1` reschedule with `"retry poll failed"` + `context deadline exceeded` wrapped in, no dispatch, claim not orphaned.
- `TestRetryFire_RoutedAwayIssueReleasesClaim` — routing-aware lister returns empty; asserts the absent-issue contract (no dispatch, retry + claim released).
- `TestRuntimePollerInstallsRoutedCandidateListerForServiceRouting` — end-to-end check that `pollerForSnapshot` installs `routedActiveIssueLister` on the orchestrator when `Services` are configured; in-route issue passes with `ServiceName` stamped, off-route issue is filtered out.

All existing `TestRetryFire_*` still pass under `-race` (`ok 3.466s`).

### Rebase notes

Cherry-picked the original `a4a5d81` onto current `main` (`5b0bdfa`). Conflicts vs. #304 (recover guards) and #301 (dispatch preflight) resolved by keeping both sides — the PR's refactored `rearmRetryCapacityRecheck` now also accepts a `site` parameter so the recover guards retain their granular `orchestrator.retry_capacity_timer` / `orchestrator.retry_state_capacity_timer` site names. The new fetch followup is automatically panic-protected by the existing `safeGo("orchestrator.followup", …)` wrapper from #304.

---

hey-yvan:
## Follow-up #2: post-merge audit fixes (commit `a1bdc44`)

After force-pushing `ea7ef73`, ran an **independent adversarial audit** (separate subagent, blind brief, no access to my conclusions). It found one HIGH and three MEDIUM/LOW gaps. Three are fixed in `a1bdc44` on top of `ea7ef73`; the fourth is deferred to #306 with the rationale below.

### Fixed in `a1bdc44`

**[HIGH] Retry-fire lister mirrored `selectRoutedCandidates` but not `filterEligibleCandidates`** — `runtime_poller.go`. The poll loop applies both between `ListActiveIssues` and dispatch (`poller.go:152`); the retry-fire path was missing the second one. A queued failure-retry whose Todo issue gained a non-terminal blocker between schedule and fire would still dispatch, while the poll loop would refuse the same issue on the next tick. Exactly the cross-cutting class the new `AGENTS.md` "Audit adjacent paths" checklist (merged in #305) calls out — `ea7ef73` stopped one filter short. New `eligibleActiveIssueLister` extends the chain to `active → routed → eligible`, matching `poller.go` exactly. Pinned by `TestRuntimePollerRetryListerMirrorsPollLoopFilters` (integration) + `TestRetryFire_TodoIssueBlockedByOpenDependencyReleasesClaim` (actor-level).

**[MEDIUM] Misnamed/redundant test** — `TestRetryFire_RoutedAwayIssueReleasesClaim` (added in `ea7ef73`) installed a stub lister that just returned an empty slice, so `selectRoutedCandidates` was never exercised. Functionally identical to `TestRetryFire_ReleasesClaimWhenIssueAbsentFromCandidates`. Rewritten to install the real `routedActiveIssueLister{inner: stub, cfg: workflowCfg}`, flip the stub's `ProjectSlug` between schedule and fire, and assert the routing layer drops it.

**[LOW] `retryFireAfterFetchOp.apply` did not refresh `entry.Identifier`** — `actor.go`. Upstream `handle_active_retry` uses the refreshed `issue.identifier` on every subsequent reschedule (`orchestrator.ex:1142-1161`), so a Linear identifier rename between schedule and fire was leaving runtime events stamped with the stale value. Now mirrors upstream. Pinned by `TestRetryFireAfterFetch_RefreshesEntryIdentifier`.

**[LOW] `retryPollFailedOp.apply` shutdown race** — `actor.go`. If `o.runCtx` was cancelled between fetch and apply, the followup's `ScheduleRetry` would fail but the `RuntimeEventFailed` had already recorded, leaving `entry.Timer == nil` with no replacement. Short-circuit when `o.runCtx.Err() != nil` so shutdown reclaims state cleanly.

### Deferred to #306

**[MEDIUM] `rearmRetryCapacityRecheck` diverges from upstream `handle_active_retry`** — capacity-full re-arms a 100 ms recheck timer instead of `ScheduleRetry(attempt+1, "no available orchestrator slots")`. Pre-existing behavior that PR #287 only refactored (extracted into a helper); fixing it is outside this PR's "candidate fetch" scope. Tracked separately per AGENTS.md rule 2.

### Dimensions the audit checked and found clean

- Followup goroutine panic safety (via `safeGo` at `actor.go:207`)
- `context.WithTimeout` derivation under already-cancelled `o.runCtx`
- `entry.Timer` actor-goroutine-only access discipline
- Ambiguous-route semantics (routing error → reschedule, not silent release)
- Backoff progression in `retryPollFailedOp` (scheduler honors `nextAttempt`)
- Mergeable against current `main`

### History on this PR

| commit | scope |
|---|---|
| `5791bcd` | original SPEC §16.6 candidate fetch (rebased onto current main) |
| `ea7ef73` | service routing + fetch timeout + thread `r.kind` (adversarial review round 1) |
| `a1bdc44` | eligibility filter + identifier refresh + shutdown race (adversarial review round 2 — this comment) |

Companion: #305 (the AGENTS.md checklist), #306 (the deferred capacity-recheck divergence).

@codex review

---

hey-yvan:
## Follow-up #3: round-3 audit fixes (commit `d755630`)

Independent round-3 audit on `a1bdc44` returned `MERGE-READY` with five LOW findings. Three are addressed in `d755630`; two are deferred with the rationale below.

### Fixed in `d755630`

**[LOW, but real — placebo test] `TestRetryFireAfterFetch_RefreshesEntryIdentifier` did not pin the new code.** The round-3 audit did a mutation test: deleting `entry.Identifier = id` from `retryFireAfterFetchOp.apply` left the test passing. Root cause — the assertion read `disp.issueAt(0).Identifier`, which is fed by the pre-existing `entry.Issue = *r.found` line, not by the new mutation. Same class of failure as the misnamed test caught in round 2 (`TestRetryFire_RoutedAwayIssueReleasesClaim`).

Rewritten as `TestRetryFireAfterFetch_RefreshedIdentifierSurfacesInSnapshot`: cap `MaxConcurrentAgents=1`, fill the slot with a blocker, fire the retry. The fetch refreshes the entry (`Issue` + `Identifier`), `retryFireDispatchTail` trips the capacity gate and re-arms via `rearmRetryCapacityRecheck` — leaving the entry in `RetryAttempts`. `RetryView.Identifier` reads `RetryEntry.Identifier` directly (`state.go:818`), so asserting `v.Retrying[0].Identifier == "NEW-1"` does pin the mutation. **Mutation-verified**: deleting the line now produces `condition not met within 2s`.

**[LOW] No-routing branch of retry-fire wrap had no coverage.** `TestRuntimePollerRetryListerMirrorsPollLoopFilters` only exercised the Linear+Services branch. A future refactor that dropped the `eligibleActiveIssueLister` wrap from the Gitea/service-less branch would not break any test. New `TestRuntimePollerRetryListerWrapsEligibilityFilterWithoutServiceRouting` feeds a Todo+non-terminal-blocker through the installed lister, asserts it's dropped; positive control with a clean issue confirms the wrap isn't over-aggressive.

**[LOW] Dead `o.runCtx != nil` guard in `retryPollFailedOp.apply`.** The audit traced submit/Run ordering: `o.runCtx` is set before `close(o.started)`, and `submit` blocks on the unbuffered ops channel until the actor loop iterates — by which time `runCtx` is non-nil. The `!= nil` half of the conjunction was dead defensive code. Dropped. Comment also corrected — state isn't "reclaimed" via mutation, it's dropped when the process exits.

### Deferred (not in this PR's scope)

**[LOW cosmetic]** Two-layer type assertion in `TestRuntimePollerRetryListerMirrorsPollLoopFilters` is brittle under benign tracing-wrap insertion. The behavioral matrix below the type assertion already covers the wire contract; the type cast adds documentation value about the intended shape. Trade-off call; leaving as-is.

**[observability]** Upstream `handle_retry_issue_lookup` logs three distinct lines for terminal / non-retry-candidate / absent branches (`orchestrator.ex:1086-1098`); the Go port collapses all three into a single `"retry released: issue absent from active candidates"` event. **Pre-existing pattern, not introduced by this PR series.** Worth tracking separately under `area:spec-alignment` if the operator UX gap matters; out of scope here.

### History on this PR

| commit | scope |
|---|---|
| `5791bcd` | original SPEC §16.6 candidate fetch (rebased) |
| `ea7ef73` | round-1 audit fixes: service routing + fetch timeout + thread `r.kind` |
| `a1bdc44` | round-2 audit fixes: eligibility filter + identifier refresh + shutdown race |
| `d755630` | round-3 audit fixes: strengthened identifier test + no-routing coverage + drop dead guard |

### Process note

Round 3 found a real flaw — the placebo test — that round 2 missed despite both rounds being adversarial audits with no access to my conclusions. Pattern: each round catches one weak test from the previous round's fix commit. The new `AGENTS.md` checklist item 3 ("Run an adversarial pass on your own diff") is doing what it's supposed to; the lesson is that **mutation testing** (delete the new line and re-run) is the cheap mechanical check that distinguishes a real test from a placebo. Worth adding as a sub-bullet under item 3 the next time `AGENTS.md` rotates.

@codex review

---

Issue #288, opened 2026-05-22
[P3][tests] workspace/worker test fixtures inherit host `commit.gpgsign` and break in signing-required environments


hey-yvan:
## Summary

The git-using tests in `internal/workspace` and `internal/worker` set per-repo `user.email` / `user.name` on the fixture repos they create, but they do **not** explicitly disable `commit.gpgsign`. When the host environment has `commit.gpgsign=true` configured globally (corporate dev boxes, signed-commits-by-default CI images, the Claude Code on the web sandbox, etc.), the fixture `git commit -q -m seed` calls inherit it and try to sign. If the host's signing path is missing or broken, every test that builds a fixture repo fails — even though the test has nothing to do with signing.

This is observable today: a full `go test ./...` in an environment with broken signing produces ~30 failures in `internal/worker` and `internal/workspace`, all with the same shape:

```
[git -C /tmp/.../seed commit -q -m seed]: exit status 128
error: ... signing failed: ...
fatal: failed to write commit object
```

The orchestrator / runner / tracker packages are unaffected because they don't shell out to `git commit` for fixtures.

## Why this matters

1. **Cross-environment portability.** A contributor on a host with `commit.gpgsign=true` global cannot run the test suite locally without re-configuring git globally — a footgun for anyone who has signed commits enforced by their org.
2. **CI-image surprise.** If the project ever switches to a base image that enables signing by default, every test in these two packages would start failing simultaneously, with errors that don't point at the real cause.
3. **Cheap to fix.** Setting `commit.gpgsign=false` per fixture repo at the same place the tests already set `user.email` and `user.name` is two extra lines per setup helper.

## Evidence

`internal/workspace/mirror_test.go:34-35`, `manager_test.go:29-30, 648-649`, `internal/worker/runtask_integration_test.go:40-41, 858-859` all set:

```go
{"git", "-C", work, "config", "user.email", "u@example.com"},
{"git", "-C", work, "config", "user.name", "u"},
```

…but never set `commit.gpgsign=false` (or `tag.gpgsign=false`). The fixture repos inherit whatever the host's `~/.gitconfig` says.

## Suggested fix

Add the disables next to the existing identity config in every fixture setup helper:

```go
{"git", "-C", work, "config", "commit.gpgsign", "false"},
{"git", "-C", work, "config", "tag.gpgsign", "false"},
```

[...]

PR #290, opened 2026-05-22
fix(workspace): SPEC §4.2 workspace key sanitization (#229)

### LINKED ISSUE #229: [P2][spec-alignment] §4.2 + §9.5: workspace key sanitization accepts Unicode letters (unicode.IsLetter) and replaces invalid chars with `-` instead of `_`
## Summary

SPEC §4.2 defines the workspace key normalization rule precisely:

> Derive from `issue.identifier` by replacing any character not in `[A-Za-z0-9._-]` with `_`.

SPEC §9.5 Invariant 3 restates it:

> Only `[A-Za-z0-9._-]` allowed in workspace directory names.
> Replace all other characters with `_`.

The Go implementation diverges from SPEC §4.2 in three places at once (`internal/workspace/manager.go:920-952`):

1. **Lowercases the input** before sanitization (`strings.ToLower(strings.TrimSpace(s))`). SPEC does not lowercase — `ABC-123` and `abc-123` should produce different keys; we conflate them.
2. **Accepts any Unicode letter** via `unicode.IsLetter(r)`. SPEC restricts to `[A-Za-z]` (ASCII only). CJK identifiers, Cyrillic identifiers, etc. survive sanitization unchanged.
3. **Replaces invalid characters with `-`**, not `_`. SPEC explicitly says `_`.

A separate prior issue (#87, closed) reframed the original D13 fix around "task ID → identifier", but did not realign the sanitization rule itself. The current implementation's choices were tested into stability, but they are **not SPEC-conformant**.

## Evidence

`internal/workspace/manager.go:928-952`:

```go
func sanitizeComponent(s string, preserveUnderscore bool) string {
    var b strings.Builder
    b.Grow(len(s))
    inSeparator := false
    for _, r := range strings.ToLower(strings.TrimSpace(s)) { // <-- (1) ToLower
        if unicode.IsLetter(r) || unicode.IsDigit(r) || (preserveUnderscore && r == '_') {
            // <-- (2) unicode.IsLetter accepts CJK / Cyrillic
            b.WriteRune(r)
            inSeparator = false
            continue
        }
        if !inSeparator && b.Len() > 0 {
            b.WriteByte('-')                                 // <-- (3) `-` not `_`
            inSeparator = true
        }
    }
    // ...
}
```

The function is reused for source-type sanitization and (via `SanitizeComponent`) for the workspace identifier key.

### Concrete divergences

| Input identifier | SPEC §4.2 expected | Current output |
|---|---|---|
| `ABC-123` | `ABC-123` (preserved) | `abc-123` |
| `日本語-456` | `__________-456` | `日本語-456` |
| `feat/new-ui` | `feat_new-ui` | `feat-new-ui` |
| `a b c` | `a_b_c` | `a-b-c` |
| `___leading___` | `___leading___` (preserved underscores) | `leading` (collapsed, no underscores) |

The third row is the most operationally visible: a Linear branch slug `feat/new-ui` produces different workspaces on aiops-platform vs. an upstream Symphony Elixir worker.

## Why this matters

1. **Cross-implementation portability.** Switching between aiops-platform and Symphony Elixir for the same project produces different on-disk workspaces. Hooks that reference the workspace by identifier (e.g., `~/.aiops/workspaces/ABC-123`) break on cutover.
2. **Case-conflation collisions.** Linear allows `ABC-123` and `abc-123` as two distinct identifiers (rare but possible across workspaces). Today we collapse them into one workspace, opening agent runs on top of each other's state.
3. **Unicode acceptance is silent data corruption.** A CJK / Cyrillic identifier survives sanitization unchanged, then crosses git's own restrictions for branch names later, producing confusing downstream errors.

## Suggested fix

Replace the body of `sanitizeComponent` with the SPEC §4.2 rule directly:

```go
func sanitizeComponent(s string, preserveUnderscore bool) string {
    s = strings.TrimSpace(s)
    var b strings.Builder
    b.Grow(len(s))
    for i := 0; i < len(s); i++ {
        c := s[i]
        switch {
        case c >= 'A' && c <= 'Z',
             c >= 'a' && c <= 'z',
             c >= '0' && c <= '9',
             c == '.', c == '-', c == '_':
            b.WriteByte(c)
        default:
            b.WriteByte('_') // SPEC §4.2: replace with `_`
        }
    }
    out := b.String()
    if out == "" { return "unknown" }
    if utf8.RuneCountInString(out) > maxSanitizedLength {
        out = string([]rune(out)[:maxSanitizedLength])
    }
    return out
}
```

`preserveUnderscore` becomes irrelevant (underscores are always kept now). `SanitizeSourceType` can collapse into `SanitizeComponent`.

### Migration

Existing on-disk workspaces use the old sanitization. A pure cutover would orphan them. Two options:

- **Hard cutover.** Run a one-shot migration on worker startup that renames `<old-sanitized>` → `<new-sanitized>` for issues still active. Add a `docs/runbooks/workspace-cache.md` migration section.
- **Soft cutover.** Honor the old sanitization for `MIGRATION_DEADLINE = 2 weeks`; warn loudly; switch over after.

This is the right time to do it — pre-release per `DEVIATIONS.md`. Pick hard cutover.

## Acceptance criteria

- [ ] `sanitizeComponent` replaces non-`[A-Za-z0-9._-]` characters with `_`.
- [ ] No lowercasing; case is preserved.
- [ ] ASCII letters only (Unicode letters are not in the allowed set).
- [ ] Existing tests updated; new tests cover the SPEC examples (Unicode, mixed case, underscores).
- [ ] `internal/workspace/manager.go:SanitizeSourceType` reviewed for consistency.
- [ ] `docs/runbooks/workspace-cache.md` documents the migration story.
- [ ] DEVIATIONS.md row for D13 / D29 updated to record the realignment.

## Refs

- SPEC §4.2, §9.5 Invariant 3
- Code: `internal/workspace/manager.go:914-953`
- Related: #87 (D13 closure — moved key from task ID to identifier but did not realign the sanitization rule itself)

hey-yvan:
Closes #229.

## Summary

`workspace.SanitizeComponent` now follows the SPEC §4.2 rule exactly:

> Derive from `issue.identifier` by replacing any character not in
> `[A-Za-z0-9._-]` with `_`.

The previous sanitizer lowercased the input, accepted any
`unicode.IsLetter` rune (so CJK identifiers passed through unchanged),
and substituted `-` for invalid characters with adjacent-separator
collapse and edge trim. All three divergences called out in #229 are
gone:

| Input identifier | Before #229 | After (SPEC §4.2) |
|---|---|---|
| `ABC-123` | `abc-123` | `ABC-123` |
| `日本語-456` | `日本語-456` | `___-456` |
| `feat/new-ui` | `feat-new-ui` | `feat_new-ui` |
| `a b c` | `a-b-c` | `a_b_c` |
| `___leading___` | `leading` | `___leading___` |

Two harness-only guards stay on top of the SPEC rule, both documented
in the function comment and in `docs/runbooks/workspace-cache.md`:

- **120-rune length cap** so the result fits common filesystem name
  limits.
- **Single-component path-traversal block**: the *exact* strings `.`
  and `..` (and the empty string) map to `"unknown"` so `PathFor`
  never emits a `filepath.Join` traversal segment. A `..` *substring*
  inside a longer component (e.g. `a..b`) is left intact — `filepath.Join`
  treats it as a name.

`preserveUnderscore` is gone (underscores are always kept now), and
`SanitizeSourceType` is collapsed into `SanitizeComponent`. The one
external caller in `internal/worker/policy_feedback.go` now uses
`SanitizeComponent` directly.

## Migration

The cutover is hard, per `AGENTS.md` §SPEC-alignment-is-a-hard-requirement
+ `DEVIATIONS.md` pre-release posture. The runbook
(`docs/runbooks/workspace-cache.md`) gains a `SPEC §4.2 path-component
sanitization` section that explains the rule and gives operators an
audit-then-`rm -rf` recipe to clear orphaned dirs created under the
old sanitizer before rolling the worker forward.

Active **Rework** workspaces survive the cutover automatically:
`reworkWorkspaceKeyPrefixes` (`internal/worker/reconcile.go`) now
emits both the canonical `_rework_` prefix (current sanitizer) and the
legacy `-rework-` prefix (pre-SPEC layout), so an in-flight Rework
workspace from before #229 still matches its active issue on the
first reconcile after deploy. Plain (non-rework) per-issue dirs
created under the old sanitizer are not back-compat-matched and will
be reconciled away with `reason:unknown`.

## DEVIATIONS

D13 row now links to #229 as the realignment that closed the
"partial closure" gap noted in the issue (the row had been marked
`Closed` for the task-ID → identifier move alone, without realigning
the sanitization rule itself).

## Test plan

- [x] `go test ./internal/workspace/... ./internal/worker/... ./cmd/worker/...`
  passes for every SPEC-§4.2 / `PathFor` / rework-prefix /
  startup-reconcile test on this branch.
- [x] New `TestSanitizeComponentFollowsSpec` table covers every SPEC
  example in #229 plus the harness guards (rune cap, `.`/`..`/empty
  fallback, interior `..` left untouched).
- [x] New `TestSanitizeComponentCasesDoNotCollide` guards the
  case-conflation collision the old `ToLower` caused for distinct
  Linear identifiers (`ABC-123` vs `abc-123`).
- [x] Existing `TestPathForUsesStableSanitizedIssueIdentifier` updated
  to the new layout, plus an explicit `HasPrefix("/workspaces/")`
  assertion on the adversarial `../Linear Issue//Needs_Safety` source
  type so we keep the workspace-root-containment guarantee that the
  pre-#229 edge-trim used to provide implicitly.
- [x] `TestReworkWorkspaceKeyPrefixesMatchCanonicalAndLegacyOffsetSuffixes`
  now asserts both prefix forms are returned.
- [x] `TestStartupReconcileKeepsServiceRoutedReworkWorkspaceAfterUpdatedAtChanges`,
  `TestStartupReconcileConfigPreservesServiceRoutedActiveWorkspaceKey`,
  `TestReconcileStartupMatchesCurrentSanitizedWorkspaceLayout`,
  `TestReconcileStartupKeepsWorkspaceWhenActiveAndTerminalKeysConflict`
  all updated to the new layout (case preserved, `_` separators).
- [x] Full `go test ./...` was run against `main` and against this
  branch; the only failures in either run are the pre-existing
  host-environment failures (gpgsign-required global config without a
  working signer, separately tracked in #288 / PR #289). My branch
  introduces zero new failures relative to `main`.

## Out of scope

- No changes to `sanitizeLegacyWorkspaceKey`. After this PR it is
  effectively equivalent to `SanitizeComponent` for most inputs, but
  it differs on `TrimSpace` and edge-trim semantics, so the dual
  lookup in `workspaceKeysForRawIssueKeys` is kept verbatim to
  preserve any cross-vintage workspace-discovery match.

@codex review

---
_Generated by [Claude Code](https://claude.ai/code/session_01B5fqH1s7fwuuaeYcr1s24M)_


hey-yvan:

Codex adversarial-review (gpt-5.4-mini, effort=high) against `origin/main` flagged one blocking finding.

**[high] Legacy rework matching misses old lowercased workspace keys** — `internal/worker/reconcile.go:316-326`
The compatibility logic only emits the current case-preserved key with `_rework_` and the same key with `-rework-`. But the pre-SPEC sanitizer (the same one being replaced in this diff) lowercased the entire key and timestamp, so an active legacy workspace such as `lin-123-rework-2026-05-16t10-00-00z` will not match an active issue whose identifier is `LIN-123`. In that case `activeReworkIssueForWorkspace` fails to classify the directory as active, and once terminal issues are fetched `ReconcileStartup` can remove it as `unknown`, losing an in-flight agent workspace on the first boot after upgrade.

This contradicts the PR description's migration promise:
> Active **Rework** workspaces survive the cutover automatically… so an in-flight Rework workspace from before #229 still matches its active issue on the first reconcile after deploy.

*Recommendation:* Add a pre-SPEC sanitizer for reconciliation compatibility (lowercasing, dash separator collapse, lowercased timestamp handling), generate prefixes from that legacy base, and add a regression test with an uppercase identifier and an old `-rework-…t…z` directory.

PR #291, opened 2026-05-22
fix(workspace+worker): SPEC §9.1 workspace reuse + §9.4 after_create gating (#245)

### LINKED ISSUE #245: [P2][spec-alignment] §9.1 + §9.4 + §17.2: workspaces are nuked on every run, and `after_create` fires on every run (not "only on new creation")
## Summary

Two related SPEC violations in the worker's per-task workspace flow, both stemming from the same root cause: `PrepareGitWorkspace`'s `createdNow` signal is discarded by the worker, and the function itself unconditionally tears down the existing worktree.

### SPEC §9.1 — workspaces reused across runs

> Workspace persistence:
> - Workspaces are reused across runs for the same issue.
> - Successful runs do not auto-delete workspaces.

`internal/workspace/manager.go:212-215`:

```go
_ = runQuiet(ctx, mirror, "git", "worktree", "remove", "--force", workdir)
_ = os.RemoveAll(workdir)
if err := runQuiet(ctx, mirror, "git", "worktree", "prune"); err != nil {
    return "", false, fmt.Errorf("worktree prune: %w", err)
}
```

Every call to `PrepareGitWorkspace` removes the existing worktree before recreating it with `git worktree add ... -B`. The on-disk per-issue workspace is nuked on every run; only the **bare mirror** is preserved across runs. SPEC's "reused across runs" invariant is broken.

### SPEC §9.4 — `after_create` runs only on creation

> - `hooks.after_create`
>   - Runs only when a workspace directory is newly created.

`internal/worker/runtask.go:201`:

```go
workdir, _, err := mgr.PrepareGitWorkspace(ctx, t)
```

The worker discards the `createdNow` boolean. The next call to the hook (`runtask.go:212`):

```go
if err := runWorkspaceHook(ctx, ev, t.ID, workdir, workspace.HookAfterCreate, hooks.AfterCreate, hooks.TimeoutMs); err != nil {
```

runs `after_create` unconditionally. The hook never gets the "new creation only" semantics SPEC mandates.

SPEC §17.2 Core Conformance:

> `after_create` hook runs only on new workspace creation

This conformance test fails today.

## Why this matters

1. **Reuse semantics are observable.** Long-running iteration on one issue (multiple `Rework` cycles) builds up cached build artifacts, node_modules, downloaded test fixtures, etc. The current implementation discards all of that every run.
2. **`after_create` is no longer the "one-time init" hook it's documented as.** Workflows that put expensive bootstrap commands in `after_create` (e.g. `npm ci`, `pip install -r requirements.txt`) re-run them on every task tick. Operators have no idempotent place to put first-touch setup.
3. **Cross-implementation drift.** Symphony Elixir reuses workspaces by design; an aiops-platform worker watching the same issue burns more disk + network than a SPEC-conforming worker would.
4. **D13 closure assumption.** `internal/orchestrator/state.go:27-31` already calls this out:
   > "this Manager keys them by task ID (deviation D13, tracked under #87)."
   But that comment is stale — D13 was closed by switching the leaf key to the issue identifier (#87) without fixing the reuse semantics.

## Suggested fix

Two-step:

### Step 1 — stop nuking the worktree

Replace the unconditional `worktree remove --force` + `RemoveAll` block with a reuse-or-create branch:

```go
if createdNow {
    // First touch — create the worktree.
    if err := run(ctx, mirror, "git", "worktree", "add", "--no-track", "-B", t.WorkBranch, workdir, startRef); err != nil {
        return "", false, fmt.Errorf("worktree add: %w", err)
    }
} else {
    // Reuse — reset to the new base, but keep cached artifacts.
    if err := run(ctx, workdir, "git", "fetch", "origin", t.BaseBranch); err != nil { ... }
    if err := run(ctx, workdir, "git", "checkout", "-B", t.WorkBranch, startRef); err != nil { ... }
    // Leave .aiops/ etc. intact; the runner resets RUN_SUMMARY.md explicitly.
}
```

### Step 2 — gate `after_create`

```go
workdir, createdNow, err := mgr.PrepareGitWorkspace(ctx, t)
// ...
if createdNow {
    if err := runWorkspaceHook(ctx, ev, t.ID, workdir, workspace.HookAfterCreate, hooks.AfterCreate, hooks.TimeoutMs); err != nil {
        removeWorkdirAfterHookFailure(...)
        return &RunTaskError{Cfg: wcfg, Err: err}
    }
}
```

The `removeWorkdirAfterHookFailure` path is still appropriate when `after_create` fails on a brand-new workspace (SPEC §9.4: "failure aborts workspace creation").

## Migration / tests

- Existing on-disk workspaces are already valid worktrees off the mirror; no migration needed for first deploy after the change — the next run reuses them instead of nuking.
- Add `internal/worker/runtask_test.go` cases:
  - `after_create` fires once and not again on the second tick.
  - A file the runner wrote to the workspace survives the next tick.
  - `before_run` still fires on every tick.
  - A non-fast-forward base branch update is handled by `git checkout -B` not by re-clone.

## Acceptance criteria

- [ ] `after_create` runs only when `createdNow == true`; covered by unit test.
- [ ] Workspace is preserved across runs for the same issue; covered by integration test (file from run N visible at start of run N+1).
- [ ] `before_run` / `after_run` / `before_remove` semantics unchanged.
- [ ] `DEVIATIONS.md` updated: either close D13 cleanly (with this issue link) or open a new row noting the realignment.
- [ ] `docs/runbooks/workspace-cache.md` (or equivalent) documents the new reuse model.

## Refs

- SPEC §9.1 (persistence), §9.4 (hook semantics), §17.2 (conformance test)
- Code: `internal/workspace/manager.go:190-237`, `internal/worker/runtask.go:201-215`
- Stale comment: `internal/orchestrator/state.go:27-31`
- Related: #87 (D13 — partial closure), #229 (D29 — sanitization realignment)

hey-yvan:
Closes #245.

## Summary

`PrepareGitWorkspace` previously nuked the per-issue worktree on every call (`git worktree remove --force` + `RemoveAll`), and the worker ignored the `createdNow` return value, so `after_create` fired on every tick. Both violate SPEC and the §17.2 conformance test:

- §9.1 — *"Workspaces are reused across runs for the same issue."*
- §9.4 — *"`after_create` runs only when a workspace directory is newly created."*

The fix is in two pieces.

### Workspace reuse (`internal/workspace/manager.go`)

After the mirror lock + fetch, probe the existing workdir with `git rev-parse --git-dir`. If it's a valid worktree, take the reuse path:

```sh
git reset --quiet HEAD -- .
git checkout --force --no-track -B <work-branch> origin/<base>
```

The `git reset` is **load-bearing**: the previous run's `EnforcePolicy` calls `git add --intent-to-add --all` (via `Diffstat`) which stages every untracked file pointing at the empty blob. Without the reset, `git checkout --force -B` would treat those staged paths as "in-index-but-not-in-target" and silently delete cached deps, hook artifacts, and `.aiops/POLICY_VIOLATION_FEEDBACK.md` from the working tree. The reset clears the index back to HEAD so only tracked files get updated. `--no-track` matches the create path's `worktree add --no-track` so the work branch never tracks the base branch.

If the workdir exists but `rev-parse --git-dir` fails (a crashed prepare left a partial dir, an operator ran `rm -rf .git`, etc.), fall through to the original nuke-and-recreate path so a corrupted leftover never pins the worker in a bad state — and `createdNow` is reported `true` so `after_create` fires on the recovery run.

### `after_create` gate (`internal/worker/runtask.go`)

Capture `createdNow` from `PrepareGitWorkspace` and only run the hook when it's true. The existing rollback path (`removeWorkdirAfterHookFailure`) is preserved verbatim — it still fires on first-touch hook failures, which is the only case where SPEC requires the workdir teardown.

## Documentation

- `docs/runbooks/workspace-cache.md` gains a *"Workspace reuse across runs (SPEC §9.1 + §9.4)"* section explaining which files survive (untracked) and which get reset (tracked + `RUN_SUMMARY.md` + analysis-only `PLAN.md`), and a *"Recovery from a corrupted workspace"* subsection covering the `rev-parse` fallback.
- `DEVIATIONS.md` D13 row now links to #245 as the realignment that closed the §9.1/§9.4/§17.2 gap left open when D13 was originally closed for the task-ID → identifier move alone.

## Tests

New / replaced:

- `TestPrepareGitWorkspace_RerunReusesWorkspaceAcrossRuns` replaces `_RerunReusesPathIdempotently` (which asserted nuke-on-rerun). Writes both an untracked cache file and a tracked-file modification before the second prepare, then asserts: untracked file survives, tracked modification is discarded, `HEAD == origin/main`, branch == work branch.
- `TestPrepareGitWorkspace_RecreatesCorruptedWorkdir` exercises the `.git` linkage corruption fallback: prepare → `rm -rf .git` → prepare again → assert `createdNow=true` and `.git` is back.
- `TestRunTaskReusesWorkspaceAcrossRunsAndGatesAfterCreate` replaces `TestRunTaskExecutesAfterCreateHookOnRecreatedWorkspace`. Wires `after_create` + `before_run` hooks that append to a shared `hook.log`, runs `RunTask` twice on the same task, asserts the log reads `"after_createbefore_runbefore_run"` (after_create once, before_run twice), no `EventWorkspaceHookStart` for `after_create` in the second run's emitter, and a `stale.txt` untracked artifact survives the reuse.

Unchanged:

- `TestRunTaskExecutesWorkspaceHooksAroundRunner` (single-run hook ordering) — create path is unchanged.
- `TestRunTaskRunsBeforeRemoveHookWhenAfterCreateFails` (after_create failure rollback) — first-touch hook failure path is unchanged.
- `TestPrepareGitWorkspaceDoesNotSetBaseBranchAsWorkBranchUpstream` — the reuse path also uses `--no-track`.

## Test plan

- [x] `go test ./internal/workspace/... ./internal/worker/... ./internal/orchestrator/...` against this branch — pre-existing failures only (`TestRunWorkspaceHookStopsOnNonZeroAndCapturesOutput`, `TestRunVerifyEnvDropsNonAllowlistedSecretsByDefault`, `TestRunTaskPolicyViolationFeedbackWriteErrorIsNonRetryable`), all documented in PR #289 as host-env issues (`sh -lc` sourcing nvm.sh; `chmod 0o555` not honored when running as root). This branch adds zero new failures relative to `main`.
- [x] `go test -race` on the touched packages for the SPEC §9.1 / §9.4 tests above.
- [x] `go vet ./...` clean.

## Out of scope

- Pre-existing host-env test failures (#288 / PR #289).
- Backfilling tracked artifacts (e.g. `RUN_SUMMARY.md`) into a separate "first-run" branch — the existing explicit resets in `RunTask` already handle that for the reuse path.

@codex review

---
_Generated by [Claude Code](https://claude.ai/code/session_01QbBUYFTnRNYMePrCUuh6ve)_


hey-yvan:

Codex adversarial-review (gpt-5.4-mini, effort=high) against `origin/main` flagged one blocking finding.

**[high] Workspace reuse accepts arbitrary git repos at the task path** — `internal/workspace/manager.go:203-262`
The new reuse gate treats `git rev-parse --git-dir` success as sufficient proof that `workdir` is the expected worktree. That only proves the path is some git repository. A previous agent run can delete/recreate its workspace as an independent repo, or replace the workspace path with a symlink to another git repo; on the next tick this code takes the reuse path, runs `git reset`/`git checkout -B` in that repo, skips `after_create`, and then runs hooks/agent work against a tree that may not be the cached mirror for `t.CloneURL`. Impact ranges from starting the task from attacker-controlled stale content to destructive resets of a repo outside the workspace root. This is a regression from the old path, which always removed the existing path before adding a mirror worktree.

*Recommendation:* Use `os.Lstat` to reject symlinks at the workspace path, and before reuse verify the repo is the expected git worktree — e.g. check `git rev-parse --git-common-dir`/`.git` linkage points under the computed mirror and that the worktree is registered for this exact path. If any validation fails, remove only the workspace path safely and recreate via `git worktree add`. Add regression tests for a workspace path replaced by an independent git repo and by a symlink to another git repo.

Issue #306, opened 2026-05-23
orchestrator: capacity-full retry diverges from upstream handle_active_retry (no-slots reschedule)


hey-yvan:
## What

When a fired failure-retry timer finds the issue but capacity is full, the Go orchestrator arms a 100 ms `retryCapacityRecheckDelay` timer instead of rescheduling via the configured backoff. This diverges from upstream `handle_active_retry` (`elixir/lib/symphony_elixir/orchestrator.ex:1142-1161`), which calls `schedule_issue_retry(state, issue.id, attempt + 1, %{identifier: issue.identifier, error: "no available orchestrator slots"})`.

Surfaced by a post-merge independent audit of PR #287 (commits `a1bdc44`, `ea7ef73`). The audit covered the candidate-fetch path that PR #287 reworked; this divergence is in the adjacent capacity-gate path that PR #287 only refactored (extracted `rearmRetryCapacityRecheck`), so it was deferred from #287's scope per `AGENTS.md` rule 2 (every accepted deviation either closes an existing entry or files a new one).

## Where

- `internal/orchestrator/actor.go` — `rearmRetryCapacityRecheck` (post-PR-#287; both call sites are `retryFireDispatchTail` when global or per-state capacity is full).
- `retryCapacityRecheckDelay = 100 * time.Millisecond`.

## Why it matters

1. **Backoff is bypassed under sustained capacity pressure.** A noisy state hitting its `max_concurrent_agents_by_state` cap will produce a 100 ms re-fire loop forever; upstream would grow `attempt+1` through the configured exponential backoff (`min(10s*2^(attempt-1), max_retry_backoff_ms)` per SPEC §8.4) and eventually exhaust `max_failure_retries`, producing a typed terminal failure.

2. **No `"no available orchestrator slots"` runtime event.** Upstream operators see this string surface in the runtime log/event stream as the retry's `error` field. The Go port currently has no observability for the cap-pressure case at all — the runtime event timeline silently goes quiet between the dispatch and the eventual re-fire.

3. **Inconsistent attempt counter.** A queued failure-retry that's stuck behind capacity for 10 minutes will still have `attempt == 1` on the next dispatch, even though it has fired the recheck timer ~6000 times. Operators inspecting `/api/v1/state` will see the stale attempt count.

## Acceptance criteria

[...]

PR #313, opened 2026-05-23
fix(orchestrator): align capacity-full retry with upstream handle_active_retry (#306)

hey-yvan:
When a fired failure-retry timer found global or per-state capacity full,
the orchestrator armed a 100 ms recheck timer (rearmRetryCapacityRecheck)
instead of rescheduling through the configured backoff. That diverged
from upstream handle_active_retry (orchestrator.ex:1142-1161), which
calls schedule_issue_retry with attempt+1 and "no available orchestrator
slots" metadata.

Replace rearmRetryCapacityRecheck with rescheduleRetryCapacityFull, which
records a typed RuntimeEventFailed and submits a followup that calls
ScheduleRetry(attempt+1, "no available orchestrator slots") — same shape
as retryPollFailedOp's fetch-error branch. The scheduler's exponential
backoff (SPEC §8.4) and max_retry_backoff_ms ceiling now apply to
capacity pressure, the attempt counter advances on every recheck so
/api/v1/state surfaces accurate retry progress, and operators see the
slot pressure as runtime events.

Replace TestRetryFire_CapacityDeferralUsesShortRecheckDelay (which pinned
the 100 ms behavior) with two conformance tests covering both the global
cap and per-state cap paths; update the comment in
TestRetryFireAfterFetch_RefreshedIdentifierSurfacesInSnapshot to match
the new reschedule flow (the refreshed identifier still surfaces via the
followup's new entry).

### DIFF
--- a/internal/orchestrator/actor.go
+++ b/internal/orchestrator/actor.go
@@ -89,7 +89,6 @@ type RetryScheduler struct {
 	MaxBackoff time.Duration
 }
 
-const retryCapacityRecheckDelay = 100 * time.Millisecond
 const continuationRetryDelay = time.Second
 
 // retryFetchTimeout caps the SPEC §16.6 candidate-fetch call a fired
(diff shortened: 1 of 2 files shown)


hey-yvan:
Superseded by #311 (`b5a6c97`), which already landed on `main` and fixes the same SPEC §16.6 capacity-full retry divergence with the same upstream `handle_active_retry` (orchestrator.ex:1142-1161) alignment.

The merged version covers everything this PR proposes, plus three things this PR is missing:

1. **Shutdown guard.** `capacityDeferRetry` short-circuits with `if o.runCtx.Err() != nil { return nil }` so a fired retry tripping the cap during orchestrator shutdown doesn't leak a misleading `"no available orchestrator slots"` `RuntimeEventFailed` whose follow-up `ScheduleRetry` would silently fail to submit. The `rescheduleRetryCapacityFull` helper here has no such guard.
2. **DEVIATIONS.md D27 row.** #311 added the row marking this realignment Closed against [#306](https://github.com/xrf9268-hue/aiops-platform/issues/306). This branch is based on a stale `main` that pre-dates the row, so rebasing would just re-add what's already there.
3. **More robust test assertions.** #311's tests match the error via `strings.Contains(..., "no available orchestrator slots")`; this PR pins exact equality, which would over-fit if the error string ever grows a suffix.

The one extra line here — `entry.Timer = nil` before returning the follow-up — is a no-op in practice: the timer that fired is what scheduled this `retryFireOp` in the first place, so it's already done. The subsequent `scheduleRetryOp`'s call to `OrchestratorState.ScheduleRetry` handles the replace either way.

Closing as duplicate; no new issue needed.

---

PR #333, opened 2026-05-23
fix(tracker/linear): drop non-existent customFieldValues field from poll query (#326)

### LINKED ISSUE #326: [P0][tracker/linear] ListIssuesByStates queries non-existent `customFieldValues` field, every Linear poll returns 400
## Summary

`internal/tracker/linear.go:158` requests `customFieldValues(first: 50) { ... }` in
`ListIssuesByStates`. Linear's current GraphQL schema does **not** expose a
`customFieldValues` field on `Issue`. Every poll request therefore returns HTTP
`400 Bad Request` with `GRAPHQL_VALIDATION_FAILED`, and the worker cannot poll
**any** Linear-backed deployment, single-service or multi-service.

## Repro (no fix applied)

End-to-end against a freshly-created Linear project + issue (org `aishuode`,
team `AIS`, project slug `01d2b95da4b2`, issue `AIS-10` in `Todo`), using
`tracker.kind: linear` and a valid `LINEAR_API_KEY`:

```
2026/05/23 23:08:42 event=startup_reconcile_active_fetch_failed error="linear request failed: 400 Bad Request" note="SPEC §8.6: log and continue; no cleanup performed"
2026/05/23 23:08:42 event=tracker_poll_error error="linear request failed: 400 Bad Request"
2026/05/23 23:08:48 event=tracker_poll_error error="linear request failed: 400 Bad Request"
2026/05/23 23:08:53 event=tracker_poll_error error="linear request failed: 400 Bad Request"
```

Reproducing the worker's exact GraphQL payload against `https://api.linear.app/graphql`:

```json
{
  "errors": [{
    "message": "Cannot query field \"customFieldValues\" on type \"Issue\".",
    "locations": [{"line": 8, "column": 7}],
    "extensions": {"code": "GRAPHQL_VALIDATION_FAILED", "http": {"status": 400}}
  }]
}
```

Linear schema introspection (`query{__type(name:"Issue"){fields{name}}}`):

```
custom* fields on Issue: ['customerTicketCount']
```

No `customFieldValues`, `customField`, or anything similar is exposed on
`Issue` in the live Linear API. This is consistent across the public developer
docs — custom fields are not part of the GraphQL surface today.

## Affected code

- `internal/tracker/linear.go:143-163` — `ListIssuesByStates` GraphQL query
  includes `customFieldValues(first: 50) { nodes { customField { name } value } }`.
- `internal/tracker/linear.go:191-198`, `:240-246` — Go struct + ingestion of
  the (non-existent) `customFieldValues` payload.

Other Linear queries (`FetchIssueStatesByIDs`, blocker traversal, state
mutation, comment create, `findStateID`) do **not** request this field and
keep working in isolation.

## Blast radius

- 100% of Linear polling fails on startup reconciliation and on every poll
  tick. No issue ever reaches dispatch.
- `cmd/worker` (`tracker.kind: linear`) is non-functional end-to-end.
- `cmd/linear-poller` (legacy queue path) shares the same client and is
  affected identically.
- Unit tests pass because the test suite mocks the GraphQL transport with
  hand-crafted JSON, so the schema mismatch is invisible to CI.

## When introduced

PR #144 / commit `d6bba94 feat: route Linear issues by service config` —
2026-05-19. The route-predicate ingestion expected a `customFieldValues`
surface that Linear's API does not provide.

## Suggested direction

Custom fields are referenced only by the multi-service routing predicates
(`services[].tracker.custom_fields`, `internal/workflow/loader.go:338`).
Options, ordered by smallest blast-radius first:

1. **Drop `customFieldValues` from the polling query** and treat
   `custom_fields` route predicates as unsupported (validation error at
   `WORKFLOW.md` load, plus a DEVIATIONS entry). This unblocks every
   non-`custom_fields` deployment immediately.
2. Replace the custom-field surface with whatever Linear exposes today
   (e.g. labels, project metadata, or per-team attributes) before re-enabling
   the route predicate.

Either way, an integration test that exercises the real Linear schema (or a
contract test using `linear-api-types`) would catch this class of regression
— the current mock-based tests will not.

## Environment

- Go: from `golang:1.25-bookworm` (CI parity).
- Linear org/team: live API, not a sandbox.
- Branch: `main` at `608ae16` (HEAD as of 2026-05-23).

hey-yvan:
## Summary

Fixes #326 — Linear tracker polling is non-functional today; every poll returns HTTP 400 `GRAPHQL_VALIDATION_FAILED` because `internal/tracker/linear.go:158` requests `customFieldValues(first: 50) { customField { name } value }`, which Linear's GraphQL schema does not expose.

## Authoritative API confirmation

Live introspection against `https://api.linear.app/graphql` on 2026-05-23 with a Pro-tier personal API key (`Yy B` / `aishuode` org):

```
Issue type has 83 fields total.
Fields containing 'custom' (case-insensitive):
  - customerTicketCount : Int
Fields containing 'field' (case-insensitive):
  (none)
Types containing 'custom' (Schema-wide):
  - CustomView*, Customer*, IssueTitleSuggestionFromCustomerRequestPayload, ...
  (no IssueCustomField* anywhere)
Query root fields containing custom/field:
  - customViews, customView, customViewDetailsSuggestion, customViewHasSubscribers,
    customerNeeds, customerNeed, issueTitleSuggestionFromCustomerRequest,
    customers, customer, customerStatuses, customerStatus, customerTiers, customerTier
  (no issueCustomFields* root either)
```

No `customField*`, no `*FieldValue`, no `*FieldValues` exists on Issue or its connections. Renaming to `custom_fields` does not help either — the schema does not expose any custom-field surface on Issue under any name.

The upstream Elixir reference [`elixir/lib/symphony_elixir/linear/client.ex`](https://github.com/openai/symphony/blob/main/elixir/lib/symphony_elixir/linear/client.ex) (the SPEC tiebreaker per AGENTS.md) **also omits any custom-field fragment** for the same reason.

## Changes

- Remove `customFieldValues` GraphQL fragment + matching Go struct from `ListIssuesByStates`. Comment in `linear.go` documents the introspection result.
- `Issue.CustomFields` stays in the type (consumed by `services[]` routing predicate matching in `worker/reconcile.go` and `orchestrator/poller.go`). Linear always returns nil; the matching loops handle nil safely.
- `workflow/loader.go` now rejects `services[].tracker.custom_fields:` at workflow load time with a clear error pointing to #326 (the predicate can never match a Linear-polled issue). Removes `custom_fields` from the predicate list in the "must define at least one predicate" error.
- Replace `TestListIssuesByStatesMapsNonStringCustomFieldValues` with `TestListIssuesByStatesUsesLinearSupportedQueryShape` that pins the regression by asserting the query MUST NOT request `customFieldValues` or `customField {`.
- Update `TestListIssuesByStatesMapsSpecDomainFields` to drop the customFieldValues fragment from the assertion and add the same negative assertion.

## Verification

- ✅ `gofmt -l .` clean
- ✅ `go test -race -count=1 -timeout 5m ./...` — 14/14 packages green
- ✅ Mutation test: re-introducing `customFieldValues` into the query (one-line `sed` patch) makes the new test fail with the exact `customFieldValues` negative-assertion message; restoring the fix makes it pass again. Confirms the new test catches the regression.
- ✅ Live verified against `tracker.kind: linear` worker poll loop on org `aishuode` (project slug `01d2b95da4b2`, issue `AIS-10`): after this patch, `event=reconcile_end status=ok active_issues=1` followed by full workspace prep → mock runner → verify → `Succeeded` chain. Without the patch, every poll logs `linear request failed: 400 Bad Request`.

## Test plan

- [x] `gofmt -l .` clean
- [x] `go vet ./...` clean
- [x] `go test -race -count=1 -timeout 5m ./...` all green
- [x] Mutation: re-inject `customFieldValues` → new tests FAIL; restore → PASS
- [x] Live Linear API tested end-to-end with patched binary

### DIFF
--- a/docs/workflows/services-routing.md
+++ b/docs/workflows/services-routing.md
@@ -33,8 +33,6 @@ services:
       project_slug: api-platform
       team_key: ENG
       labels: [backend]
-      custom_fields:
-        Runtime: go
 ```
 
 Each service object has these fields:
@@ -49,7 +47,10 @@ Each service object has these fields:
 | `tracker.project_slug` | string | conditional | top-level `tracker.project_slug` if omitted | Linear project slug predicate and polling scope for this service. |
 | `tracker.team_key` | string | no | none | Linear team key predicate. |
 | `tracker.labels` | string array | no | empty | Linear labels that must be present on the issue. |
-| `tracker.custom_fields` | string map | no | empty | Linear custom field name/value predicates that must match the issue. |
+
+`tracker.custom_fields` is currently rejected at workflow load: Linear's GraphQL
+schema does not expose Issue custom fields, so the predicate can never match a
+polled issue. See [#326](https://github.com/xrf9268-hue/aiops-platform/issues/326).
 
 The top-level `tracker.project_slug` may provide the polling scope for services
 that omit `services[].tracker.project_slug`. A service-only Linear workflow may
@@ -68,7 +69,7 @@ Validation happens when `WORKFLOW.md` is loaded:
 - Every service requires `name` and `repo.clone_url`.
 - Service names must be unique case-insensitively.
 - Every Linear service route must define at least one route predicate among
-  `project_slug`, `team_key`, `labels`, or `custom_fields`.
+  `project_slug`, `team_key`, or `labels`.
 - Every Linear service route must have a project slug either in
   `services[].tracker.project_slug` or top-level `tracker.project_slug`, so the
   poller has an explicit project scope.
--- a/examples/WORKFLOW.md
+++ b/examples/WORKFLOW.md
@@ -64,8 +64,8 @@ polling:
 #       project_slug: api-platform
 #       team_key: ENG
 #       labels: [backend]
-#       custom_fields:
-#         Runtime: go
+# `tracker.custom_fields:` is currently rejected at workflow load —
+# Linear's GraphQL schema does not expose Issue custom fields. See #326.
 
 workspace:
   root: ~/aiops-workspaces/personal
--- a/internal/orchestrator/poller_test.go
+++ b/internal/orchestrator/poller_test.go
@@ -1234,6 +1234,12 @@ func TestFilterEligibleCandidatesUsesOnlyConfiguredTerminalSet(t *testing.T) {
 	}
 }
 
+// TestSelectRoutedCandidatesMatchesLinearProjectTeamLabelAndCustomField pins
+// the routing-predicate matching primitive at the orchestrator layer. The
+// custom_fields predicate is rejected at workflow load (#326) because
+// Linear's GraphQL schema does not expose Issue custom fields, but the
+// matching loop itself is still defensive — keeping the primitive intact
+// preserves a deprecation runway if Linear ever surfaces the field.
 func TestSelectRoutedCandidatesMatchesLinearProjectTeamLabelAndCustomField(t *testing.T) {
 	cfg := workflow.Config{Services: []workflow.ServiceConfig{
 		{
--- a/internal/tracker/linear.go
+++ b/internal/tracker/linear.go
@@ -140,6 +140,14 @@ func (c *LinearClient) ListIssuesByStates(ctx context.Context, states []string) 
 	if len(nonEmpty) == 0 {
 		return nil, nil
 	}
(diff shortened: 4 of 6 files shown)


hey-yvan:
@codex review

hey-yvan:
@codex review

PR #334, opened 2026-05-23
fix(runner/codex): use codex `granular:` policy instead of obsolete `reject:` (#329)

### LINKED ISSUE #329: [P0][runner/codex_app_server] default `approval_policy.reject.*` payload rejected by codex JSON-RPC: `unknown variant \`reject\``
## Summary

When `agent.default: codex-app-server` is selected, `internal/runner/codex_app_server.go:262` sends the raw `workflow.TrackerConfig.Codex.ApprovalPolicy` value to codex's `thread/start` JSON-RPC. The default value (`internal/workflow/loader.go:701`, `internal/workflow/config.go:551`) is:

```go
ApprovalPolicy: map[string]any{"reject": map[string]any{"sandbox_approval": true, "rules": true, "mcp_elicitations": true}}
```

Codex CLI `0.133.0` rejects this with:

```json
{"error":{"code":-32600,"message":"Invalid request: unknown variant `reject`, expected one of `untrusted`, `on-failure`, `on-request`, `granular`, `never`"},"id":2}
```

i.e. `reject` is an internal vocabulary used by aiops-platform's `protocolServerRequestResult` / `approvalRuleRequiresReview` (lines 871–948 of the same file) to decide *what to do with server-requested approvals*, but it is **not** a codex wire-format value. Sending it directly fails every `thread/start` request, which makes the entire `codex-app-server` agent non-functional with default config.

## Repro

```yaml
agent:
  default: codex-app-server
codex:
  command: codex app-server
  profile: safe
# (no explicit approval_policy)
```

Result on next poll tick:

```
event=startup_failed phase:thread/start
  error:rpc error: map[code:-32600 message:Invalid request: unknown variant `reject`, expected one of `untrusted`, `on-failure`, `on-request`, `granular`, `never`]
event=runner_end ... error: codex app-server thread/start: rpc error: ...
```

Workaround: explicitly set `codex.approval_policy: never` in `WORKFLOW.md` — the harness then passes a codex-accepted value through and the run proceeds (verified: `event=session_started` fires, codex enters a real turn).

## Affected code

- `internal/workflow/loader.go:701` — default install of the `reject` policy on workflows that omit `codex.approval_policy`.
- `internal/workflow/config.go:551` — `defaultCodexConfig` with the same `reject` payload.
- `internal/runner/codex_app_server.go:262` — verbatim forward of `Codex.ApprovalPolicy` to codex `thread/start`.

## Suggested direction

Two clean fixes; either alone unblocks the runner:

1. **Translate the internal `reject` schema at the boundary**: keep the `reject` policy as an internal-only filter for `protocolServerRequestResult` (it serves a real safety purpose, line 923-925), and convert it to `never` (the codex equivalent of "never auto-approve, always deny") before sending to codex `thread/start`. The internal harness-hardening behavior (reject MCP elicitations / rule prompts) is preserved client-side.
2. **Change the default to a codex-accepted variant** (`never` or `untrusted`) and re-implement the harness-hardening filter to layer on top of that without sending the unsupported field shape.

The current state means `agent.default: codex-app-server` cannot be enabled without a manual `WORKFLOW.md` override that is not documented in `examples/WORKFLOW.md` or the README.

## Environment

- aiops-platform: `main` at `608ae16` (2026-05-23)
- codex CLI: `0.133.0`
- Linear org/team: live API; harness configured per #326 patch

hey-yvan:
## Summary

Fixes #329 — `agent.default: codex-app-server` is unusable with default config: every `thread/start` returns `-32600 Invalid request: unknown variant \`reject\`, expected one of \`untrusted\`, \`on-failure\`, \`on-request\`, \`granular\`, \`never\``.

## Authoritative source — codex protocol

Codex's `AskForApproval` enum is the source of truth. The current shape lives at:

- [`codex-rs/protocol/src/protocol.rs` `pub enum AskForApproval`](https://github.com/openai/codex/blob/main/codex-rs/protocol/src/protocol.rs)
- [`codex-rs/app-server-protocol/src/protocol/v2/shared.rs`](https://github.com/openai/codex/blob/main/codex-rs/app-server-protocol/src/protocol/v2/shared.rs)

The variant **`Reject` was renamed to `Granular`** with field polarity inverted in [codex commit b7dba72db (PR openai/codex#14516, 2026-03-12)](https://github.com/openai/codex/commit/b7dba72db). The rename diff:

```rust
-    /// Fine-grained rejection controls for approval prompts.
-    ///
-    /// When a field is `true`, prompts of that category are automatically
-    /// rejected instead of shown to the user.
-    Reject(RejectConfig),
+    /// Fine-grained controls for individual approval flows.
+    ///
+    /// When a field is `true`, commands in that category are allowed. When it
+    /// is `false`, those requests are automatically rejected instead of shown
+    /// to the user.
+    #[strum(serialize = "granular")]
+    Granular(GranularApprovalConfig),
```

That is: the same JSON shape `{"sandbox_approval": true}` used to mean **reject**, now means **allow**. Sending the obsolete `reject:` wrapper makes codex emit `unknown variant`.

The upstream Elixir Symphony reference still ships the pre-rename default ([`elixir/lib/symphony_elixir/config/schema.ex:162-170`](https://github.com/openai/symphony/blob/main/elixir/lib/symphony_elixir/config/schema.ex#L162)); aiops-platform inherited the bug by porting from it. Treating codex's source as canonical (AGENTS.md "When SPEC text is ambiguous, the reference's behavior is the tiebreaker" — extended here because the protocol contract itself moved).

## Changes

- **`internal/workflow/config.go` + `internal/workflow/loader.go`**: default `ApprovalPolicy` is now
  ```yaml
  granular:
    sandbox_approval: false
    rules: false
    skill_approval: false
    request_permissions: false
    mcp_elicitations: false
  ```
  Semantically equivalent to the old `reject: all-true` intent (refuse every approval category), expressed in the codex-accepted enum variant.
- **`internal/runner/codex_app_server.go`**:
  - Add `codexWireApprovalPolicy()` wrapping both `thread/start` and `turn/start`. Pass-through for strings (`"never"`/`"untrusted"`/`"on-failure"`/`"on-request"`) and `{"granular": {...}}`. Migrate legacy `{"reject": {flag:bool}}` to `{"granular": {flag:!bool}}` so pre-rename WORKFLOW.md files keep their original intent across the polarity flip. Unknown shapes pass through verbatim so codex's own validation error reaches the operator.
  - Drop the now-dead `approvalRuleRequiresReview` helper and the `reject:` branch from `autoApproveRequest`. The harness handler now mirrors codex's enum exactly.
- **`internal/runner/codex_app_server_test.go`**: new `TestCodexWireApprovalPolicy` pinning every translation case with a long doc comment that names the codex commit + PR. Existing matrix label updated to document that the harness no longer special-cases `reject:` — the boundary translator handles back-compat.

## Why this won't be reverted

Long doc comments on `codexWireApprovalPolicy` and the test explain the codex commit (`b7dba72db`), the PR (`openai/codex#14516`), the date (2026-03-12), the polarity flip, and warn future maintainers to check codex's enum definition before "fixing" the translation back. Codex source is best-practice; the upstream Elixir reference is downstream of it and currently lags this change.

## Verification

- ✅ `gofmt -l .` clean.
- ✅ `go vet ./...` clean.
- ✅ `go test -race -count=1 -timeout 5m ./...` 14/14 green.
- ✅ Mutation: replacing `codexWireApprovalPolicy` with identity makes the legacy-translation cases of `TestCodexWireApprovalPolicy` fail with the exact polarity-flip diff; restoring makes them pass.
- ✅ Live verified end-to-end against `agent.default: codex-app-server` + real Linear API. Before the patch: `thread/start` returns `-32600 unknown variant \`reject\``. After the patch (default config, no operator override): `event=session_started thread_id=… turn_id=… session_id=…` fires, the codex session enters real turns.

## Test plan

- [x] `gofmt -l .` clean
- [x] `go vet ./...` clean
- [x] `go test -race -count=1 -timeout 5m ./...` all green
- [x] Mutation: revert translator to identity → translation tests FAIL; restore → PASS
- [x] Live `codex-app-server` runner test against real Linear issue (`event=session_started` confirmed)

### DIFF
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -259,7 +259,7 @@ func (c *appServerClient) run(ctx context.Context, in RunInput, prompt string) e
 	}
 
 	threadResult, err := c.request(ctx, "thread/start", map[string]any{
-		"approvalPolicy": in.Workflow.Config.Codex.ApprovalPolicy,
+		"approvalPolicy": codexWireApprovalPolicy(in.Workflow.Config.Codex.ApprovalPolicy),
 		"sandbox":        in.Workflow.Config.Codex.ThreadSandbox,
 		"cwd":            in.Workdir,
 		"dynamicTools":   appServerDynamicToolSpecs(in.Workflow.Config),
@@ -296,7 +296,7 @@ func (c *appServerClient) run(ctx context.Context, in RunInput, prompt string) e
 			"input":          input,
 			"cwd":            in.Workdir,
 			"title":          appServerTurnTitle(in),
-			"approvalPolicy": in.Workflow.Config.Codex.ApprovalPolicy,
+			"approvalPolicy": codexWireApprovalPolicy(in.Workflow.Config.Codex.ApprovalPolicy),
 			"sandboxPolicy":  in.Workflow.Config.Codex.TurnSandboxPolicy,
 		})
 		if err != nil {
@@ -904,6 +904,19 @@ func protocolServerRequestAutoApproved(method string, approvalPolicy any) bool {
 	}
 }
 
+// autoApproveRequest reports whether protocolServerRequestResult should
+// auto-approve a server-side approval prompt that codex sent the harness.
+// The decision tracks codex's own `AskForApproval` semantics (codex-rs
+// protocol/src/protocol.rs):
+//
+//   - "never"                     — codex never asks; if a prompt still
+//     surfaces, auto-approve it (matches codex's "failures returned to the
+//     model" intent).
+//   - "on-failure"                — codex asks only on failure; the harness
+//     auto-approves to keep the unattended loop moving.
+//   - "untrusted" / "on-request"  — operator-supervised modes; decline.
+//   - {"granular": {...}}         — per-method bool flags where TRUE means
+//     ALLOW (codex semantics since #14516, b7dba72db, 2026-03-12).
 func autoApproveRequest(method string, approvalPolicy any) bool {
 	if policy, ok := approvalPolicy.(string); ok {
 		switch strings.ToLower(strings.TrimSpace(policy)) {
@@ -920,19 +933,7 @@ func autoApproveRequest(method string, approvalPolicy any) bool {
 	if granular, ok := policy["granular"].(map[string]any); ok {
 		return approvalRuleAllowsRequest(granular, method)
 	}
-	if reject, ok := policy["reject"].(map[string]any); ok {
-		return !approvalRuleRequiresReview(reject, method)
-	}
 	return false
-}
-
-func approvalRuleRequiresReview(rules map[string]any, method string) bool {
-	if method == "item/permissions/requestApproval" {
-		return approvalRuleEnabled(rules, "request_permissions") ||
-			approvalRuleEnabled(rules, "sandbox_approval") ||
-			approvalRuleEnabled(rules, "rules")
-	}
-	return approvalRuleEnabled(rules, "sandbox_approval") || approvalRuleEnabled(rules, "rules")
 }
 
 func approvalRuleAllowsRequest(rules map[string]any, method string) bool {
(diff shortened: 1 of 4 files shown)


hey-yvan:
@codex review

PR #339, opened 2026-05-24
fix(orchestrator): SPEC §18.1 active-transition workspace cleanup for terminal runs (#331)

### LINKED ISSUE #331: [P2][orchestrator] §18.1 "workspace cleanup for terminal issues — active transition" not honored; only startup sweep cleans
## Summary

SPEC §18.1 REQUIRED includes:

> Workspace cleanup for terminal issues (startup sweep + active transition)

Two paths are required:

1. **Startup sweep** — implemented and verified. On worker boot, terminal-state issues' workspaces are removed and `before_remove` fires (`event=reconcile_workspace action=remove reason=terminal state=Done`).
2. **Active transition** — **not** implemented. When a *currently-running* issue transitions to a terminal state mid-run (§8.5 active-run reconciliation), the worker correctly cancels the run, but it does **not** remove the workspace directory or fire the `before_remove` hook. The workspace lingers until the next worker restart's startup sweep.

## Repro

```yaml
agent: { default: mock, max_concurrent_agents: 1 }
verify: { commands: [ "sleep 30 && echo ok" ] }
workspace:
  root: /tmp/aiops-e2e/workspaces
  hooks:
    before_remove:
      commands: ['echo "before_remove fired" >> /tmp/log.txt']
```

1. Start worker; `AIS-10` (in `Todo`) is picked up, run enters `Finishing → verify` with the 30s sleep.
2. While the verify sleep is running, transition Linear `AIS-10` → `Done`.
3. Worker logs:
   - `event=verify_end status=canceled error="context canceled" exit_code=-1`
   - `event=run_phase_transition msg="Failed"` (was `Finishing → Failed`)
   - `counts.running: 0` within ~3s of the Linear flip
4. Observe: `before_remove` never fires; `/tmp/aiops-e2e/workspaces/e2e/testrepo/linear_issue/AIS-10/` still on disk.
5. Restart the worker → startup reconcile sees `AIS-10` is `Done`, removes the workspace AND fires `before_remove`. Confirms the cleanup machinery exists; the active-transition path just doesn't call it.

## Why this matters

- An operator who never restarts the worker accumulates terminal workspaces on disk indefinitely.
- `before_remove` hooks intended for "tear-down" semantics (push notifications, archive logs, free GPUs, release leased credentials) silently no-op for the most common scenario: an in-flight run getting cancelled by tracker state changes.
- Failure mode is invisible — orchestrator state correctly shows the run as finished; only `du` on the workspace root reveals the leftover.

## Affected code

- `internal/orchestrator/actor.go:~1570` — `FinishRunReconciledCancelled` removes the entry from `Running`/`Claimed` but does not invoke workspace cleanup or hooks.
- Compare with the startup-reconcile path in `internal/orchestrator/recover.go` / startup reconciliation logic, which DOES call the `before_remove` hook before removing the workspace.

## Suggested direction

Have the §8.5 active-transition cancel path call the same workspace-remove sequence the startup sweep uses (`before_remove` hook → remove workspace dir → emit `event=reconcile_workspace action=remove reason=terminal`).

## Environment

- aiops-platform: `main` at `608ae16`
- Verified against `mock` runner + Linear (org `aishuode`, slug `01d2b95da4b2`, issue `AIS-10`).

hey-yvan:
Closes #331.

## Problem
SPEC §18.1 requires terminal-issue workspace cleanup on **both** the startup sweep **and** the active transition. Only the startup sweep was implemented. When a *running* issue moved to a terminal tracker state mid-run, the §8.5 reconcile pass cancelled the worker but never removed its workspace or fired `before_remove` — the dir lingered until the next worker restart, and tear-down hooks (push notifications, archive logs, freed GPUs/leased credentials) silently no-op'd for the most common cancel scenario.

## Fix
Mirror upstream `orchestrator.ex` `terminate_running_issue(state, id, cleanup_workspace)`: clean the workspace **only on a terminal transition** (`cleanup_workspace=true`), not on a merely-inactive / route-change cancel (which keep the dir for reuse).

- **`internal/worker`**: extract `RemoveIssueWorkspace` (before_remove hook → `SafeRemove` → `reconcile_workspace` remove event) as the shared removal routine; the startup sweep's `removeWorkspace` now delegates to it (behavior unchanged). This mirrors upstream sharing one `Workspace.remove_issue_workspaces` across both paths.
- **`internal/orchestrator`**: add a nil-safe `WorkspaceCleaner` dependency. The inactive reconcile op flags terminal-state running entries (`RunningEntry.ReconcileCleanupWorkspace`) and refreshes the stored issue to the terminal state; finalize fires the cleaner on a **bounded, panic-safe followup after the worker has exited** (so the dir is no longer in use). `RuntimeDispatcher` implements the cleaner against the **live workflow snapshot's** hook config (honors hot reload).

## Scope / known limitation
- **Non-routing (default) deployments — the issue's exact repro — are fully covered.** A terminal mid-run transition fires `before_remove` and emits `reconcile_workspace action=remove reason=terminal`.
- **Routing mode** cancels terminal runs in an earlier pass (`reconcileTrackerIssuesOp`) that lacks terminal-state info, so it still falls back to the startup sweep. Documented inline at `poller.go` and tracked as a follow-up issue.

## Tests
- `TestReconcileTerminalRunFiresActiveWorkspaceCleanup` — terminal mid-run cancel → cleaner invoked with the run's path, `reason=terminal`, `state=Done`, after worker exit; run cleared from state.
- `TestReconcileInactiveNonTerminalRunKeepsWorkspace` — inactive (non-terminal) cancel releases the run but does **not** clean the workspace (upstream gating).
- **Mutation-tested** both: dropping the cleanup flag fails the positive test; removing the terminal gate fails the negative test.

`go build ./...`, `go vet`, `go test ./...` all green; `gofmt` clean.


hey-yvan:
@codex review

hey-yvan:
@codex review

hey-yvan:
@codex review

hey-yvan:
@codex review

hey-yvan:
@codex review

hey-yvan:
@codex review

PR #353, opened 2026-05-24
fix(orchestrator): SPEC §18.1 active-transition cleanup through routing reconcile (#340)

### LINKED ISSUE #340: [orchestrator] §18.1 active-transition workspace cleanup not wired through the service-routing reconcile pass
## Summary

Follow-up from #331 / PR #339. The SPEC §18.1 active-transition workspace cleanup (before_remove → remove dir → `reconcile_workspace action=remove reason=terminal`) landed in #339 for the **default / non-routing** deployment, but **routing mode is not covered**.

## Why routing mode is missed

In `internal/orchestrator/poller.go` `reconcileTick`, when `p.routing != nil`:

1. Non-active refreshed issues (including terminal ones) are deleted from `activeIssuesByID` (the `delete(activeIssuesByID, id)` block).
2. `ReconcileTrackerIssuesAndWait(ctx, activeIssuesByID, activeStateKeys, …)` then cancels the running issue because it is absent from the active set — but this pass has **no terminal-state set**, so it cannot set `RunningEntry.ReconcileCleanupWorkspace`. It also `waitForReconciledWorkers`, so the entry is typically finalized (removed from `st.Running`) before the later inactive pass.
3. `ReconcileInactiveTrackerIssuesAndWait(…, terminalStates, …)` — the pass that DOES carry the terminal set and performs the cleanup (#339) — then no longer sees the already-finalized entry.

Net effect: in routing mode a running issue that goes terminal mid-run is cancelled correctly, but its workspace is reclaimed only by the next startup sweep — the exact gap #331 describes, still open for routed runs.

## Upstream reference

Upstream is single-project (no service routing), so `reconcile_issue_state` (`elixir/lib/symphony_elixir/orchestrator.ex:396-418`) always reaches `terminate_running_issue(state, issue.id, true)` for a terminal running issue. The aiops-platform routing extension splits cancellation across two passes; the terminal-cleanup signal must reach whichever pass actually cancels a routed terminal run.

## Affected code

- `internal/orchestrator/poller.go` — routing branch (`p.routing != nil`) + the `delete(activeIssuesByID, id)` loop. See the `NOTE(#331)` comment added in #339.
- `internal/orchestrator/actor.go` — `reconcileTrackerIssuesOp` has no `terminalStates` and does not flag `ReconcileCleanupWorkspace`.

## Suggested direction

Thread the terminal-state set into the routing cancel decision (e.g. pass `terminalStates` to `reconcileTrackerIssuesOp` and supply the refreshed terminal issues so it can flag `ReconcileCleanupWorkspace`), OR reorder so the terminal-aware inactive pass runs before the routing cancel for terminal issues. Either way, reuse the `WorkspaceCleaner` seam added in #339.

## Acceptance criteria

- A routed running issue moved to a terminal state mid-run fires `before_remove` and emits `reconcile_workspace action=remove reason=terminal` mid-run (not at next startup), matching the non-routing path.
- A routed run cancelled for a route change / non-terminal inactive state still **keeps** its workspace.
- Unit test in the routing configuration covering both.

hey-yvan:
## Summary

Follow-up from #331 / PR #339. The SPEC §18.1 active-transition workspace cleanup landed in #339 for the **default / non-routing** path, but **routing mode was not covered**.

In routing mode the routing-aware reconcile pass (`reconcileTrackerIssuesOp`) cancels — and waits for — a running issue that goes terminal mid-run *before* the terminal-aware inactive pass (`reconcileInactiveTrackerIssuesOp`) can see it. That pass carried no terminal-state set, so it never flagged `ReconcileCleanupWorkspace`; the workspace was reclaimed only by the next startup sweep. The routed **blocked** path also used a bare `os.RemoveAll` — no `before_remove` hook, no `reconcile_workspace` event, no terminal gating.

Upstream is single-project (no routing) and reaches `terminate_running_issue` with cleanup in one pass (`orchestrator.ex:396-418`); the aiops routing extension splits cancellation across two passes, so the terminal-cleanup signal must reach whichever pass actually cancels the routed terminal run.

## Changes

- Thread `terminalStates` + the refreshed post-transition issue states into `reconcileTrackerIssuesOp`.
- Running terminal cancel → flag `ReconcileCleanupWorkspace` so finalize fires `before_remove` + `reconcile_workspace action=remove reason=terminal` mid-run.
- Blocked terminal release → route through the shared `WorkspaceCleaner` (reusing the `terminalWorkspaceForCleanup` seam from #339) instead of bare `os.RemoveAll`.
- Route-change / non-terminal inactive cancel → **keep** the workspace for reuse.
- Unify both reconcile passes' off-actor followup; delete the dead `os.RemoveAll` path (and now-unused `os`/`log` imports).

## Acceptance criteria

- [x] A routed running issue moved to terminal mid-run fires `before_remove` + `reconcile_workspace reason=terminal` mid-run (not at next startup).
- [x] A routed run cancelled for route change / non-terminal inactive keeps its workspace.
- [x] A routed blocked issue moved to terminal fires `before_remove` + `reconcile_workspace reason=terminal`; a non-terminal inactive transition keeps the workspace.
- [x] Unit tests in the routing configuration covering all four cases.

## Test plan

- [x] 4 new regression tests (`TestReconcileRouting{TerminalRunFiresActiveWorkspaceCleanup,RouteChangeKeepsWorkspace,TerminalBlockedFiresWorkspaceCleanup,NonTerminalBlockedKeepsWorkspace}`), each mutation-verified (flip the terminal gate → corresponding test FAILs; restore → PASSes).
- [x] `gofmt -l` empty, `go mod tidy` no diff, `go vet ./...`, `go build ./...`.
- [x] `go test -race ./internal/orchestrator/` green.

Closes #340

https://claude.ai/code/session_01NgwhK9c7pVQ3MiymGe6FXC

---
_Generated by [Claude Code](https://claude.ai/code/session_01NgwhK9c7pVQ3MiymGe6FXC)_


hey-yvan:
@codex review

PR #495, opened 2026-05-29
refactor(workflow): extract config validation into validate.go, decomposed (#410)

hey-yvan:
## What & why

`validateConfig` was a 190-line, gocognit-~50 monolith — the #410 burn-down's named hotspot #2 (RunTask, hotspot #1, was split in #494). This is a **pure structural refactor**: no business logic, no API, no behavior change.

It does two coupled things:

1. **Decompose** `validateConfig` into a fixed-order orchestrator + per-section validators (tracker/repo, Linear routes, services, supported values, sandbox, server port, codex/claude, agent limits, timeouts), each a single nameable responsibility, all well under the 80-line / gocognit-10 budgets.
2. **Relocate** the whole validation family out of `loader.go` into a new `validate.go`. This brings `loader.go` from **1071 → 744 lines**, back under the 800-line file budget (AGENTS.md clean-code rule 7), and folds a stale doc comment (described `validateConfig` but sat above `hasExplicitServiceRoute`) into the orchestrator doc where it belongs.

The two are intentionally one PR: the function split alone would have pushed an already-over-budget `loader.go` (960) further past the file budget; the extraction is what keeps the refactor from regressing it.

## Behavior preservation

The validator slice runs in the **exact original order**, so for any `Config` the first failing check — and its error string — is identical to before. tracker.kind is still validated first (SPEC §6.4). All 42 `fmt.Errorf` strings were moved verbatim; the two flattened early-return blocks (`validateTrackerAndRepo` on `clone_url!=""`, `validateSandbox` on `network!=allowlist`) are logically equivalent to the original nested forms.

An adversarial 3-lens self-review (behavior-equivalence / clean-code / test-adequacy) found **zero behavior divergences**; its only actionable finding was the `loader.go` file budget, fixed by the extraction above.

## Tests

Added two characterization tests for the only two `validateConfig` branches that had **no coverage** and are easiest to mis-extract during a split:

- `TestLoadRejectsLinearServiceCustomFields` — Linear service `custom_fields` rejection (#326), reachable via `Load`.
- `TestValidateConfigRejectsUnsupportedTurnSandboxPolicyType` — the defensive `codex.turn_sandbox_policy.Validate` guard, reachable only via a programmatically constructed `Config` (the YAML decoder rejects bad types first), so it is pinned with a direct `validateConfig` call.

Both are **mutation-verified**: each fails when its production branch is neutered, then restored. The new `validate*` region has 100% block coverage; package statement coverage 78.3% → 79.2%.

## Verification

```
gofmt -l <tree>            # empty
go mod tidy                # go.mod/go.sum clean
go vet ./...               # clean
go build ./cmd/worker ./cmd/tui   # OK
go test -race -covermode=atomic ./internal/workflow/ ./internal/worker/   # ok
```
- workflow-package funlen/gocognit findings: **19 → 15**, no `validate*` function remains a finding.
- Full `./...` race suite is green **except** the pre-existing flaky `internal/runner.TestCodexAppServerInitializeTimeoutDiagnostics` (timing-sensitive process-start diagnostics, same area as #489/#476). Confirmed flaky on clean `main` (3/3 fails with this branch's changes stashed) and `internal/runner` is byte-identical to `main` on this branch — not a regression from this PR.

## Size

`size-gated: justified overage` — the diff is ~615 changed LOC because it relocates the validation family (mostly a move, reviewable as such). The relocation is what brings `loader.go` under the file budget and cannot be split from the decomposition without leaving `loader.go` worse. Net new LOC is +183.

## Scope

Refs #410 (does **not** close it — umbrella issue; remaining hotspots: `actor.go`, `cmd/worker/main.go`, `internal/runner/codex_app_server.go` file splits). Following the #494 precedent of referencing without closing.


hey-yvan:
@codex review

PR #501, opened 2026-05-29
refactor(runner): decompose awaitTurnCompletion into per-concern handlers (#499)

hey-yvan:
## What & why

First and highest-complexity target of #499: safely decompose
`(*appServerClient).awaitTurnCompletion` — the Codex JSON-RPC turn loop —
which was a single **gocognit 78 / ~114-line** function fusing four concerns
(stall-budget-aware reading, read-failure classification, message demux, and
server-request/notification handling). #410/#497/#500 only moved this function
into its own file; this PR does the internal decomposition #410 deferred as
behavior-sensitive.

Split along upstream Symphony's `receive_loop → handle_incoming →
handle_turn_method → maybe_handle_approval_request` seam
(`elixir/lib/symphony_elixir/codex/app_server.ex`):

| helper | responsibility |
|---|---|
| `readTurnMessage` | stall-budget deadline + `read_timeout_ms` suspension around one read |
| `classifyTurnReadError` | malformed-skip vs hard-decode vs stall-reclassify |
| `dispatchTurnMessage` | terminal `turn/*` vs tool-call vs other-message demux |
| `handleTurnMethod` | server-request vs notification fork |
| `handleServerRequest` | reply + declined/input-required surfacing |
| `handleTurnNotification` | input-required / stall / record fallback |

**Zero behavior change.** Read→classify→dispatch ordering, every
`lastTerminal` stall-clock reset, the `read_timeout_ms` suspend/restore, the
`turn/failed` vs `turn/cancelled` split, declined-approval → `InputRequiredError`,
and every recorded runtime event are identical. The pre-read budget-exhausted
`*StallError` (originally a direct `return`) now flows out of `readTurnMessage`
and through `classifyTurnReadError` unchanged — it is neither `DeadlineExceeded`
nor a read-timeout, so it falls through the reclassification guard with
`Cause==nil` intact.

## Acceptance criteria (#499, for `awaitTurnCompletion`)

- [x] **Characterization tests added before decomposing.** 18 tests pin the
  loop's input→(returned error, recorded runtime events, continuation signal)
  at the function boundary, driving it over an in-memory JSON-RPC stream (no
  subprocess), incl. branches the e2e suite did not exercise (input-required
  notifications, non-protocol decode error, auto-approve vs. decline fork).
  Green on the **original** function before the refactor; `-count=1 -race`
  twice, no flake.
- [x] **Decomposed into single-responsibility helpers**, preserving the BEAM
[...]

Issue #499, opened 2026-05-29
[Refactor] 复杂度热点续：actor apply / codex awaitTurnCompletion 等大函数的安全分解（#410 续）


hey-yvan:
## 背景

#410 把六个复杂度热点中的**文件/函数体量**问题分批 PR 解决了（RunTask #494、validateConfig #495、parser #496、actor.go #497、main.go #498、codex_app_server.go 拆分 PR）。其中 #410 表格里点名的三个函数（`RunTask` 244、`validateConfig` 186、`parseLinearGraphQLOperation` 145）已实质拆小；三个大文件已分文件。

但 #410 的 direction #3（"当前的大 `apply` 中 switch 分支可以替换为 event-type 驱动的小 handler"）和 direction #5（"把'事件 demux'和'超时管理'拆为独立类型"）的**函数内部分解**被**刻意推迟**了：那是行为型改动，发生在编排状态机 / Codex 协议这两条最高风险路径上，必须**先补 characterization 测试再拆**，不适合塞进"纯文件搬运"的 PR（#497/codex 拆分 PR 都是零行为改动、零新增 lint 违规）。

本 issue 跟踪这批 `gocognit` 热点的安全分解。

## 范围（按 `golangci-lint --enable gocognit` 实测，降序）

**internal/runner（Codex 协议路径）**
- `(*appServerClient).awaitTurnCompletion` — gocognit **78**（最高，~114 行；turn 完成 + stall 预算 + 失败/quota 分类高度耦合）
- `(*appServerClient).run` — **46**（主消息循环 + 超时 + 错误分类）
- `(CodexAppServerRunner).Run` — **29**
- `parseRetryAfterText` — 24 · `(*appServerClient).handleDynamicToolCall` — 20 · `hasShellSyntax` — 19 · `copyAllowlistedReasonFields` — 18 · `toSnakeCase` — 17 · `extractReasonFields` — 15 · `splitAppServerCommand` / `(*appServerClient).request` — 13

**internal/orchestrator（actor 状态机）**
- `(*finalizeRunOp).apply` — **47**（~196 行；终态 RunResult → 完成/失败退避/quota/continuation 路由）
[...]

PR #503, opened 2026-05-29
refactor(runner): decompose appServerClient.run into session/turn handlers (#499)

hey-yvan:
## What & why

Third #499 target (next-highest after #501/#502): decompose
`(*appServerClient).run` — the Codex app-server JSON-RPC session driver —
a single **gocognit 46 / ~145-line** function that fused per-run config
caching, the SPEC §10.1 handshake, and the whole turn loop. Split along
upstream Symphony's `start_session` / `start_turn` / `run_turn` seam
(`elixir/lib/symphony_elixir/codex/app_server.ex`):

| helper | responsibility |
|---|---|
| `initSession` | cache sinks + per-run codex config off `RunInput` |
| `startThread` | initialize → initialized → thread/start handshake |
| `runSingleTurn` | one turn: start, await, classify error, §16.5 refresh gate |
| `startTurn` | turn/start + turn-id resolution (startup-failed on turn 1) |
| `turnInput` | first-turn prompt vs continuation nudge |
| `recordFirstTurnStarted` | once-per-run `session_started` event + phase transition |
| `classifyTurnError` | stall / turn-timeout / passthrough mapping |

`run` now reads as `init → startThread → loop(runSingleTurn) → max-turns error`.

**Zero behavior change.** The handshake order and `recordStartupFailed`
phases, the `maxTurns<=0 → 20` default, the `turn==1` gates (session_started,
turn/start startup-failed), the `continueRun` reset/return, the §16.5
refresh-active gate, the per-turn timeout context, and `classifyTurnError`'s
use of the **outer** ctx for the deadline discriminator are all identical.
The `keepGoing` contract maps the original inline `return nil`/`return err`/
continue faithfully: `(false,nil)` = clean stop / §16.5 self-stop, `(false,err)`
= abort, `(true,nil)` = next turn.

## Acceptance criteria (#499, for `run`)

- [x] **Characterization first.** The existing app-server integration suite is
  the safety net; I validated it as non-placebo by mutating the **un-decomposed**
  `run` (refresh-active gate, continueRun gate, session-started-once gate,
  turn-timeout classification) and confirming each mutation fails the suite — no
  gap found, so no redundant test was added.
- [x] **Decomposed** into single-concern helpers; the read goroutine + per-turn
  `context.WithTimeout` are unchanged from origin/main.
- [x] **gocognit eliminated:** `run` **46 → eliminated**; every helper ≤10 and
  ≤80 lines (file 645 < 800). No new gocognit/funlen findings.
- [x] **Mutation-verified non-placebo** against the **decomposed committed**
  artifact: the same 4 routing mutations each fail the matching test; tree
  restored to HEAD after each.
- [x] **No new deviation, no external API change.** Blocking golangci gate
  (`staticcheck`/`unparam`/…) run locally before push: 0 issues.

Out of scope (separate #499 PRs): `(CodexAppServerRunner).Run` (29),
`parseRetryAfterText` (24), `handleDynamicToolCall` (20), and the orchestrator
`retryFireOp.apply` (30) / `dispatchOp.apply` (22) / reconcile hotspots.

## Verification

```
gofmt -l $(git ls-files '*.go')                                   # clean
golangci-lint ... --enable-only=...,staticcheck,unparam,unused ./...  # 0 issues (blocking gate)
go vet ./internal/runner/... ; go mod tidy                        # clean
go test -race ./...                                               # green (see note)
go build ./cmd/worker ./cmd/tui                                   # ok
```

Note: `TestCodexAppServerInitializeTimeoutDiagnostics` (untouched, initialize
path) fails **locally only** — the Python `codex` stub does not write its
started-log in this sandbox; it passes in CI. Not a regression.

## Review

Pre-push dual diff-only review (protocol §3): an independent
behavior-preservation pass and a `feature-dev:code-reviewer` pass both returned
**MERGE-READY** with no findings, each verifying handshake order, startup-failed
phases, the `turn==1` gates, `classifyTurnError`'s outer-ctx discriminator, the
§16.5 gate, and the `keepGoing` contract line-by-line. The second reviewer also
ran the blocking golangci gate independently (0 issues).

**Codex review remains unavailable** on both paths (expired local CLI token +
GitHub bot usage limit — see #501/#502). `@codex review` is triggered on this
head; if still limited, the independent dual review is the protocol §4
compensation.

Pre-existing observations the reviewers surfaced (NOT introduced here, out of
scope): the `readLine` goroutine lacks `safeGo`/`recoverPanic`, and
`isAppServerReadTimeout` string-matches error text (AGENTS.md rule 8) — both in
untouched code; candidates for a follow-up, not this PR.

## Size gate

`size-gated: justified overage` — +147/-90, dominated by re-indented verbatim
handshake/turn bodies + carried comments; no test added (existing validated
suite is the net). Needs human size-gate sign-off (merge requires explicit
permission per protocol §8).

Refs #499

### DIFF
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -219,7 +219,35 @@ type codexAppServerTurnStartParams struct {
 	SandboxPolicy  workflow.CodexSandboxPolicy `json:"sandboxPolicy"`
 }
 
+// run drives a full app-server session: cache the per-run config, complete the
+// SPEC §10.1 handshake, then loop turns until one stops the run (clean exit /
+// §16.5 self-stop) or the continuation cap is hit. The init → startThread →
+// per-turn split mirrors upstream start_session / run_turn.
 func (c *appServerClient) run(ctx context.Context, in RunInput, prompt string) error {
+	c.initSession(in)
+	threadID, err := c.startThread(ctx, in)
+	if err != nil {
+		return err
+	}
+	maxTurns := in.Workflow.Config.Agent.MaxTurns
+	if maxTurns <= 0 {
+		maxTurns = 20
+	}
+	for turn := 1; turn <= maxTurns; turn++ {
+		keepGoing, err := c.runSingleTurn(ctx, in, threadID, prompt, turn)
+		if err != nil {
+			return err
+		}
+		if !keepGoing {
+			return nil
+		}
+	}
+	return fmt.Errorf("codex app-server exceeded agent.max_turns=%d", maxTurns)
+}
+
+// initSession records the run's sinks and caches the per-run codex config off
+// RunInput before any protocol traffic.
+func (c *appServerClient) initSession(in RunInput) {
 	c.runtimeEventSink = in.RuntimeEventSink
 	c.phaseTransitionSink = in.PhaseTransitionSink
 	c.nextID = 1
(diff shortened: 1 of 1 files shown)


hey-yvan:
@codex review

Issue #505, opened 2026-05-29
[flaky-test] continuation/external-blocker recheck tests race the 1ms retry due_at gate


hey-yvan:
## Background

Surfaced while decomposing `(*retryFireOp).apply` for #499 (PR for that issue). Under `go test -race ./internal/orchestrator/` full-suite contention, `TestFinalize_ContinuationBudgetExhaustedEmitsStderrLine` intermittently fails (~2/49 local runs) at the `RequestDispatchAfterTrackerRecheck` call with `orchestrator: issue already claimed` (`ErrNotDispatched`).

## Root cause (pre-existing; NOT introduced by the #499 decomposition)

`dispatchOp.apply` (internal/orchestrator/actor_dispatch.go:95) rejects a tracker-rechecked dispatch when the queued continuation/external-blocker entry is **not yet due**:

```go
if !entry.IsDue(time.Now()) {
    d.result <- ErrNotDispatched
    return nil
}
```

The affected tests schedule a continuation retry with `sequenceScheduler{delays: []time.Duration{time.Millisecond}}` (DueAt = now+1ms), then wait only for `len(v.Retrying) == 1` before calling `RequestDispatchAfterTrackerRecheck` — **without** waiting for the entry to become due. When the `waitFor(Retrying==1)` poll happens to complete within ~1ms of scheduling, the recheck lands before DueAt and is correctly rejected, failing the test.

This is a test-harness timing assumption, not a production bug — the production gate is correct (a recheck must not dispatch a not-yet-due cooldown). `git diff` for the #499 PR touches neither `actor_dispatch.go` (the IsDue gate) nor `scheduleRetryOp.apply` (which sets DueAt); the mechanism is entirely in untouched code.

## Fix (deterministic barrier)

Add a due-ness predicate to the pre-recheck `waitFor`, mirroring the idiom already used at actor_test.go:869 / 906 / 1817:

```go
waitFor(t, func() bool {
    v, err := o.Snapshot(context.Background())
    return err == nil && len(v.Retrying) == 1 && !v.Retrying[0].DueAt.After(time.Now())
}, time.Second)
```

## Scope — audit ALL recheck sites (avoid half-fixing siblings)

[...]

Issue #341, opened 2026-05-24
[orchestrator] §18.1 workspace cleanup missing on the §16.5 self-stop → continuation-retry terminal path


hey-yvan:
## Summary

Follow-up surfaced while reviewing PR #339 (#331). PR #339 wired SPEC §18.1 active-transition workspace cleanup for the **reconcile-poll** path (a running/blocked issue observed terminal by the per-tick reconcile pass). A second, independent path to the same terminal-mid-run outcome — the SPEC §16.5 per-turn issue-state refresher — still leaks the workspace, even in non-routing (default) mode.

## Path

1. The §16.5 per-turn refresher (`internal/runner/codex_app_server.go`, the `IssueStateRefresher` closure built in `runtime_poller.go`) reports `!active` when the issue left the active set, so the runner stops the turn and the worker exits **cleanly** (no reconcile cancel).
2. A clean exit schedules a continuation retry.
3. When the retry fires, `retryFireAfterFetchOp.apply` (`internal/orchestrator/actor.go` ~1612, the `found == nil` branch) drops the retry and **releases the claim without cleaning the workspace** — because a terminal issue is simply absent from the active candidate set, indistinguishable there from a deleted/inactive issue.

So a mid-run terminal transition that is caught by §16.5 rather than the reconcile poll leaves the workspace on disk until the next startup sweep — the same gap #331 describes, via a different trigger.

## Upstream reference

`elixir/lib/symphony_elixir/orchestrator.ex` `handle_retry_issue_lookup` (lines ~1082-1100) fetches the issue and **branches on terminal**: terminal → `cleanup_issue_workspace(issue.identifier, …)` + release; "left active states" (non-terminal) → release only. The Go `found == nil` branch collapses both into release-only because the candidate fetch returns active issues only and cannot tell terminal from merely-absent.

## Suggested direction

On the retry-fire release path, resolve the issue's actual state (e.g. fetch terminal/inactive state the way the reconcile pass does) and, when terminal, route through the `WorkspaceCleaner` seam + `terminalWorkspaceForCleanup` helper added in #339. Reuse the same before_remove → remove → `reconcile_workspace reason=terminal` contract.

## Acceptance criteria

- A run that self-stops via §16.5 because its issue went terminal fires `before_remove` + emits `reconcile_workspace action=remove reason=terminal` once the continuation retry resolves, instead of waiting for the next startup sweep.
[...]

PR #506, opened 2026-05-29
refactor(orchestrator): decompose retryFireOp.apply into guard/wake/fetch helpers (#499)

hey-yvan:
## What & why

Decompose `(*retryFireOp).apply` (gocognit **30**, ~108 lines) — the SPEC §16.6 retry-timer-fire handler on the orchestrator actor state machine — into single-responsibility helpers. **Behavior is unchanged**; this only lowers the blast radius of future edits on a high-risk concurrency path. Part of the #499 batch (continuation of #410's complexity-hotspot decomposition; mirrors #501/#502/#503).

### Decomposition
| new symbol | role |
|---|---|
| `matchingRetryEntry` | the entry-present + `(attempt,kind)` identity guard, now **one source of truth** shared byte-for-byte by `retryFireOp` / `retryPollFailedOp` / `retryFireAfterFetchOp` (was duplicated 3×). Mirrors upstream `pop_retry_attempt_state` `:missing` (orchestrator.ex:1045-1060). |
| `(*retryFireOp).fireWakeSignal` | the continuation / external-blocker wake-only arm. |
| `(*retryFireOp).fireWithCandidateFetch` | the off-actor SPEC §16.6 candidate-fetch followup. |
| `resolveRetryTerminalState` | the `found==nil` terminal-vs-absent classification (#341), flattened from a 4-deep `if` pyramid to guard-clause early-returns. |

### gocognit (`uudashr/gocognit`)
`(*retryFireOp).apply` **30 → 4**; `retryFireAfterFetchOp.apply` → 8, `retryPollFailedOp.apply` → 6, every new helper ≤ 5. No `funlen` regression. Blocking lint gate (`staticcheck,unparam,unused,…`) = 0 issues.

## Behavior preservation (audited)
- Timer clear stays **on-actor** (in `fireWakeSignal`/`fireWithCandidateFetch`, before the followup is returned) — a later `ScheduleRetry`'s `Stop()` is still a no-op on the fired timer.
- Followup captures `id/attempt/kind/identifier` **by value** (not the op or the live entry pointer); it runs off-actor and is launched via the existing `safeGo` panic guard. The eager `identifier` capture is equivalent to the original lazy read: `resolveRetryTerminalState` runs only when `found==nil`, and `entry.Identifier` is mutated only on the `found!=nil` path — mutually exclusive per fetch.
- **Both** `context.WithTimeout(o.runCtx, retryFetchTimeout)` budgets preserved (candidate fetch + separate terminal resolve); `resolveCancel()` kept explicit to mirror the pre-refactor prompt-cancel.
- Shared `matchingRetryEntry` returns the live `*RetryEntry` map pointer, so `retryFireAfterFetchOp`'s `entry.Issue`/`entry.Identifier` mutations still reach the stored struct.

## Acceptance criteria (#499)
- [x] **Characterization tests added before the refactor**, mutation-verified non-placebo. Three new tests pin apply branches the prior suite left uncovered: entry-absent no-op (branch 1), stale **attempt**-bump drop (the attempt arm of the identity guard — previously only the kind arm was pinned), and external-blocker wake (asserts the poll wake actually fires via `retryWakeCh()`, closing the shared wake-arm placebo).
- [x] Decompose into single-responsibility helpers, BEAM guarantees preserved (followup `context.WithTimeout`, `safeGo`, on-actor timer clear), **behavior zero-change**.
- [x] `gocognit`/`funlen` finding for `retryFireOp.apply` eliminated (30→4); mutation-verified.
- [x] No new deviation, no external API change.

## Verification
```
gofmt -l $(git ls-files '*.go')        # empty
go vet ./...                            # clean
go mod tidy && git diff --exit-code     # clean
golangci-lint --enable-only=…,staticcheck,unparam,unused  # 0 issues
go test -race -covermode=atomic ./internal/orchestrator/  # ok, 84.9% cov
go build ./cmd/worker ./cmd/tui         # ok
```
- New tests: `-race -count=200` → 0 flake (deterministic; direct `.apply()`, no actor/timers).
- **Mutation (against committed artifact `4363adab01dc0aec2eabfc04c4104b4bec30c4ed`)**: break `matchingRetryEntry` not-found guard / attempt arm / `fireWakeSignal` wake emission / apply's external-blocker OR arm → each matching test **FAILS**; `git checkout` restores; tree == HEAD.

## Current state (head 4363ada)
- **CI: green** — run 26630460807, 4/4 checks pass (Go build and test, E2E Gitea mock loop, Security and supply-chain, Docker image build).
- **Review threads: 0** unresolved, non-outdated, actionable (GraphQL).
- Merge-ready **pending human size-gate sign-off** (justified overage) + explicit merge permission.

## Pre-push dual review (diff-only)
- **Reviewer A (Claude `feature-dev:code-reviewer`):** initial HIGH on eager `identifier` capture, **self-resolved to behaviorally-equivalent** under adversarial re-analysis; remaining = LOW (test `t.Fatal` on a `func()` value, rule-9). 
- **Reviewer B (Claude `feature-dev:code-reviewer`, behavior-equivalence lens):** **MERGE-READY**, all 7 checklist items equivalent, 3 tests confirmed non-placebo; one LOW (`defer` vs explicit `resolveCancel()`).
- **Codex family unavailable on BOTH paths**: local CLI reviewer (access-token expired) **and** the GitHub `@codex review` bot (account **usage limit** reached for code reviews — bot replied with the limit message, no review produced). Per protocol §4.4, compensated by the two independent Claude-family diff-only reviews above; disclosed here.

### Recorded LOW decisions (not changed)
1. `followup != nil` checks use bare `t.Fatal` — consistent with every existing `TestRetryFire_*` in the file; the message states actual-vs-expected for a binary condition and `%v` of a `func()` is a non-diagnostic address.
2. `resolveRetryTerminalState` keeps **explicit** `resolveCancel()` to exactly mirror the pre-refactor code (zero behavior change); no early-return exists between `WithTimeout` and the cancel, so no leak.

## Deferral
- **#505** — pre-existing 1ms-`due_at` race in continuation/external-blocker recheck tests (e.g. `TestFinalize_ContinuationBudgetExhaustedEmitsStderrLine`). Mechanism is in untouched code (`dispatchOp.apply` IsDue gate); **not introduced here**. Filed for a deterministic-barrier sweep across all recheck sites.

### Size gate
- [ ] `within budget`
- [x] `size-gated: justified overage` — rationale: 2 files / 395 churn LOC (added+deleted) > 300. Pure behavior-zero-change decomposition + the acceptance-criterion-mandated, mutation-verified characterization tests, which cannot be split from the refactor without losing the safety atomicity the issue requires (same bundling as #501/#502/#503). No scope creep; needs human size-gate sign-off before merge.

Refs #499.

### DIFF
--- a/internal/orchestrator/actor_retry.go
+++ b/internal/orchestrator/actor_retry.go
@@ -260,114 +260,158 @@ type retryFireOp struct {
 	kind    RetryKind
 }
 
-func (r *retryFireOp) apply(st *OrchestratorState) func() {
-	entry, ok := st.RetryAttempts[r.id]
+// matchingRetryEntry returns the queued RetryEntry for id only while it still
+// matches this fire's (attempt, kind) identity, and false otherwise. The two
+// false cases both mean "drop this fire as a no-op": (1) the entry is absent —
+// already consumed by reconciliation's ReleaseClaim or by an earlier fire of
+// the same retry; and (2) a newer ScheduleRetry replaced the entry with a
+// different attempt/kind, so an older timer fired late (the Stop()-missed race,
+// SPEC §16.6). It mirrors upstream pop_retry_attempt_state's :missing result
+// (orchestrator.ex:1045-1060) minus the pop — the Go port keeps the entry in
+// RetryAttempts across the async candidate fetch and re-checks this guard at
+// every actor re-entry (retryFireOp, retryPollFailedOp, retryFireAfterFetchOp).
+// The returned *RetryEntry is the live pointer from the map, so callers'
+// in-place mutations (entry.Timer, entry.Issue, entry.Identifier) still reach
+// the stored struct.
+func matchingRetryEntry(st *OrchestratorState, id IssueID, attempt int, kind RetryKind) (*RetryEntry, bool) {
+	entry, ok := st.RetryAttempts[id]
 	if !ok {
-		// Already consumed by reconciliation (ReleaseClaim) or by an
-		// earlier fire of the same retry. Either is correct.
-		return nil
+		return nil, false
 	}
-	if entry.Attempt != r.attempt || entry.Kind != r.kind {
-		// A newer ScheduleRetry replaced this entry; the older timer
-		// fired late. Drop the stale fire — the newer entry will
-		// re-dispatch on its own timer.
+	if entry.Attempt != attempt || entry.Kind != kind {
+		return nil, false
+	}
+	return entry, true
+}
+
+func (r *retryFireOp) apply(st *OrchestratorState) func() {
+	entry, ok := matchingRetryEntry(st, r.id, r.attempt, r.kind)
+	if !ok {
 		return nil
 	}
 	if entry.Kind == RetryKindContinuation || entry.Kind == RetryKindExternalBlocker {
-		// Continuation and external-blocker cooldown retries are wake-up
-		// signals. They must not spawn from the cached issue snapshot or carry
-		// failure retry accounting: a poll has to observe the issue still
-		// active and call RequestDispatchAfterTrackerRecheck, which consumes
-		// the entry before spawning the next normal turn.
-		entry.Timer = nil
-		o := r.o
-		return func() { o.wakeRetryPollLoop() }
+		return r.fireWakeSignal(entry)
 	}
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

PR #511, opened 2026-05-29
refactor(runner): decompose handleDynamicToolCall into per-concern helpers (#499)

hey-yvan:
## What

Decompose `(*appServerClient).handleDynamicToolCall` (gocognit **20**, the highest remaining #499-scoped hotspot) into three single-concern helpers, **zero behavior change**:

- `resolveDynamicToolResult` — tool lookup + invoke, returning the wire result string. Name not advertised → records the SPEC §10.4 `unsupported_tool_call` event + returns the structured "unsupported dynamic tool" failure; argument-decode and execution failures → structured `dynamicToolFailure`.
- `withMutationAuditSink` — derives the tool-execution context carrying the Linear GraphQL mutation audit sink (aiops-platform extension; no upstream equivalent).
- `replyDynamicToolOutput` — parse the result and reply as a JSON-RPC result (`id` present) or `item/tool/call/output` notification (`call_id`).

Boundaries mirror upstream `elixir/.../codex/app_server.ex` `maybe_handle_approval_request("item/tool/call", …)`: `tool_call_name`/`tool_call_arguments` (already extracted) → tool execution + `normalize_dynamic_tool_result` → send/emit.

## Acceptance criteria (#499)

- [x] **Characterization tests committed first** (`8053f21`), before any decomposition. The end-to-end subprocess suite only asserts the unsupported-tool stdin reply; these pin the runtime-event, undecodable-args, tool-error, id-vs-notification, and mutation-sink branches at the function boundary. `go test -count=1 -race` clean twice (no flake).
- [x] Function purified into single-concern helpers; BEAM guarantees N/A (no new goroutine/timer/followup; `ctx` flows to `tool.Call` unchanged).
- [x] gocognit **20 → finding eliminated** (`handleDynamicToolCall` + all 3 helpers <10); funlen clean. Mutation-verified non-placebo.
- [x] No new deviation; no external API change.

## Behavior preservation note

The only removed work: the original computed `dynamicToolResult(false, "unsupported…")` **unconditionally** (and checked its error) even on the tool-found path, then discarded it. That marshal is a fixed struct of string/bool/slice scalars and cannot fail, so the checked error was dead code on every reachable path. Removing it on the found path is unobservable. (Confirmed by both pre-push reviewers.)

## Verification

```
gofmt -l … (empty) · go vet ./... · go mod tidy (clean) · go build ./cmd/worker ./cmd/tui
go test -race -covermode=atomic ./...   # all green except the known
[...]

PR #513, opened 2026-05-29
refactor(runner): dedupe and decompose reason-field extractors (#499)

hey-yvan:
## What

Dedupe **and** decompose the two reason-field extractors in `codex_app_server_turn.go`, eliminating two #499 hotspots at once with **zero behavior change**:

- `extractReasonFields` (gocognit **15**) and `copyAllowlistedReasonFields` (gocognit **18**) duplicated the same two key allow-lists and the same nested-`error` walk.
- Hoist the allow-lists into shared `reasonFieldKeys` / `nestedReasonFieldKeys` package vars — **one source of truth**, documented as an immutable redaction boundary.
- Extract the per-shape nested handlers: `firstNestedReasonString` (extract path: first non-empty trimmed nested scalar) and `scrubbedNestedReason` (copy path: fresh allow-listed sub-map).
- Both top-level loops now scan `reasonFieldKeys` and delegate the map case; the `raw, ok := m[key]; if !ok { continue }` guard collapses into the type switch (an absent / non-string/map value matches no case exactly as before).

**Why both functions in one PR:** they share the exact allow-list. Decomposing only one would leave the key lists duplicated, violating the AGENTS.md *one source of truth per concept* rule — so the coupled pair is atomic.

These are redaction-discipline guards: only an explicit allow-list reaches returned error strings or the JSON event surface (`/api/v1/state`, runtime log), so non-allow-listed protocol fields (which may carry tool output, elicitation snippets, or secrets) are dropped.

## Acceptance criteria (#499)

- [x] **Characterization committed first** (`351d52f`): direct table tests pinning top-level + nested precedence, the `error` string-vs-object dual shape, top-level **and** nested trimming, empty-skip, copy-keeps-all vs. extract-returns-first, and nested-scrubbing of non-allow-listed siblings. Plus `TestReasonFieldAllowlistsAreStable` pinning allow-list membership (see Security below). `go test -count=1 -race` clean twice.
- [x] Purified into single-concern helpers; BEAM guarantees N/A (pure functions).
- [x] gocognit **15 + 18 → findings eliminated** (all four functions <10); funlen clean.

## Security hardening (review-driven)

The adversarial review panel flagged that the shared allow-lists are **mutable package-level slices** that bound the redaction surface — a future `append`/reassign would silently widen what reaches error strings/events with no compile-time signal. Addressed by: an explicit immutability comment on the var block, and `TestReasonFieldAllowlistsAreStable`, which pins exact membership/order so an accidental widening fails CI (verified: appending a key fails the guard).

## Verification

```
gofmt -l (empty) · go vet · go mod tidy (clean) · go build ./cmd/worker ./cmd/tui
go test -race ./internal/runner/   # green except the known
                                   # TestCodexAppServerInitializeTimeoutDiagnostics
                                   # local-sandbox-only failure (passes in CI)
```

**Mutation verification** (isolated, against the committed decomposed artifact): dropping `error_code` from the shared list (fails **both** extract and copy subtests, confirming the dedup is covered on both sides), reordering the nested keys, removing top-level/nested trimming, **storing untrimmed nested values**, and copying the whole error object instead of the scrubbed submap each fail the matching subtests. ✓

## Size-gate

`within budget` — +139/-23 over two files (production 42/23, tests 97/0), under the 300-LOC default.

## Review

Pre-push 3-lens adversarial panel (behavior-equivalence / conventions+Go-correctness / test-completeness+scope): behavior-equivalence **PASS** (the type-switch collapse verified identical for absent / nil / wrong-type keys), conventions **APPROVE** (dedup confirmed complete via package grep), test-completeness **APPROVE** after the nested-trim MEDIUM gap and the mutable-allow-list LOW were both fixed in `351d52f`. `@codex review` triggered on the head; if usage-limited (as in the prior #499 sessions), the panel + green CI is the compensating control per checklist item 4 / protocol §4.

Refs #499

### DIFF
--- a/internal/runner/codex_app_server_reason_test.go
+++ b/internal/runner/codex_app_server_reason_test.go
@@ -1,0 +1,97 @@
+package runner
+
+// codex_app_server_reason_test.go pins the allow-listed reason-field extractors
+// — extractReasonFields and copyAllowlistedReasonFields — at their own function
+// boundaries before #499 decomposes them (gocognit 15 and 18) and deduplicates
+// their shared allow-list. Both are redaction-discipline guards: only an
+// explicit key allow-list reaches returned error strings or the JSON event
+// surface, so the field-precedence, string-vs-object dual shape, trimming, and
+// nested-scrubbing behavior asserted here must hold identically across the
+// refactor.
+//
+// The end-to-end TestCodexAppServerRunnerRedactsTurnFailureParams remains the
+// authority for the runner-level redaction path; these isolate the precedence
+// and dual-shape branches it only samples.
+
+import (
+	"reflect"
+	"testing"
+)
+
+// TestReasonFieldAllowlistsAreStable pins the membership and order of the shared
+// reason-field allow-lists. They are a security boundary — they bound what
+// reaches returned error strings and the JSON event surface — and are mutable
+// package-level slices, so an accidental append/reorder would silently widen the
+// redaction surface without failing any behavioral test that doesn't happen to
+// use the newly admitted key. This guard makes any change deliberate.
+func TestReasonFieldAllowlistsAreStable(t *testing.T) {
+	if want := []string{"reason", "error", "message", "error_code"}; !reflect.DeepEqual(reasonFieldKeys, want) {
+		t.Errorf("reasonFieldKeys = %#v; want %#v (widening the allow-list expands the redaction surface)", reasonFieldKeys, want)
+	}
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

hey-yvan:
**Codex review unavailable** — usage-limit message again (as #511/#512). Compensating control per checklist item 4 / protocol §4: the pre-push 3-lens adversarial panel (behavior-equivalence PASS / conventions APPROVE / test-completeness APPROVE after fixing the nested-trim MEDIUM + the mutable-allow-list LOW). CI green (4/4), 0 actionable review threads. Will re-trigger `@codex review` if the limit resets before merge.

PR #516, opened 2026-05-29
refactor(orchestrator): decompose reconcileInactiveTrackerIssuesOp.apply (#499)

hey-yvan:
## What

Decompose `reconcileInactiveTrackerIssuesOp.apply` (gocognit **19**, the highest remaining #499-scoped orchestrator actor hotspot) into three per-collection helpers, **zero behavior change**:

- `reconcileInactiveRun` — release + refresh + terminal cleanup-flag + cancel a running entry; returns nil for an unlisted issue.
- `reconcileInactiveRetry` — terminal retry → workspace cleanup (carrying the queued continuation) + release; else release only.
- `reconcileInactiveBlocked` — terminal block → workspace cleanup + release; else release only.

`apply` now just loops and collects. Boundaries mirror upstream `orchestrator.ex` per-entry functions (`reconcile_issue_state` / `handle_retry_issue_lookup` / `reconcile_blocked_issue_state`) and the sibling active op's existing helper structure.

**Translation care (actor state machine):** the `if !ok { continue }` partial-listing guard becomes each helper's early return *before* the deferred release is registered; the retry/blocked unconditional `st.ReleaseClaim(id)` moves to `defer st.ReleaseClaim(id)`, which runs after the cleanup is built (same order, same release on every listed path). The followup (`reconcileCancelFollowup`) is unchanged, so the BEAM panic-recovery/goroutine discipline is inherited.

## Acceptance criteria (#499)

- [x] **Characterization committed first** (`578068d`), pinning the apply's state changes + followup per the actor acceptance. A mutation audit + adversarial review found **two release branches the existing suite did not pin** — the retry/blocked **partial-listing** guards (an issue absent from a pagination-truncated listing must be left untouched) and the **blocked non-terminal listed release** (release without cleanup). Both are now pinned and mutation-verified.
- [x] BEAM guarantees preserved (followup unchanged; no new goroutine/timer).
- [x] gocognit **19 → finding eliminated** (all four functions <10); funlen clean.

## Verification

```
gofmt -l (empty) · go vet · go mod tidy (clean) · go build ./cmd/worker ./cmd/tui
go test -race ./internal/orchestrator/   # full suite green
```

**Mutation verification** (against the committed decomposed artifact): never-flag-terminal-cleanup on running (8 fails), invert retry/blocked terminal gate (3 / 1), drop the run/retry/blocked `!ok` guard (caught by existing `TestPollOnce*` + the new tests), and gating the blocked release behind the terminal check (caught by the new non-terminal-release test) each fail their subtests. ✓

## Size-gate

`within budget` — +250/-57 over two files (production 85/57, tests 165/0), under the 300-LOC default.

## Review

Pre-push 3-lens adversarial panel (state-machine equivalence / conventions+BEAM+correctness / followup+state-change coverage): equivalence **PASS** (all four translation risks traced + 20× -race stress, no counterexample — the `defer`-vs-explicit release reorder is observationally invisible because `ReleaseClaim` mutates only state maps, never a field the cleanup builder reads), conventions **APPROVE**, coverage **APPROVE** after the one MEDIUM gap (blocked non-terminal release) was closed in `578068d`. `@codex review` triggered on the head; if usage-limited (as in the prior #499 sessions), the panel + green CI is the compensating control per checklist item 4 / protocol §4.

Refs #499

### DIFF
--- a/internal/orchestrator/actor_reconcile.go
+++ b/internal/orchestrator/actor_reconcile.go
@@ -406,76 +406,104 @@ type reconcileInactiveTrackerIssuesOp struct {
 	result         chan<- []*RunningEntry
 }
 
+// apply releases every in-process entry whose issue is observed in the inactive
+// listing, gating SPEC §18.1 workspace cleanup on a terminal transition. Each
+// per-collection helper returns what to collect: running entries are cancelled
+// (their workspace cleanup deferred to finalize), while blocked (input-required)
+// and retry-queued (a clean §16.5 self-stop finalized the run) entries hold no
+// live worker, so a terminal one's workspace is removed now via the shared
+// WorkspaceCleaner. An entry whose issue is absent from the (possibly partial)
+// listing is left untouched.
 func (r *reconcileInactiveTrackerIssuesOp) apply(st *OrchestratorState) func() {
 	var cancelEntries []*RunningEntry
-	// terminalCleanups collects terminal-state entries that hold no live worker
-	// — blocked (input-required, stopped executing) and retry-queued (a clean
-	// SPEC §16.5 self-stop finalized the run before scheduling the continuation
-	// retry) — whose workspace must be removed now rather than deferred to the
-	// finalize path the way running entries are. Every path goes through the
-	// same WorkspaceCleaner so before_remove fires and a reconcile_workspace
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

PR #520, opened 2026-05-29
refactor(orchestrator): decompose refreshActive + reconcileStalled ops (#499)

hey-yvan:
## What

Close out the **last two** #499-scoped orchestrator hotspots (both gocognit **14**), **zero behavior change**:

- **`refreshActiveTrackerIssuesOp.apply`** — the per-tick refresh-only pass (no cancellation). Extract `activeMatch` (the shared listed-active-on-same-route gate that also refreshes the `ClaimedIssues` snapshot) so each loop just stores the fresh issue on its entry, and `doneFollowup` (the no-op completion signal). A kept entry is refreshed but **not** released; a miss (absent / non-active / route-changed) leaves the entry untouched.
- **`reconcileStalledRunsOp.apply`** — SPEC §8.5 Part A stall detection. Extract `isStalled` (the `LastEventAt`→`StartedAt`→not-stalled timing reference, stalled iff `now.Sub(ref) > timeout`) and `cancelStalledFollowup`.

`apply` mutates `st` synchronously on the actor; the followups run off-actor — the mutation discipline is preserved. No goroutine/timer added.

## Acceptance criteria (#499)

- [x] **Characterization committed first** (`b8414e0`), via direct apply-on-state tests pinning the branches the integration suite missed: the per-collection refresh actions, the route-changed **and** absent-from-listing misses, and the stall-timing reference including the **exact-budget boundary pair** (elapsed == budget is not stalled; one tick over is — pinning strict `>`).
- [x] gocognit **14 + 14 → findings eliminated** (all functions <10); funlen clean.

## Verification

```
gofmt -l (empty) · go vet · go build ./cmd/worker ./cmd/tui
go test -race ./internal/orchestrator/   # full suite green
```

**Mutation verification** (committed artifact): dropping the `activeMatch` route gate, the running/retry refresh actions, the StartedAt fallback, the zero-reference skip, and flipping the stall comparison `>` → `>=` (caught by the new exact-budget row) each fail their subtests. ✓

## Size-gate

`within budget` — +159/-26 over two files (production 50/26, tests 109/0).

## Review

Pre-PR 2-lens adversarial panel (equivalence / conventions+test-quality): equivalence **PASS** (no divergence; the `ClaimedIssues`-before-refresh order swap and the `>` ≡ `!(<=)` complement both confirmed). Two findings — a MEDIUM exact-stall-boundary gap and a LOW absent-miss-not-self-contained — both addressed in `b8414e0` (boundary pair + the R3 absent-from-listing case). `@codex review` triggered on the head.

**This is the final #499 hotspot.** With it merged, every named function in the issue's scope (the actor `apply` ops + the codex protocol path) is decomposed below the gocognit threshold; I'll confirm with a full `golangci-lint --enable gocognit` sweep.

Refs #499

### DIFF
--- a/internal/orchestrator/actor_reconcile.go
+++ b/internal/orchestrator/actor_reconcile.go
@@ -195,35 +195,47 @@ type refreshActiveTrackerIssuesOp struct {
 
 func (r *refreshActiveTrackerIssuesOp) apply(st *OrchestratorState) func() {
 	for id, run := range st.Running {
-		issue, ok := r.issuesByID[string(id)]
-		if !ok || !isActiveTrackerState(issue.State, r.activeStates) || !sameServiceRoute(run.Issue, issue) {
-			continue
+		if issue, ok := r.activeMatch(st, id, run.Issue); ok {
+			refreshRunningIssue(run, issue)
 		}
-		refreshRunningIssue(run, issue)
-		st.ClaimedIssues[id] = issue
 	}
 	for id, retry := range st.RetryAttempts {
-		issue, ok := r.issuesByID[string(id)]
-		if !ok || !isActiveTrackerState(issue.State, r.activeStates) || !sameServiceRoute(retry.Issue, issue) {
-			continue
+		if issue, ok := r.activeMatch(st, id, retry.Issue); ok {
+			retry.Issue = issue
 		}
-		retry.Issue = issue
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

Issue #521, opened 2026-05-30
[Refactor] Decompose the 10 heaviest remaining gocognit hotspots (≥30)


hey-yvan:
## Goal

Decompose the **10 heaviest remaining `gocognit` hotspots** (score ≥30) — the high-blast-radius functions where #410/#499's "safe decomposition" rationale clearly applies. The other ~48 mild (11–16) and ~22 moderate (17–29) findings are deliberately **not** in scope: forcing cohesive 11–16 functions under the threshold is largely cosmetic and risks over-abstraction. New-code regressions are handled separately by #504 (new-from-rev gate), so this is opportunistic existing-code cleanup, not a precondition for anything.

## Targets (gocognit ≥30, descending)

- [x] `policy.globMatch` — 54 (`internal/policy/policy.go`)
- [x] `worker.ReconcileStartup` — 42 (`internal/worker/reconcile.go`)
- [x] `workspace.(*Manager).PrepareGitWorkspace` — 41 (`internal/workspace/manager.go`)
- [x] `tracker.(*LinearClient).ListIssuesByStates` — 37 (`internal/tracker/linear.go`)
- [x] `gitea.(*TrackerClient).listIssuesByStateLabel` — 35 (`internal/gitea/tracker_client.go`)
- [x] `worker.listIssueWorkspaces` — 34 (`internal/worker/reconcile.go`)
- [x] `runner.(MockRunner).Run` — 34 (`internal/runner/mock.go`)
- [x] `orchestrator.(*Poller).reconcileTick` — 33 (`internal/orchestrator/poller.go`)
- [x] `tracker.(*GitHubClient).listIssuesForState` — 31 (`internal/tracker/github.go`)
- [x] `runner.(giteaIssueLabelsProxy).call` — 30 (`internal/runner/gitea_tools.go`)

## Per-PR recipe (same as #499)

1. **Characterization-first**: mutation-audit the existing suite for the target; add direct/characterization tests for the gaps (mutation-verified non-placebo) and commit them **before** decomposing. Several of these touch IO/network/subprocess (tracker pagination, git workspace prep, sandbox) — pin behavior with mocked transports first.
2. Decompose into single-concern helpers, **zero behavior change**; preserve BEAM guarantees (followup `context.WithTimeout`, `safeGo`/`recoverPanic`, timer `Stop()`).
3. Character scanners / parsers (e.g. `globMatch`) get a throwaway differential fuzz vs a frozen copy of the original.
[...]

PR #523, opened 2026-05-30
refactor(policy): decompose globMatch into wildcard helpers (#521)

hey-yvan:
## What & why

First PR of #521 (decompose the 10 heaviest `gocognit` hotspots, one function per PR, descending). Target: **`policy.globMatch`** — gocognit **54**, the heaviest baseline function. The repo's blocking gate (`.golangci.yml` `gocognit.min-complexity: 10`) fails any non-test function over 10.

`globMatch` is now a thin dispatcher; each matching concern is a single-purpose helper:

| function | role | gocognit |
|---|---|---|
| `globMatch` | dispatch loop | **6** |
| `matchDoubleStar` | `**` trailing / non-`/` suffix scan | 5 |
| `matchDoubleStarSegment` | `**/<inner>` segment skipping | 5 |
| `matchSingleStar` | `*` run that stops at `/` | 6 |
| `matchSingleChar` | one literal byte or `?` vs `path[0]` | 2 |

`globMatch` drops **54 → 6**; every helper ≤ 6. The `//nolint:gocognit // baseline (#521)` directive is removed (AGENTS.md rule 7).

## Zero behavior change — how it's verified

- The `**`/`*` switch cases already returned unconditionally → turning them into `return helper(...)` is exact.
- The original separate `?` case and literal `default` fold into one `default` calling `matchSingleChar` (same verdict: `?` never matches `/`; a literal must equal `path[0]`, so a literal `/` only matches `/`).
- **Characterization test committed first** (`TestGlobMatchBranches`, calls `globMatch` directly), pinning every branch boundary; **mutation-verified non-placebo** against the committed artifact (broke `matchSingleChar`'s `?`/`/` guard, `matchDoubleStarSegment`'s boundary, and `matchSingleStar`'s slash-stop — each caught).
- **Throwaway differential fuzz** vs a frozen copy of the pre-refactor matcher: **0 divergence** (~12.5M execs locally; the independent adversarial reviewer re-ran its own frozen-copy differential, ~50M execs cumulative, also 0). Bounded input size + `*`-count to avoid the matcher's pre-existing exponential `**`-stack blowup (identical in both copies; not introduced here).

## Verification

```
gofmt -l            # clean
go vet ./internal/policy/...
go test -race -covermode=atomic ./internal/policy/...   # ok, 94.2% cov
go run .../golangci-lint@v2.12.2 run --config=.golangci.yml \
  --new-from-rev=origin/main --enable-only=gocognit,funlen ./internal/policy/...  # 0 issues
```

### Size gate
- [x] `within budget` — 2 files, well under 12 files / 300 LOC.

## Out of scope
`Evaluate` (gocognit 14) keeps its `//nolint baseline (#521)` — not in this issue's curated 10. The other 9 targets follow as separate serial PRs.

A pre-existing latent issue surfaced while characterizing `matchSingleChar` — `?` matches **byte-wise**, not rune-wise — is deferred to **#524** (behavior change, out of scope for this decomposition).

Refs #521

---
### Review & gate status (head 6222a1c)
- **CI:** 7/7 green (lint, test, build, **golangci blocking gate**, e2e, security, docker).
- **Pre-push review:** two independent Claude adversarial reviews, both **MERGE-READY** (one ran its own differential fuzz + the gate). Reviewer-1's only LOW — a literal-`/` mismatch case — is addressed (`{"a/b","axb",false}`).
- **`@codex` review:** the GitHub `@codex` bot hit **Codex usage limits** and did not run. Per the documented usage-limit fallback (independent dual review + green CI + disclosure), coverage rests on the two Claude adversarial reviews above plus green CI. 0 unresolved review threads.

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -155,60 +155,20 @@ func Match(pattern, path string) bool {
 	return globMatch(pattern, path)
 }
 
-// globMatch is the recursive glob matcher. It walks both pattern and path
-// with explicit handling for "**" so that double-star spans path separators
-// while single "*" and "?" do not.
-func globMatch(pattern, path string) bool { //nolint:gocognit // baseline (#521)
+// globMatch is the recursive glob matcher. It is a thin dispatcher: the
+// separator-spanning "**" and single-segment "*" wildcards delegate to
+// helpers, and every other position (a literal byte or "?") consumes exactly
+// one character via matchSingleChar.
+func globMatch(pattern, path string) bool {
 	for len(pattern) > 0 {
 		switch {
 		case strings.HasPrefix(pattern, "**"):
-			rest := strings.TrimPrefix(pattern, "**")
-			// Allow leading "/" after "**" to consume an arbitrary number of
-			// path segments, including zero.
-			if strings.HasPrefix(rest, "/") {
-				inner := rest[1:]
-				// "**/" matches zero segments.
-				if globMatch(inner, path) {
-					return true
-				}
-				for i := 0; i < len(path); i++ {
-					if path[i] == '/' && globMatch(inner, path[i+1:]) {
-						return true
-					}
-				}
-				return false
-			}
-			// Trailing "**" matches the rest of path.
-			if rest == "" {
-				return true
-			}
-			// "**" followed by non-slash: try every suffix.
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

PR #527, opened 2026-05-30
fix(policy): match `?` glob wildcard rune-wise, not byte-wise

### LINKED ISSUE #524: policy: `?` glob wildcard matches byte-wise, not rune-wise (multi-byte UTF-8 paths)
## Summary

`policy.matchSingleChar` (and, before #523, the inline `?` branch of `globMatch`) matches `?` **byte-wise**, not rune-wise:

```go
// internal/policy/policy.go
func matchSingleChar(pc byte, path string) bool {
	if len(path) == 0 {
		return false
	}
	if pc == '?' {
		return path[0] != '/'   // consumes exactly ONE byte
	}
	return pc == path[0]
}
```

A `?` consumes a single byte, so against a multi-byte UTF-8 path character it matches only the first byte. Because the whole-pattern match is anchored at both ends, a lone `?` will *fail* to match a multi-byte char rather than match it — so `secret?.key` will not match `secreté.key`, and `?`-based **deny** rules behave inconsistently between ASCII and non-ASCII paths.

## Why it matters

`policy.Evaluate` is the pre-push deny/allow gate. Byte-wise `?` semantics make `?` deny patterns unreliable for non-ASCII filenames, which is surprising for a security-relevant matcher.

## Scope / provenance

- **Pre-existing** behavior; not introduced by #523 — the old inline `?` branch did the same `path[0]` byte compare. Surfaced while characterizing `matchSingleChar` during the #523 decomposition review, and deliberately scoped out of that zero-behavior-change PR.
- Fixing it is a deliberate **behavior change** (rune-wise `?`), so it belongs in its own PR with its own characterization tests, not folded into the #521 decomposition.

## Acceptance criteria

- [ ] Decide intended `?` semantics (almost certainly: `?` matches exactly one **rune** except `/`), document it in `Match`'s doc comment.
- [ ] Make `globMatch` advance by a rune for `?` (and confirm `*`/`**` UTF-8 handling is consistent).
- [ ] Add tests for multi-byte paths (`?` vs `é`, `*` not crossing `/` with multi-byte runs, deny-rule bypass regression).

Refs #521, PR #523.

hey-yvan:
## Summary

`policy.Match`'s `?` glob wildcard consumed exactly **one byte**, not one rune. Against a multi-byte UTF-8 path character it matched only the first byte; because the whole-pattern match is anchored at both ends, a lone `?` *failed* to match a multi-byte rune rather than matching it. So `secret?.key` did not match `secreté.key`, and `?`-based **deny** rules behaved inconsistently between ASCII and non-ASCII paths in the pre-push deny/allow gate (`policy.Evaluate`) — a surprising gap in a security-relevant matcher.

Closes #524.

## Change

- `globMatch` now advances `path` by a full rune for `?` via `utf8.DecodeRuneInString` (new `matchSingleRune` helper), rejecting empty paths and `/`. Invalid UTF-8 decodes to a single-byte `RuneError`, so `?` still advances and never stalls.
- Literal and `*`/`**` matching stay byte-wise: UTF-8 is self-synchronizing, a rune matches iff all its bytes match, and `/` (0x2F) never appears inside a multi-byte rune — so those paths are already rune-correct and never split a rune or cross a separator. Confirmed by new multi-byte `*`/`**` test cases.
- Documented the rune semantics in `Match`'s doc comment (`"?" matches exactly one rune (Unicode code point) except "/"`).
- Removed the now-obsolete byte-oriented `matchSingleChar` helper.

This is a deliberate **behavior change** (rune-wise `?`), scoped to its own PR with characterization tests as the issue requested — not folded into the #521/#523 zero-behavior-change decomposition.

## Acceptance criteria

- [x] Decide & document intended `?` semantics (one rune except `/`) in `Match`'s doc comment.
- [x] Make `globMatch` advance by a rune for `?`; confirmed `*`/`**` UTF-8 handling is consistent (byte-wise but rune-correct, with tests).
- [x] Tests for multi-byte paths: `?` vs `é`, `*` not crossing `/` with multi-byte runs, and a deny-rule bypass regression at the `Evaluate` gate.

## Verification

```
gofmt -l internal/policy/*.go        # empty
go mod tidy && git diff --exit-code  # clean
go vet ./internal/policy/
golangci-lint run --config=.golangci.yml internal/policy/...   # 0 issues
go test -race -covermode=atomic ./...  # all pass (policy 94.6% cov)
```

**Mutation verification** (against the committed artifact): reverting `matchSingleRune` to byte-wise (`return 1`) fails the new tests at all three levels — `TestMatch` (`secret?.key` vs `secreté.key`), `TestGlobMatchBranches` (`?` vs `é`, `??` vs `é`), and `TestEvaluateDenyQuestionMarkMultiByte` (deny gate bypass) — and they pass with the fix.

## Size

`within budget` — small, atomic single-package change.

## Risk / SPEC alignment

The glob matcher is an aiops-platform policy extension with no Symphony/Elixir analog to port, so this is a local correctness fix with no SPEC-alignment or DEVIATIONS.md impact. No new deviation introduced.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W6GnTP5Q2dnEPhZFzk2qPL)_

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -6,6 +6,7 @@ package policy
 import (
 	"fmt"
 	"strings"
+	"unicode/utf8"
 )
 
 // Config mirrors the relevant fields of workflow.PolicyConfig. It is
@@ -134,7 +135,7 @@ func matchAny(patterns []string, path string) (string, bool) {
 
 // Match reports whether path matches the glob pattern. Supported syntax:
 //
-//   - "?"  matches any single character except "/"
+//   - "?"  matches exactly one rune (Unicode code point) except "/"
 //   - "*"  matches any sequence of characters except "/"
 //   - "**" matches any sequence of characters, including "/"
 //   - "/**" as a trailing component matches the empty string or "/<anything>"
@@ -157,8 +158,8 @@ func Match(pattern, path string) bool {
 
 // globMatch is the recursive glob matcher. It is a thin dispatcher: the
 // separator-spanning "**" and single-segment "*" wildcards delegate to
-// helpers, and every other position (a literal byte or "?") consumes exactly
-// one character via matchSingleChar.
+// helpers, "?" consumes exactly one rune via matchSingleRune, and a literal
+// position consumes one byte.
 func globMatch(pattern, path string) bool {
 	for len(pattern) > 0 {
 		switch {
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

---

PR #525, opened 2026-05-30
refactor(worker): decompose ReconcileStartup into single-concern helpers (#521)

hey-yvan:
## What & why

Second PR of #521 (decompose the 10 heaviest `gocognit` hotspots, one per PR, descending). Target: **`worker.ReconcileStartup`** — gocognit **42** + funlen-flagged (the 2nd heaviest baseline function). The blocking gate (`.golangci.yml` `gocognit.min-complexity: 10`) fails any non-test function over 10.

`ReconcileStartup` becomes a thin orchestrator; each concern is a single-purpose helper:

| helper | role |
|---|---|
| `validateReconcileConfig` | required-input guards |
| `fetchReconcileIssues` | active/terminal fetch + SPEC §8.6 skip/partial outcomes |
| `newReconcileIndex` | active/terminal key maps + `canRemoveUnknown` |
| `reconcileWorkspaces` | per-workspace loop + removed/kept tally + `ctx.Err()` check |
| `reconcileWorkspace` | keep/remove decision for one workspace |
| `reconcileEndPayload` | `reconcile_end` payload assembly |

`ReconcileStartup` drops **42 → 8**; every helper ≤ 9. The `//nolint:gocognit,funlen` baseline directive is removed.

## Zero behavior change — how it's verified

Preserved exactly: active-then-terminal fetch **order**; the skip-on-active-fail path (`reconcile_end{status:skipped}` + return nil); the terminal-fail partial path (`terminalIssues=nil`, `terminalFetchOK=false`, `status:partial`); `canRemoveUnknown` derivation; the `activeKeysForIssue` nil-default; the per-iteration `ctx.Err()` check; the mass-delete guard; the per-workspace classification; and — the subtlest invariant — **a declined removal (`removeWorkspace` returns false) counts as NEITHER kept nor removed** (`reconcileWorkspace` returns `(false,false)`).

- The existing reconcile suite (24 tests: terminal/unknown removal, skipped/partial status, rework keep, mass-delete guard, layouts, hooks, validation) passes unchanged.
- **New characterization** pins the `reconcile_end` kept/removed counts (ok path, terminal-fetch partial, and the `active_rework` prefix branch) that existing tests didn't assert — **mutation-verified non-placebo** (status flip, kept/removed swap, and active_rework `keptOne` flip each caught).

## Verification

```
go vet ./internal/worker/...
go test -race -covermode=atomic ./internal/worker/...   # ok (~25s)
go build ./cmd/worker ./cmd/tui
[...]

PR #533, opened 2026-05-30
refactor(tracker): decompose ListIssuesByStates into single-concern helpers (#521)

hey-yvan:
## What

Pure extract-function decomposition of `(*LinearClient).ListIssuesByStates` (`internal/tracker/linear.go`) for the #521 complexity-baseline cleanup. The function becomes a thin orchestrator delegating to single-concern helpers, matching the merged #523/#525/#526 style. **Zero behavior change** — identical output, error wrapping/ordering/classification, side effects and ordering, and request shape for identical input.

## Helpers extracted (single concern each)

- **`listLinearIssuesQuery`** (package const) — hoisted the fixed ListIssues GraphQL query out of the function body. Byte-identical to the original literal (verified by diff); the #326 "no customFieldValues" explanatory comment block stays attached.
- **`linearIssueNode`** / **`linearIssuesResponse`** (named types) — name the previously-inline anonymous response structs. Every `json` tag copied byte-for-byte.
- **`requireListIssuesConfig`** — the two precondition guards (API key checked before project slug, same sentinels `CategoryMissingTrackerAPIKey` / `CategoryMissingTrackerProjectSlug`, same order).
- **`normalizeRequestedStates`** — trim each requested state, drop empties (cap `len(states)`). The SPEC §17.3 `if len(nonEmpty)==0 { return nil, nil }` short-circuit stays in the caller.
- **`fetchLinearIssuesPage`** — one page request (`c.graphql`) + the `out.Errors -> linearGraphQLErrors` guard + node mapping, returning mapped issues plus a typed `linearPageInfo`. The page-control branches (HasNextPage / EndCursor / cursor advance) and the terminal plain `fmt.Errorf` stay visible in the parent loop.
- **`mapLinearIssueNode`** — maps one node: `parseLinearIssueTime("createdAt")` then `parseLinearIssueTime("updatedAt")` (first malformed field wins), `isTodoState`-gated `c.linearBlockersForIssue` (blockers fetched only for Todo state, error returned verbatim), label normalization into a fresh `make([]string,0,len(n.Labels.Nodes))` with trim+ToLower and skip-empty, and the `Issue` literal (`CustomFields` stays nil). Same `ctx` threaded into `graphql` / `linearBlockersForIssue`.

> Deviation from the literal plan: instead of separate `mapLinearIssueNodes` + `nextLinearIssuePage` helpers, the final `fetchLinearIssuesPage` returns mapped issues + a typed `linearPageInfo` and the parent keeps the three page-control branches inline. The blocking CI gate is golangci-lint `gocognit` with `min-complexity: 10` (stricter than the `-over 29` verification helper); the merged #525/#526 orchestrators all land under 10. This factoring reaches 9 while keeping page-control fully visible in the parent, as the plan required.

## Complexity (before/after)

| function | gocognit before | gocognit after |
|---|---|---|
| `(*LinearClient).ListIssuesByStates` | **37** | **9** |

funlen: the target was also funlen-flagged (`//nolint:gocognit,funlen`); the decomposed orchestrator is now well under the 80-line / 50-statement budget. All new helpers are under both gates and add **no** new `//nolint`.

The `//nolint:gocognit,funlen // baseline (#521)` directive was **removed** from the target. golangci-lint scoped to `./internal/tracker/...` reports **0 issues**, and the target no longer appears in the `gocognit -over 29` list.

## Zero-behavior-change invariants preserved

- **Max-pages terminal error stays a plain `fmt.Errorf("linear pagination exceeded %d pages", ...)`** — NOT `ErrIssueListingCapped`. Linear is intentionally not "aligned" with github/gitea cap semantics; this keeps `TestListIssuesByStatesErrorsWhenMaxPagesExceeded` green.
- **`isTodoState` gate stays gated** — blockers fetched only for Todo state. Re-verified by `TestListIssuesByStatesPaginates` (requests==3) plus the new test below.
- **nil-vs-empty preserved** — the accumulator `var issues []Issue` stays nil-initialized; a `!HasNextPage` page that mapped no nodes returns a nil slice (`fetchLinearIssuesPage` returns a nil issue slice for an empty page; `append(nil)` of zero elements stays nil).
- **Missing-endCursor error path returns `nil, err`**, not the accumulated slice (matched the original `return nil, NewError(...)`).
- **Fresh labels slice per node; double-applied trim+lower intentional; same `ctx` threaded** (not a fresh context) so `TestLinearClient_EnforcesRequestTimeout` still pins the deadline. No named returns on the orchestrator.
- **GraphQL query byte-identical**, including the absence of `customFieldValues` (#326).

## Characterization tests (committed first, mutation-verified non-placebo)

Added to `internal/tracker/linear_test.go` (package `tracker`, httptest pattern), pinning coverage gaps the existing suite left unasserted:

- `TestListIssuesByStatesSkipsBlockerFetchForNonTodoIssue` — a non-Todo issue ends with empty `BlockedBy` AND triggers no `ListIssueInverseRelations` request. Mutation: flipping `isTodoState` to `!isTodoState` makes it fail.
- `TestListIssuesByStatesSkipsBlankLabelNames` — blank / whitespace-only label names are dropped; survivors trimmed+lowered. Mutation: dropping the `if strings.TrimSpace(label.Name) != ""` skip yields `,,backend`, failing the test.
- `TestListIssuesByStatesReturnsParseErrorForMalformedTimestamp` — a malformed `createdAt` or `updatedAt` routed THROUGH `ListIssuesByStates` surfaces the parse error (createdAt parsed first). Mutation: swallowing the parse errors (`_ =`) returns nil error, failing both subcases.

Each mutation was run against the branch, confirmed to fail the targeted test, then restored with `git checkout`; the working tree was reconfirmed clean against HEAD after each restore (#469/#483 footgun guard). Tests committed in a separate first commit before the decomposition.

## Local gates (all pass)

- `gofmt -l` on touched files: empty
- `go vet ./...`: clean
- `go test -race -covermode=atomic ./internal/tracker/...`: ok (86.3% coverage)
- golangci-lint (`./internal/tracker/...`): 0 issues; target gone from funlen/gocognit
- `go build ./cmd/worker ./cmd/tui`: ok
- `gocognit -over 29 ./internal ./cmd`: target absent
- consumer packages `internal/worker`, `internal/orchestrator` race tests: ok

## Size-gate classification

**within budget** — small, atomic refactor + the characterization coverage it depends on.

Refs #521

### DIFF
--- a/internal/tracker/linear.go
+++ b/internal/tracker/linear.go
@@ -129,33 +129,17 @@ func (c *LinearClient) ListActiveIssues(ctx context.Context) ([]Issue, error) {
 	return c.ListIssuesByStates(ctx, c.Config.ActiveStates)
 }
 
-func (c *LinearClient) ListIssuesByStates(ctx context.Context, states []string) ([]Issue, error) { //nolint:gocognit,funlen // baseline (#521)
-	if c.APIKey == "" {
-		return nil, NewError(CategoryMissingTrackerAPIKey, "Linear API key is required", nil)
-	}
-	projectSlug := strings.TrimSpace(c.Config.ProjectSlug)
-	if projectSlug == "" {
-		return nil, NewError(CategoryMissingTrackerProjectSlug, "Linear project slug is required", nil)
-	}
-	// SPEC §17.3: empty fetch_issues_by_states([]) returns empty without API call.
-	nonEmpty := make([]string, 0, len(states))
-	for _, s := range states {
-		if t := strings.TrimSpace(s); t != "" {
-			nonEmpty = append(nonEmpty, t)
-		}
-	}
-	if len(nonEmpty) == 0 {
-		return nil, nil
-	}
-	// customFieldValues is intentionally NOT requested. Linear's GraphQL
-	// schema does not expose a customFieldValues field on Issue (only
-	// customerTicketCount matches `custom*`); requesting it produced
-	// HTTP 400 GRAPHQL_VALIDATION_FAILED on every poll (#326). The
-	// upstream Elixir reference (elixir/lib/symphony_elixir/linear/client.ex)
-	// also omits any custom-field fragment for the same reason.
-	// `services[].tracker.custom_fields` route predicates are rejected at
-	// workflow load time until Linear surfaces a working query field.
-	query := `query ListIssues($projectSlug: String!, $states: [String!], $first: Int!, $after: String) {
+// listLinearIssuesQuery is the fixed ListIssues GraphQL query.
+//
+// customFieldValues is intentionally NOT requested. Linear's GraphQL
+// schema does not expose a customFieldValues field on Issue (only
+// customerTicketCount matches `custom*`); requesting it produced
+// HTTP 400 GRAPHQL_VALIDATION_FAILED on every poll (#326). The
+// upstream Elixir reference (elixir/lib/symphony_elixir/linear/client.ex)
+// also omits any custom-field fragment for the same reason.
+// `services[].tracker.custom_fields` route predicates are rejected at
+// workflow load time until Linear surfaces a working query field.
+const listLinearIssuesQuery = `query ListIssues($projectSlug: String!, $states: [String!], $first: Int!, $after: String) {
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

PR #534, opened 2026-05-30
refactor(orchestrator): decompose reconcileTick into single-concern helpers (#521)

hey-yvan:
## What

Pure extract-function decomposition of `(*Poller).reconcileTick`
(`internal/orchestrator/poller.go`) for the #521 gocognit-baseline cleanup.
`reconcileTick` implements SPEC §16.3 `reconcile_running_issues`; this is a
**structural extraction only — zero behavior change**.

## Helpers extracted (single concern each)

| Helper | Concern |
|--------|---------|
| `mergeRefreshedActiveStates` | Fold the narrow per-issue state refresh back into the active set: value-copy state update for issues still active, drop issues that left the active set. Mutates `activeIssuesByID` by reference; `refreshedIssuesByID` read-only. |
| `(p *Poller) reconcileActiveRunsPartB` | The SPEC §16.3 Part B routing-vs-non-routing branch (`ReconcileTrackerIssuesAndWait` when routed, else `RefreshActiveTrackerIssues`). Its error stays **fatal** to the tick. Keeps the #340/§18.1 comment on the routing arm. |
| `(p *Poller) deriveInactiveIssues` | Build the inactive/terminal set: refreshed running issues that left the active set into a configured inactive/terminal state. |
| `(p *Poller) appendInactiveFromStateGroups` | The explicit terminal/inactive state-group fan-out (`ListIssuesByStates`), joining per-group errors in loop order, non-fatally. |
| `addInactiveListedIssues` | Per-listing accumulation: skip empty IDs and still-active issues, add the rest. |

`deriveInactiveIssues` was split one level further than the original plan
(`appendInactiveFromStateGroups` + `addInactiveListedIssues`) so every new
helper clears the repo's local `gocognit.min-complexity: 10`
threshold without any new `//nolint`.

The parent `reconcileTick` keeps Part A inline (its fatal-vs-non-fatal split is
load-bearing) and keeps the `fetchErr` accumulation + the SPEC §16.3
Part-A-before-Part-B ordering visible.

## gocognit

| Function | Before | After |
|----------|--------|-------|
| `(*Poller).reconcileTick` | **33** | **6** |

New helpers: `deriveInactiveIssues` 5, `mergeRefreshedActiveStates` 7,
`appendInactiveFromStateGroups` 3, `addInactiveListedIssues` 5,
`reconcileActiveRunsPartB` 1 — all well under the gate. `reconcileTick` is gone
from `gocognit -over 29 ./internal ./cmd`. funlen was **not** flagged on this
target (the baseline directive was `//nolint:gocognit`, gocognit-only).

## nolint baseline removed

The `//nolint:gocognit // baseline (#521)` directive on `reconcileTick` is
removed. No new `//nolint` added anywhere.

## Zero-behavior-change invariants preserved (the high-risk ones from the spec)

- **errors.Join order is unchanged.** `fetchErr` accumulates in the same order
  (Part-A-nonfatal, then `refreshRunningIssueStates` err, then the per-group
  fan-out errors via `deriveErr`), and the final return is
  `errors.Join(reconcileErr, fetchErr)` with **reconcileErr first**. The
  per-group errors are now returned from `deriveInactiveIssues` as a single
  joined `deriveErr` and joined into `fetchErr` by the caller; `errors.Join`
  flattening makes the `Error()` string and `errors.Is`/`errors.As` traversal
  identical (verified against `TestPollOnceCancelsTerminalIssueBeforeReturningLaterInactiveFetchError`
  (substring) and the `errors.Is`-based reconciliation tests).
- **Fatal returns stay fatal with a nil map.** The Part A `ctx.Err() != nil`
  early-return and **both** Part B arms return `(nil, err)` from
  `reconcileTick` — never converted into accumulated errors.
- **Map mutation by reference.** `activeIssuesByID`/`activeByID` are mutated by
  reference; **`refreshedIssuesByID` is read-only** in merge, derive, and the
  `ReconcileTrackerIssuesAndWait` hand-off. No helper writes to it.
- **Value-copy state update** preserved in `mergeRefreshedActiveStates`
  (`existing := m[id]; existing.State = …; m[id] = existing`), not pointer
  mutation.
- `normalizedStates(...)` recomputed identically where used;
  `WorkerExitTimeout` passed unchanged to every orchestrator call. No named
  returns.

## Characterization tests (committed first, mutation-verified non-placebo)

New file `internal/orchestrator/poller_reconciletick_test.go`:

- `TestReconcileTickPartACancelledContextReturnsFatally` — Part A
  ctx-cancelled fatal early-return surfaces the error and returns a nil map.
  Killed by mutating `return nil, err` → `return nil, nil`.
- `TestReconcileTickDropsNonActiveNonInactiveRefreshedIssueFromInactiveSet` —
  a refreshed running issue in a non-active, non-configured-inactive state is
  dropped from the active set but **not** added to `inactiveByID`. Killed by
  flipping the `!p.isConfiguredInactiveState` guard.
- `TestReconcileTickSkipsEmptyIDIssueFromStateGroupFanOut` — the empty-ID skip
  in the fan-out loop. Killed by neutralizing `if issue.ID == "" { continue }`
  in the fan-out helper.
- `TestReconcileTickRequiresStateTracker` — the nil-`stateTracker` sentinel
  guard. Killed by neutralizing the guard.

Each test was mutation-verified non-placebo **after the decomposition** by
breaking the exact targeted branch in its new helper location, confirming the
test FAILS, then restoring; the working tree was re-confirmed against HEAD with
`git status`/`git diff` after each restore (the #469/#483 footgun).

### Note on the "always-accumulate" Part A mutation

The orthogonal mutation that drops the Part A guard entirely (so the
cancelled-ctx case falls through to Part B instead of bailing) is **not
behaviorally observable** through the public Poller/Orchestrator surface: under
a cancelled context every downstream orchestrator submit also fails with the
same context error, so the `(nil map, context error)` result is identical with
or without the guard, and the only divergent signal (a non-nil inactive map) is
gated behind a non-deterministic actor `select` that cannot be forced. The
complementary over-eager-fatal direction (treating a **live-ctx** Part A timeout
as fatal) is already pinned by the existing
`TestPollOnceContinuesReconciliationWhenStalledRunCleanupTimesOut`. Together
they fence the fatal-vs-non-fatal split. This is documented in the test file's
header comment.

## Local gates (all pass, scoped to package)

- `gofmt -l` on touched files: clean
- `go vet ./internal/orchestrator/...`: clean
- `go test -race -covermode=atomic ./internal/orchestrator/...`: ok (85.7%)
- `golangci-lint run --config=.golangci.yml ./internal/orchestrator/...`: **0 issues**
- `go build ./cmd/worker ./cmd/tui`: ok
- `gocognit -over 29 ./internal ./cmd`: `reconcileTick` gone
- `go mod tidy`: go.mod/go.sum clean

## Size gate

**within budget** — refactor + characterization tests; no scope creep.

Refs #521

### DIFF
--- a/internal/orchestrator/poller.go
+++ b/internal/orchestrator/poller.go
@@ -201,7 +201,7 @@ func (l activeIssueListerFromStates) ListActiveIssues(ctx context.Context) ([]tr
 	return l.tracker.ListIssuesByStates(ctx, l.states)
 }
 
-func (p *Poller) reconcileTick(ctx context.Context, activeIssues []tracker.Issue) (map[string]tracker.Issue, error) { //nolint:gocognit // baseline (#521)
+func (p *Poller) reconcileTick(ctx context.Context, activeIssues []tracker.Issue) (map[string]tracker.Issue, error) {
 	if p.stateTracker == nil {
 		return nil, errors.New("orchestrator poller reconciliation requires state tracker")
 	}
@@ -224,6 +224,22 @@ func (p *Poller) reconcileTick(ctx context.Context, activeIssues []tracker.Issue
 	if err != nil {
 		fetchErr = errors.Join(fetchErr, err)
 	}
+	mergeRefreshedActiveStates(activeIssuesByID, refreshedIssuesByID, activeStateKeys)
+	if err := p.reconcileActiveRunsPartB(ctx, activeIssuesByID, refreshedIssuesByID, activeStateKeys); err != nil {
+		return nil, err
+	}
+	inactiveByID, deriveErr := p.deriveInactiveIssues(ctx, activeIssuesByID, refreshedIssuesByID, activeStateKeys)
+	fetchErr = errors.Join(fetchErr, deriveErr)
+	reconcileErr := p.orchestrator.ReconcileInactiveTrackerIssuesAndWait(ctx, inactiveByID, normalizedStates(p.reconcile.TerminalStates), p.reconcile.WorkerExitTimeout)
+	return inactiveByID, errors.Join(reconcileErr, fetchErr)
+}
+
+// mergeRefreshedActiveStates folds the narrow per-issue state refresh back into
+// the active set: issues still in an active state keep their stored entry with
+// the refreshed state copied in (value-copy update, not pointer mutation),
+// while issues that left the active set are dropped from activeIssuesByID. It
+// mutates activeIssuesByID in place and treats refreshedIssuesByID as read-only.
+func mergeRefreshedActiveStates(activeIssuesByID, refreshedIssuesByID map[string]tracker.Issue, activeStateKeys map[string]struct{}) {
 	for id, issue := range refreshedIssuesByID {
 		if isActiveTrackerState(issue.State, activeStateKeys) {
 			if existing, ok := activeIssuesByID[id]; ok {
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

Issue #542, opened 2026-05-31
doctor: --deploy=binary real mode passes "Codex sandbox compatible" on a host where the codex bwrap sandbox cannot start commands (bwrap uid-map EPERM)


hey-yvan:
## Summary

On a binary (non-Docker) deploy, `worker --doctor --deploy=binary --mode=real`
reports:

```
PASS Codex sandbox: selected profile is compatible with this preflight
```

…but at runtime **every agent shell command fails immediately** with:

```
bwrap: setting up uid map: Permission denied
```

so the codex app-server agent cannot inspect the repo, edit files, run the
verify commands, commit, push, or open the PR. The doctor's "sandbox
compatible" check is a static/config-level check — it never actually exercises
the codex sandbox's ability to start a command — so the failure surfaces only
**after** a full agent turn has been dispatched and tokens burned (and then
re-dispatched on every poll because the issue stays in an active state).

## Environment

- Host: Ubuntu-family, kernel 6.17, binary deploy of `dist/worker`.
- codex-cli 0.135.0, `agent.default: codex-app-server`,
  `codex.thread_sandbox: workspace-write`,
  `turn_sandbox_policy.type: workspaceWrite` (networkAccess: true).
- Root cause is host policy: `kernel.apparmor_restrict_unprivileged_userns = 1`
  (Ubuntu 24.04+ AppArmor restriction on unprivileged user namespaces). bwrap
  0.9.0 cannot set up the uid map. Reproducible directly:
  ```
  $ bwrap --ro-bind / / --unshare-user --uid 0 echo ok
  bwrap: setting up uid map: Permission denied
  ```

## Why this is a doctor gap (not just an environment problem)

The host restriction is environmental, but the harness-engineering contract is
"failures are configuration problems, caught at preflight." The doctor already
runs a live codex app-server JSON-RPC probe (which PASSes here) — it should also
verify the sandbox can actually **execute a command**, because that is the
single most load-bearing capability for `codex-app-server` and the most common
way a Linux host silently can't run it.

Observed blast radius when it isn't caught at preflight:
- Agent burns a full turn (here ~590k input tokens) producing only a "blocked"
  message; nothing is edited/tested/pushed.
- Because the dispatched issue moves to an active tracker state, the worker
  re-dispatches it every poll → repeated token burn until an operator notices.

## Suggested fix

In `internal/doctor` (real mode, codex path), add a check that actually runs a
trivial command through the codex sandbox the runner will use — e.g. a
[...]

PR #554, opened 2026-05-31
fix(orchestrator): cool down recurring sandbox-startup failures instead of hot re-dispatch (#550)

### LINKED ISSUE #550: orchestrator: back off hot re-dispatch when a run fails with a recurring sandbox-startup error
## Summary

Deferred from #542 (PR #549). The #542 doctor probe catches an unusable codex bwrap sandbox at **preflight**, but if a host regresses at runtime (or the operator skips `--doctor`), a dispatched run still fails immediately with `bwrap: setting up uid map: Permission denied` and — because the issue stays in an active tracker state — the worker **re-dispatches it every poll**, burning tokens in a tight loop until an operator notices (the blast radius quoted in #542).

## Why this is separate from #542
#542 is a preflight-detection change in `internal/doctor`. This is a **runtime reconcile/retry-path** change: when a run fails with a sandbox-startup error that will recur identically, the orchestrator should mark it blocked / apply backoff rather than hot-retry. That path overlaps #543 (orchestrator cancel/retry classification), so it belongs with the retry machinery, not the doctor.

## Suggested approach
- Classify a sandbox-startup failure (e.g. bwrap uid-map EPERM) as a non-retryable-without-backoff outcome in the runner outcome classifier (`internal/runner/codex_app_server.go`) and the worker failure path.
- In the orchestrator retry path, apply an `external_blocker`/cooldown-style backoff (see `docs/runbooks/batch-issue-processing.md` BLOCKED.json semantics) instead of immediate re-dispatch.
- Earned by the #542 blast-radius observation: ~590k input tokens burned per failed turn, repeated every poll.

## Cross-refs
- #542 (preflight detection), #543 (reconcile cancel/retry classification).

hey-yvan:
Closes #550.

## Problem
When a dispatched run fails with a recurring codex/bwrap sandbox-startup denial (`bwrap: setting up uid map: Permission denied`), the issue stays in an active tracker state, so the worker re-dispatches it on the standard failure-retry backoff and burns a full agent turn's tokens (~590k input/turn, the #542 blast radius) every cycle until an operator notices. #542 catches this at **preflight** (`internal/doctor`); this is the **runtime** catch for a host that regresses after preflight or an operator who skipped `--doctor`.

## Approach
Classify the failure and route it to the **existing external-blocker cooldown machinery** instead of the hot failure-retry loop — no near-duplicate retry kind (harness principle: few sharp tools).

1. **`runner.SandboxStartupError`** (typed) + `classifySandboxStartupFailure` (`internal/runner/sandbox_startup.go`): the denial originates in codex's vendored bwrap subprocess, so there is no typed Go error to match — text classification of the external tool's output is the only signal (the AGENTS.md-sanctioned exception, same as the quota-backoff retry-after parser and the doctor probe). Matches bubblewrap's own fatal-line prefix (`bwrap:`) co-occurring with a denial token, plus the bwrap-only `setting up uid/gid map` phrases as a standalone signal. Runs only against the classified error + captured output and **does not** mask the more-specific outcomes (timeout, stall, read/turn timeout, input-required, quota backoff). `Detail` is a fixed, output-free string so raw subprocess output (potential secrets) never reaches an error string.
2. **Worker** (`internal/worker/runtask.go`): `handleRunnerFailure` → `sandboxStartupBlocked` parks the run on the `ExternalBlocked` cooldown (1h) with a synthesized in-memory blocker (never touches `.aiops/BLOCKED.json`, so the agent's own artifact is not shadowed) and emits a dedicated `task.EventSandboxStartupBlocked` so an operator can tell a host sandbox denial apart from an agent-declared dependency block. This goes through the **failure** path — it does **not** stamp `PhaseSucceeded`.
3. The orchestrator side (`applyExternalBlocked` → `scheduleExternalBlockerRetry` → `fireWakeSignal`) is reused unchanged: park blocked, re-check on the cooldown, re-dispatch via the poll loop; attempt budget is not consumed (a host problem must not exhaust the agent's failure-retry budget).

## SPEC alignment
Not a SPEC deviation — additive runtime hardening within the existing external-blocker/quota-backoff extension envelope (same class as those). No new `DEVIATIONS.md` row. Adjacent-path audit (AGENTS.md cross-cutting checklist item 1): `ExternalBlocked` / `RetryKindExternalBlocker` / `BlockerArtifact` consumers (`poller.go` Spawn, `actor_finalize.go` applyExternalBlocked, `actor_retry.go` fireWakeSignal) are generic on the flag and handle the new producer path correctly; `isReconcileCancel` keys on the context cause (not the error value), so the reconcile-cancel guard still wins.

## Acceptance criteria
- [x] Sandbox-startup failure classified as non-retryable-without-backoff in the runner outcome classifier (`internal/runner/codex_app_server.go` → `classifySandboxStartupFailure`) and the worker failure path.
- [x] Orchestrator applies an `external_blocker`/cooldown-style backoff (1h) instead of immediate re-dispatch.

## Validation
- `gofmt -l`, `go vet ./...`, `go mod tidy` clean; `go build ./cmd/worker ./cmd/tui`.
- `go test -race -covermode=atomic ./...` green; `golangci-lint` 0 issues (funlen/gocognit included — `runAgent` kept under budget by extracting `handleRunnerFailure`).
- **Mutation-verified** (against the committed artifact): blanking `hasBwrapDenial` fails the runner classification + `Run`-level tests; disabling the `IsSandboxStartup` worker branch fails the cooldown-routing test.
- Coverage: unit table (positive uid/gid-map + netns variant, negative `bwrap`-without-denial + unrelated permission-denied), precedence guard (6 specific outcomes not masked), a **`Run`-level** test driving a real `turn/failed` bwrap denial through `CodexAppServerRunner.Run` (pins the load-bearing wiring), and worker routing (ExternalBlocked + 1h + dedicated event + no RUN_SUMMARY/succeeded).

## Size gate
`size-gated: justified overage` (~438 added; ~243 is regression/`Run`-level/negative test coverage + earned comments, atomic to the change and not compressible without losing coverage). Surfaced for size-gate sign-off.

## Review notes
The local Claude pre-push review was applied (broadened signature to cover non-uid-map bwrap denials, dropped an over-generic phrase, added the `Run`-level + negative coverage, corrected a comment). `@codex review` will be triggered on the head commit.

### DIFF
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -105,7 +105,12 @@ func (CodexAppServerRunner) Run(ctx context.Context, in RunInput) (Result, error
 		res.OutputHead = string(head)
 	}
 	res.OutputTail = tail
-	return classifyAppServerOutcome(ctx, res, runErr, waitErr, client.readTimeoutMs, start, elapsed)
+	res, outcomeErr := classifyAppServerOutcome(ctx, res, runErr, waitErr, client.readTimeoutMs, start, elapsed)
+	// Re-tag a recurring codex/bwrap sandbox-startup denial so the worker parks
+	// it on a cooldown instead of hot-retrying every poll (#550). Done here, not
+	// inside classifyAppServerOutcome, because the captured output (OutputHead/
+	// Tail, where bwrap's stderr lands) is only assembled above.
+	return res, classifySandboxStartupFailure(outcomeErr, res)
 }
 
 // setupAppServerCommand builds the `codex app-server` *exec.Cmd ready to start:
--- a/internal/runner/runner.go
+++ b/internal/runner/runner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/xrf9268-hue/aiops-platform/internal/task"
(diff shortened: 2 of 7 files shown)


hey-yvan:
@codex review

Issue #544, opened 2026-05-31
codex-app-server: agent `go test` fails under workspace-write sandbox (default GOCACHE ~/.cache/go-build not writable)


hey-yvan:
## Summary

On a binary deploy with the codex-app-server runner under the default
`workspace-write` sandbox, the agent's `go test ./...` fails because the default
Go build cache (`~/.cache/go-build`) is outside the sandbox's writable roots.
The agent observed and worked around it (from its committed `RUN_SUMMARY.md`):

> Verification: `GOCACHE=/tmp/go-build go test ./...` passed. A plain `go test
> ./...` was attempted first but the sandbox could not read/write the default Go
> build cache under `~/.cache/go-build`.

Every Go workflow run through `codex-app-server` with `thread_sandbox:
workspace-write` will hit this on the agent's first `go test`. It is non-fatal
(a capable agent redirects `GOCACHE`), but it wastes a turn and is a sharp edge
for the documented binary-deploy + Go path.

Note: the worker's own `verify.commands` `go test ./...` is unaffected because
the worker runs verify outside the codex sandbox; only the agent's in-sandbox
commands hit it.

## Suggested fix (pick one, document it)

- Document in `docs/runbooks/binary-deployment.md` that Go workflows should add
  `~/.cache/go-build` (and `~/.cache/go-mod`/`GOMODCACHE`) to
  `codex.turn_sandbox_policy.writableRoots`, or set `GOCACHE` to a workspace/tmp
  path via a workspace `before_run` hook / env.
- Or have the runner default `writableRoots` to include the standard Go/tool
  caches when a Go toolchain is detected.
- Or set `GOCACHE`/`GOMODCACHE` into the agent environment to a writable,
  sandbox-visible path.

## Environment

codex-cli 0.135.0, `agent.default: codex-app-server`,
`thread_sandbox: workspace-write`, `turn_sandbox_policy.type: workspaceWrite`
(`writableRoots: []`, `networkAccess: true`), Go 1.25, Linux binary deploy.

PR #555, opened 2026-05-31
fix(runner): reap stale agent Go build caches

### LINKED ISSUE #551: runner: reap per-workspace agent Go build caches under TMPDIR (TTL/size cap)
## Summary

Follow-up from #544 (PR #548). The codex agent's GOCACHE is pinned to a per-workspace directory under `os.TempDir()/aiops-go-cache/build/<key>` so it is sandbox-writable and isolated across repos/issues (#544 / codex review #548 P1). These directories live **outside** the per-issue workspace tree, so the workspace GC (see `docs/runbooks/workspace-cache.md`) never reaps them — on a long-lived worker they accumulate one GOCACHE dir per distinct workspace.

GOMODCACHE is shared (`.../aiops-go-cache/mod`) and bounded by dedup, so this is specifically about the per-workspace GOCACHE dirs.

## Acceptance criteria
- A reaper removes `aiops-go-cache/build/<key>` entries by TTL and/or a total-size cap, OR cache teardown is tied to workspace destroy (remove the matching `<key>` dir when its workspace is GC'd).
- Bounded growth is verified by a test (e.g. stale-entry eviction).
- `docs/runbooks/workspace-cache.md` documents the agent Go cache lifecycle.

## Earned by
Codex review on PR #548 (MEDIUM): caches outside the workspace tree are never reaped by existing cleanup.

hey-yvan:
## Summary
- Remove per-workspace Codex agent Go build caches when the matching worker workspace is removed.
- Add a lazy `TMPDIR/aiops-go-cache/build/<key>` TTL sweep on Codex app-server setup that skips currently active app-server cache keys and aggregates per-entry cleanup errors.
- Document the agent Go cache lifecycle, including that the 7-day TTL uses top-level `build/<key>/` directory mtime rather than last cache hit.

Closes #551

## Acceptance criteria
- [x] `aiops-go-cache/build/<key>` entries are bounded by workspace teardown plus a lazy 7-day TTL sweep.
- [x] Bounded growth is verified by tests for stale eviction, active-cache skipping, continued cleanup after one removal error, and workspace-removal cache teardown.
- [x] `docs/runbooks/workspace-cache.md` documents the agent Go cache lifecycle.

## Verification
- `gofmt -l $(git ls-files '*.go')`
- `go vet ./...`
- `go test -race -covermode=atomic ./...`

## Mutation checks
- Disabled the setup-time reaper call; `TestSetupAppServerCommandReapsStaleGoBuildCaches` failed as expected, then passed after restore.
- Changed workspace teardown to remove the wrong cache key; `TestRemoveIssueWorkspaceRemovesGoBuildCache` failed as expected, then passed after restore.

## Pre-push review
- Codex diff-only review on `77a5b08`: `MERGE-READY`, no findings.
- Claude diff-only review found TTL wording ambiguity; addressed by documenting mtime-based TTL semantics. Final Claude rerun is still pending because this PR was opened immediately per maintainer request.

## Size gate
- [x] `within budget` — 7 files, 295 changed LOC; fits default `policy.max_changed_files` 12 / `policy.max_changed_loc` 300.
- [ ] `size-gated: justified overage` — rationale: n/a
- [ ] `size-gated: split recommended` — rationale: n/a

### DIFF
--- a/docs/runbooks/workspace-cache.md
+++ b/docs/runbooks/workspace-cache.md
@@ -243,6 +243,15 @@ recommended strategy is age-based pruning, gated on operator preference:
   cache. They survive worktree cleanup intentionally, because deleting
   them costs you a fresh `git clone` on the next task. Only purge a
   mirror when you change its upstream URL or the cache disk fills up.
+- **Agent Go caches** (`$TMPDIR/aiops-go-cache/...`): the Codex
+  app-server runner injects a shared `mod/` cache plus a per-workspace
+  `build/<key>/` cache so sandboxed Go commands can write their cache
+  files. `build/<key>/` entries are removed when the worker deletes the
+  matching workspace and are also lazily reaped on app-server setup when
+  the top-level `build/<key>/` directory mtime is older than 7 days. That
+  TTL is not a last-cache-hit signal, so a warm but idle cache can be
+  rebuilt. `mod/` is intentionally shared and is not managed by workspace
+  cleanup.
 
 The Go API for cleanup is `(*workspace.Manager).Cleanup(ctx, maxAge)`.
 It walks `$AIOPS_WORKSPACE_ROOT`, deletes each task directory whose mtime
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -43,6 +44,8 @@ const (
 const PromptPath = ".aiops/PROMPT.md"
 
 func (CodexAppServerRunner) Run(ctx context.Context, in RunInput) (Result, error) {
+	releaseGoBuildCache := markActiveGoBuildCache(in.Workdir)
+	defer releaseGoBuildCache()
 	if err := validateAppServerWorkdir(in.Workdir); err != nil {
 		return Result{}, err
 	}
@@ -124,6 +127,9 @@ func setupAppServerCommand(ctx context.Context, in RunInput) (cmd *exec.Cmd, dir
 	// Pin the agent's Go toolchain caches to a sandbox-writable, per-workspace
 	// path so its first `go test` does not fail on the default $HOME-based
 	// caches that lie outside codex's workspace-write sandbox (#544).
+	if err := reapSandboxGoBuildCaches(); err != nil {
+		log.Printf("event=go_build_cache_reap_failed error=%q", err)
+	}
 	env = withSandboxGoToolchainCaches(env, in.Workdir)
 	cmd, directCodexExec, err = buildCodexAppServerCmd(ctx, in, env)
 	if err != nil {
--- a/internal/runner/go_cache.go
+++ b/internal/runner/go_cache.go
@@ -179,7 +179,11 @@ func reapGoBuildCacheEntry(root string, entry os.DirEntry, cutoff time.Time) err
 		return nil
 	}
 	// Directory mtime is only a stale-cache signal after active runs release
-	// their key; live app-server sessions are excluded above.
+	// their key; live app-server sessions are excluded by the check above.
+	// That active check and this removal are deliberately not atomic: a run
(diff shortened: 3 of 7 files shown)


hey-yvan:
@codex review

hey-yvan:
## Local code review

Reviewed the full diff against `AGENTS.md`. Ran locally: `gofmt -l` clean, `go vet ./internal/runner ./internal/worker` clean, `go build ./...` clean, and `go test -race` on the affected packages green (the `internal/runner` package-level `FAIL` is the environment git signing-hook artifact — `fatal: failed to write commit object` in the `TestMockRunner*` git-commit tests — not a Go test failure).

**Verdict: approve-with-nits.** The design is sound: the cache key has a single source of truth (`SandboxGoBuildCachePath` → `workspaceCacheKey`), used identically by the setup, teardown, and active-set paths, so teardown can't delete the wrong directory. Path construction is safe (keys are FNV hex hashes, never attacker-controlled; the reaper only deletes direct `IsDir()` children of `aiops-go-cache/build/`). Error aggregation uses `errors.Join` + `%w` + `errors.Is(os.ErrNotExist)` and continues past per-entry failures. Active-set access is mutex-guarded. This is a legitimate runtime/harness concern (managing filesystem state the worker itself created in #544), not a SPEC deviation — no `DEVIATIONS.md` entry needed.

### HIGH
None.

### MEDIUM
1. **TOCTOU between the active-check and the remove is benign but undocumented** (`internal/runner/go_cache.go`, `reapGoBuildCacheEntry`). The active check and `goCacheRemoveAll` are not atomic. The in-flight run's own key is protected (marked active in `Run` before setup), so the only exposure is a *different* concurrent run marking active mid-sweep with an already-stale mtime — and the impact is bounded to deleting a cold build cache, which Go transparently rebuilds (no data loss). The risk is that a future reader mistakes it for a missing lock and "fixes" it by holding the mutex across the filesystem walk, serializing every run's startup. → **Addressed** in `d005b06`: expanded the existing comment to state the gap is benign-by-design and that the lock must not be widened over the walk.

2. **No concurrency/race test for the active-set.** `goBuildCacheActive` is new package-shared state mutated from per-run goroutines and read by the reaper; the three new tests exercise it serially only. `AGENTS.md` requires race coverage for new shared state. → **Addressed** in `4ccd5ce`: added `TestReapSandboxGoBuildCachesConcurrentWithActiveChurnIsRaceFree` (4 reapers vs. 8 mark/unmark goroutines), asserting a held key is never reaped and the active set drains back to only the held key. Mutation-checked: removing the mutex lock/unlock makes `-race` trip `DATA RACE`, so the test fails when the guard regresses.

3. **Unbounded synchronous sweep on a latency-sensitive startup path** (`internal/runner/codex_app_server.go`, the reaper call in `setupAppServerCommand`). Correctly best-effort and non-fatal, but every app-server launch now does `os.ReadDir` + per-entry `os.RemoveAll` with no work cap. Fine at expected scale; could add startup latency proportional to entry count, or stall on an NFS-backed `TMPDIR`. *Not addressed — flagging for awareness; acceptable as-is at expected scale. Consider bounding work per sweep if the build root can grow large.*

### LOW (not addressed; acceptable as-is)
- Global mutable test seams (`goCacheNow`, `goCacheRemoveAll`, `goBuildCacheActive`) are package-level vars — tests correctly avoid `t.Parallel()` and restore via `t.Cleanup`; just a latent hazard if a future parallel test in this package touches them.
- TTL boundary (`info.ModTime().After(cutoff)`) correctly treats age ≥ 7d as eligible — no off-by-one.

### Test coverage & size gate
Coverage is adequate and would catch a regression: each test exercises the real production path (`setupAppServerCommand` actually invokes the reaper; `RemoveIssueWorkspace` actually deletes via the public helper), uses real `os.Chtimes` mtime manipulation, and failure messages print actual + expected. Reverting the feature would fail them. The size-gate `within budget` self-classification holds — net new production code is ~90 LOC, every function is under the 80-line budget, and no new `//nolint` was added (the two follow-up commits add a race test + a comment, both outside the functional-LOC budget).

---

Issue #561, opened 2026-06-01
Move remaining worker post-run gates (secret-scan, RUN_SUMMARY, policy) off the agent boundary (D33 follow-up)


hey-yvan:
Follow-up to **DEVIATIONS D33** and PR #560 (which removed the worker `verify` gate). Umbrella: #67.

## Background

The 2026-05-15 gap audit ([`docs/audits/2026-05-15-spec-vs-go-gap-audit.md`](../blob/main/docs/audits/2026-05-15-spec-vs-go-gap-audit.md), "Deliberate extensions") flagged four worker-side post-run gates as sitting on the wrong side of the **SPEC §1** agent boundary — verification is the agent's job, the orchestrator is a scheduler/runner/tracker-reader. PR #560 removed the first one (`verify`). The remaining three still run as worker phases in `runPostRunGates` / `enforcePostRunPolicy`, after the agent's turn:

- **secret-scan** (`runSecretScan`, `verify.secret_scan`) — audit item 5: "defensible behavior, wrong side of the agent boundary." Also: post-#76 the agent pushes, so by the time the worker scans, the branch is already pushed — the gate flags but cannot prevent a leak.
- **RUN_SUMMARY gate** (`checkRunSummary`, `workspace.CheckSummary`) — audit item 7: "Tied to D8 — once PR creation moves to the agent (#76), the orchestrator has no PR to gate, so this disappears." It did not disappear.
- **policy** (`enforcePostRunPolicy`, `internal/policy`) — audit: "SPEC §15.5 frames this as a deployment-specific control, not an orchestrator phase. Better expressed as a workflow extension key than a hard-coded orchestrator step."

## Why this matters

Any worker post-turn phase races the D9 reconcile-cancel / §16.5 self-stop the same way `verify` did (#557): when the agent moves its issue inactive as its final action and reconcile-cancel wins, the gate is skipped. #560 eliminated that for `verify`; these three remain raceable.

## Acceptance criteria

- [ ] **secret-scan**: decide between (a) move to the agent's WORKFLOW/tool surface, or (b) keep as a documented SPEC §15.5 harness control with its own DEVIATIONS row and a written rationale for staying worker-side. If kept, address the race (it can still be skipped on reconcile-cancel).
- [ ] **RUN_SUMMARY gate**: evaluate whether it still earns its keep post-#76 (the audit predicted it disappears); either remove it or justify + track it.
[...]

Issue #557, opened 2026-06-01
Reconcile-cancel skips the worker verify gate when the agent moves its issue to an inactive state mid-run


hey-yvan:
## Summary

When the coding agent moves its tracker issue into an **inactive** state (e.g. `In Review`) as its final turn action, the next reconciliation tick cancels the still-active run with `CanceledByReconciliation` **before the worker reaches its `Verifying` phase** — so the worker's own `verify` gate (`go build` / `go test`) is silently skipped. The run also never appears as `completed` in `/api/v1/state`.

This was observed during a real binary-deployment E2E (codex `app-server`, 3 sequential Linear issues → 3 merged PRs on a throwaway `todo-cli` repo).

## Evidence (3 runs, identical workflow/prompt)

Same `WORKFLOW.md`, same prompt (prompt instructs the agent to move the issue to `In Review` as its final external action; `In Review` ∈ `tracker.inactive_states`).

| Run | Phase trace (tail) | verify ran? |
|-----|--------------------|-------------|
| F1 (AIS-49) | `StreamingTurn → Finishing → verify_start → verify_end(ok) → Succeeded` | ✅ yes |
| F2 (AIS-50) | `StreamingTurn → CanceledByReconciliation` | ❌ skipped |
| F3 (AIS-51) | `StreamingTurn → CanceledByReconciliation` | ❌ skipped |

So the verify gate firing is a **timing race**, and the skip is the **majority outcome (2/3)** — F1 only verified because it happened to finish before a reconcile tick (poll interval 30s) caught the issue already sitting in an inactive state.

`GET /api/v1/state` corroborates the second-order impact:
```
counts.completed = 1
completed = ["68764833-... (F1)"]   # F2/F3 absent from completed AND failed
```
i.e. a deployment that opened and merged 3 PRs reports `completed: 1`.

## Why this matters

- **`verify` is a safety gate** (`go build`/`go test`, secret-scan, etc.). Making it race-dependent means the worker's independent check is effectively optional: if the agent pushed broken code, the F2/F3 timing would not catch it. Only the agent's own self-run verification would stand.
- **Misleading run accounting**: a run that finished successfully (PR open, tracker `In Review`) shows up as `CanceledByReconciliation` and is invisible in the state snapshot, so operators can't tell "genuinely cancelled" from "completed but reaped by the finalization race".

## Root cause

[...]

Issue #76, opened 2026-05-13
[P0][spec-alignment] Orchestrator overreach: PR creation, git push, and ticket state writes belong to the agent (SPEC §1 boundary)


hey-yvan:
## Background

SPEC.md §1 *Problem Statement* draws a sharp boundary between the
orchestrator's role and the agent's role:

> "Symphony is a scheduler/runner and tracker reader."
> "Ticket writes (state transitions, comments, PR links) are typically
> performed by the coding agent using tools available in the workflow/runtime
> environment."
> "A successful run may end at a workflow-defined handoff state (for example
> `Human Review`), not necessarily `Done`."

The OpenAI [Symphony blog post (April 27, 2026)](../blob/main/docs/research/2026-04-27-openai-symphony-blog.md)
reinforces the same boundary:

> "We use [dynamic tool calls] to expose the raw `linear_graphql` function
> that executes arbitrary requests against Linear, without relying on MCP or
> exposing the access token to containers."

I.e. ticket mutations happen **inside the agent** through a dynamic tool the
orchestrator advertises, not from the orchestrator's own goroutine.

Today this platform crosses that boundary in three places in
`internal/worker/runtask.go`:

1. **PR creation**: `CreatePR` / `CreatePRWith` builds the body via
   `BuildPRBody` and calls `gitea.Client.CreatePullRequest` directly. The
   orchestrator is opening the PR.
2. **Git push**: `workspace.CommitAndPush` is called by the orchestrator,
   not the agent. The orchestrator is producing the artifact the PR is
   built from.
3. **Linear status transitions**: `OnClaim` writes "In Progress" and
   `OnPRCreated` writes "Human Review" via
   `tracker.Transitioner.MoveIssueToState` from the worker loop. The
   orchestrator is writing ticket state.

Each of these is documented today as a feature ("Linear status transitions
after handoff", #14) when SPEC treats them as the agent's responsibility.

## Severity

**P0** for the alignment trajectory. This is a deeper deviation than the
Postgres queue or the Gitea webhook because it shapes what the agent itself
is responsible for. Every other deviation we close (D1 app-server protocol,
D2 status transitions, D6 queue, D7 webhook) is downstream of getting this
boundary right:

- D1 (#64) is the *mechanism* by which the agent gains the tools to write
  tickets/PRs. Once app-server + dynamic tool calls land, this issue is
  unblocked.
- D2 (#14) is currently scoped as "the orchestrator writes Linear status";
[...]

Issue #67, opened 2026-05-12
[meta][spec-alignment] Track current D1–D24 Symphony SPEC deviations


hey-yvan:
## Purpose

Track every known deviation between `aiops-platform` and the upstream OpenAI Symphony SPEC / Elixir reference. This is a meta issue: implementation work happens in the linked child issues.

## Authoritative sources

- Upstream SPEC: https://github.com/openai/symphony/blob/main/SPEC.md
- Reference implementation: https://github.com/openai/symphony/tree/main/elixir
- Local project posture and rules: `AGENTS.md`
- Current deviation table: `DEVIATIONS.md`
- Latest full gap audit: `docs/audits/2026-05-15-spec-vs-go-gap-audit.md`

## Current scope

`DEVIATIONS.md` is the source of truth. It currently tracks D1–D24, including:

- D1 app-server / agent runner protocol (#64)
- D2/D8/D23 tracker and PR-write boundary (#14, #76, #97)
- D3 startup reconciliation (#68)
- D4/D10 workflow-file semantics (#72, #84)
- D6/D21 queue and orchestrator runtime state (#73, #95)
- D7 Gitea webhook vs polling (#74)
- D9 per-tick reconciliation (#78)
- D11–D24 additional SPEC gaps from the 2026-05-15 audit (#85–#98)

There are currently **no accepted deliberate extensions**. Anything that fails SPEC alignment should stay tracked as a deviation until either fixed or explicitly accepted under the high bar documented in `AGENTS.md`.

## Closing criteria

- [ ] Every D1–D24 row in `DEVIATIONS.md` is either fully closed by implementation or explicitly moved to an accepted-deviation section with rationale.
- [ ] Linked issues are closed or updated to point at their replacement issue.
- [ ] `README.md`, `AGENTS.md`, `DEVIATIONS.md`, and `docs/symphony-integration.md` no longer describe stale D1–D5-only posture or removed “deliberate extensions.”

hey-yvan:
All deviation issues filed. Reconciliation (D3, #68) is the next P0 after the existing #14 Linear status transitions land.

---
_Generated by [Claude Code](https://claude.ai/code)_

hey-yvan:
**Status: switching to fork**

[...]

PR #564, opened 2026-06-01
fix(worker): drop worker secret-scan gate, leave leaks to CI (#561)

hey-yvan:
## What & why

The worker ran the configured secret scanner (`verify.secret_scan`) as a **post-turn phase** in `runPostRunGates`, after the agent's turn (`runAgent`) returned. Per **SPEC §1** push is the coding **agent's** responsibility (#76), so by the time the worker scanned, **the branch was already pushed** — the gate could only *flag* a leak, never *prevent* it. Upstream Symphony has **no** such phase.

It also raced the D9 reconcile-cancel / §16.5 self-stop **exactly like the verify gate did in #557**: when the agent moves its issue to an inactive state as its final turn action and reconcile-cancel wins, it cancels the run mid-turn — **before** the post-turn secret scan — so the gate is silently skipped. A detection control that can be silently skipped is a broken detection control.

**Fix the deviation rather than patch around the race** (same move as PR #560 for verify). Remove the worker secret-scan gate. PR CI / the host's secret scanning is the backstop; an operator who wants pre-push scanning runs it from the WORKFLOW prompt, where it can actually prevent the push. With no worker post-turn scan there is nothing for reconcile-cancel to skip.

## Scope

This is the **secret-scan** gate of the D33 follow-up (#561). Each gate ships independently:
- ✅ **secret-scan** — removed here.
- ⏭️ **RUN_SUMMARY** gate — follow-up PR (remove; the audit predicted it disappears post-#76).
- ⏭️ **policy** — follow-up PR (kept as a documented SPEC §15.5 harness control; it drives the policy-violation corrective retry loop a prompt can't replicate).

## Changes

- **Removed:** `internal/workspace/secretscan.go` (whole file, + its test); the worker `runSecretScan`/`runSecretScanWith` wrappers, the `secretScanFn` type, the inline `secret_scan_{start,clean,violation,error}` events, and the secret-scan call in `runPostRunGates`.
- **Removed config:** `workflow.SecretScanConfig` + `VerifyConfig.SecretScan` (`VerifyConfig` now carries only `Commands`). `cappedBuffer` stays (shared with hooks).
- **Rejected at load:** `verify.secret_scan` via `rejectRemovedVerifyFields` so a stale workflow fails loud instead of believing the worker still scans (mirrors the `verify.{timeout,allow_failure,env_passthrough}` rejects from #560).
- **Docs:** deleted `docs/runbooks/secret-scanning.md`; scrubbed README / AGENTS / `security-posture.md` / `task-api.md` references. Incidentally fixed two #560 leftovers in `security-posture.md` (a stale "verify subprocesses" mention and a `verify.env_passthrough` mention — both already removed in #560).
- **DEVIATIONS D33** updated (verify + secret-scan removed; RUN_SUMMARY/policy follow-up).

## Verification (local, all green)

```
gofmt -l $(git ls-files '*.go')                 # clean
go mod tidy && git diff --exit-code go.mod go.sum # clean
go vet ./...                                      # clean
go build ./cmd/worker ./cmd/tui                   # ok
go test -race -covermode=atomic ./...             # all pass except the documented
                                                  #   sandbox-flaky codex app-server
                                                  #   timeout diagnostics test (untouched
                                                  #   here; passes in CI)
golangci (blocking gate, changed pkgs)            # 0 issues
```

**Mutation-verified:** removing the `verify["secret_scan"]` reject branch in `loader.go` makes `TestLoad_RejectsRemovedVerifyExecutionFields/secret_scan` fail (`Load(...) = nil; want rejection error`); restored with `git checkout` and confirmed the tree matches HEAD.

## Tests

- Deleted the secret-scan unit tests (`internal/workspace/secretscan_test.go`, `internal/worker/secretscan_test.go`) — they only pinned the removed gate.
- Added a `secret_scan` case to `TestLoad_RejectsRemovedVerifyExecutionFields` asserting the load-time rejection.

## Size-gate

`size-gated: justified overage` — 16 files (12 modified + 4 deleted), but net **−106 LOC** (37 insertions / 143 deletions). It is the atomic removal of one gate's machinery + its doc/comment references; splitting would leave dangling references mid-stream. No new functionality.

## Review notes (pre-push dual review)

Two independent diff-only reviewers (Codex `codex exec review` + Claude `feature-dev:code-reviewer`). Substantive diff passed both; all findings were LOW doc/comment hygiene (stale prose surviving the deleted gate), all fixed before push:
- `runtask.go` RunTask doc + `checkRunSummary` comment; `internal/worker/export_test.go` + `testhelpers_test.go` comments.
- `docs/security-posture.md` item 12 rewritten (was guidance + dead link for the removed hook).
- `DEVIATIONS.md` D29 `Failed`-map description aligned with the updated `state.go` comment (one source of truth).

`DECISION.md`'s dated "salvageable ~800 LOC" inventory still lists `internal/workspace.secretscan` (and `RunVerify`, already removed in #560) — left as-is: it is a point-in-time strategic record, not a live behavior doc.

Closes part of #561 (D33 follow-up). Umbrella #67.


hey-yvan:
@codex review

Issue #74, opened 2026-05-13
[P1][spec-alignment] Replace Gitea webhook trigger with Gitea poller (mirror cmd/linear-poller)


hey-yvan:
## Background

SPEC.md describes the trigger model as **tracker polling**: the orchestrator
periodically reads the tracker for issues in active states and dispatches
them. The reference implementation does not include a webhook ingress.

Today this platform takes two paths:

- **Linear**: `cmd/linear-poller` polls Linear active states (`AI Ready`,
  `In Progress`, `Rework`) and enqueues tasks. SPEC-aligned.
- **Gitea**: `cmd/trigger-api` runs an HTTP server that accepts Gitea
  issue-comment webhooks (`/v1/events/gitea`), verifies the signature,
  and enqueues tasks. **Not in SPEC.**

The Gitea webhook path was previously listed as a "deliberate extension"
in `DEVIATIONS.md`, justified on the grounds of low latency and not
consuming poll-side rate limit. Re-examining that justification under the
project's pre-release alignment posture:

## Why the deviation does not hold up

- **Latency value is marginal.** Agent runs are minute-scale. Polling at
  the same cadence as `cmd/linear-poller` (a few seconds) is well below
  the noise floor of a real run. Saving 5 seconds on dispatch does not
  matter when the task itself takes 10–30 minutes.
- **Rate-limit value is moot for self-hosted Gitea.** Self-hosted Gitea
  has no meaningful rate limit on `GET /repos/.../issues`; we already
  poll Linear (a hosted SaaS with stricter limits) without trouble at
  the same cadence.
- **Cost is real and ongoing:**
  - A second ingress surface (`cmd/trigger-api`) that has to be exposed
    publicly, secured with `GITEA_WEBHOOK_SECRET`, and reachable by Gitea.
  - Webhook signature verification code (`internal/gitea/webhook.go`)
    that must stay correct under signature spec changes upstream.
  - Two different code paths for "issue → task" depending on which
    tracker; they drift over time (event payload shape, dedupe semantics,
    error handling).
  - Operator-facing setup is asymmetric: Linear is `set the API key and
    go`, Gitea is `configure a webhook on every repo, expose a public
    URL, set a shared secret`.
- **The official SPEC approach is genuinely better here.** A Gitea
  poller mirroring `cmd/linear-poller` would unify the two paths under a
  single `tracker.Client`-driven model, simplify deployment (no public
  ingress required for the trigger), and make the SPEC mental model
[...]

PR #565, opened 2026-06-01
docs(agents): add harness principle 6 — upstream absence = over-design signal, delete don't relocate

hey-yvan:
## What & why

This port has **built-then-unwound the same class of misplacement repeatedly**: a worker/orchestrator phase took on a responsibility SPEC §1 assigns to the agent (or that SPEC excludes), it shipped, then it was removed — Postgres queue (#73/#407), Gitea webhook (#74), orchestrator PR/push/tracker-writes (#76), the worker verify gate (#557), and the worker secret-scan gate (#561; the RUN_SUMMARY gate is being unwound in the same #561 effort). Each was a dedicated removal on top of the original build, plus the bugs the misplacement caused — the #557 reconcile-cancel race was a *direct consequence* of running verify as a worker post-turn phase.

The existing harness principles (1 behavior-first, 2 earned-rules) evaluate a component *after* it exists; none is a **preventive up-front gate**. New **principle 6** adds that gate:

- Before adding any worker/orchestrator phase/gate/artifact/config that acts on the agent's output, `grep` the Elixir reference **first**; upstream absence is an over-design signal, not a feature gap.
- The default home for "check the agent's work before handoff" is the WORKFLOW prompt (agent-owned, pre-push, *preventive*), not a worker post-turn phase (post-push → can only flag, never prevent; and it races the D9 reconcile-cancel / §16.5 self-stop — exactly #557).
- When a component is on the wrong side of the boundary, **delete it** — don't relocate (move-to-prompt) or merely document (a DEVIATIONS row); that preserves scaffolding (principle 4). Keep only what names a behavior the scheduler/runner boundary genuinely permits *and* a prompt can't replicate (e.g. the `policy`-violation corrective retry loop).

## Scope

Governance-only; **no code change**. Single addition to AGENTS.md's "Harness engineering principles". Companion to #561 (the gate-removal work that earned the rule).

## Verification

Docs-only. Pre-push dual diff-only review (Codex `codex exec review` + Claude `feature-dev:code-reviewer`): final round **MERGE-READY from both**, zero findings. Earlier rounds caught and fixed a RUN_SUMMARY overclaim and a #561/D33 cross-doc divergence (the latter resolved by rebasing onto current main after #564 merged, so DEVIATIONS D33 now reads "#557 / #561"). Citations are issue-only and each resolves in DEVIATIONS.md (D6/D7/D8/D33).

Earned-by reflection prompted by the operator after the #561 session. Umbrella #67.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -129,6 +129,30 @@ govern *how* we evaluate components inside the SPEC-aligned envelope.
    focused tools (`linear_graphql`, one Gitea PR tool, etc.); resist
    the temptation to wrap every Gitea / Linear endpoint as a separate
    tool.
+6. **Upstream absence is an over-design signal; delete, don't relocate.**
+   Before adding any worker/orchestrator phase, gate, artifact, or config
+   that acts on the agent's output (verify, secret scan, run summary, diff
+   policy, push, PR, tracker write, …), `grep` the Elixir reference for an
+   equivalent *first*. **Upstream having no equivalent is a strong signal the
+   component is over-design, not a feature gap to fill.** The default home for
+   "check the agent's work before handoff" is the WORKFLOW prompt (agent-owned,
+   pre-push, *preventive*), not a worker post-turn phase — a worker phase runs
+   *after* the agent has already pushed (#76), so it can only flag, never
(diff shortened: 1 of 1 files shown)


hey-yvan:
@codex review

PR #625, opened 2026-06-03
docs(validation): autonomous E2E report + upstream Symphony comparison (#621–#624)

hey-yvan:
Closes #626

## Summary
Adds the validation report + evidence for a full **autonomous** binary + Linear + real-Codex E2E run of `cmd/worker` (`main` @ `c42c33d`, `aiops-platform/0.136.0`; includes #616/#617/#619). `docs/validation/2026-06-03-binary-linear-codex-autonomous-e2e.md` + `assets/` (6 screenshots, 2 `cmd/tui --raw` frames, 4 `/api/v1/state` snapshots) — same path/format as the prior `2026-05-26-docker-linear-e2e.md`.

10 disposable Linear issues, `max_concurrent_agents: 2`: 6 feature issues delivered autonomously (focused draft PR each, agent-owned `Todo → In Progress → In Review`); 4 pathological issues (2 blocking on absent secrets, 2 with impossible/unreachable acceptance) all degenerated into the **same** unbounded continuation loop instead of reaching `blocked`/`failure`-retry. Findings filed as **#621 #622 #623 #624**.

## SPEC alignment (AGENTS.md principle 6/7)
Check exactly one. When this PR changes a SPEC-sensitive path
(`internal/workflow/config.go`, a newly-added `internal/orchestrator/` or
`internal/worker/` file), the **PR Metadata** gate rejects the first option —
an extension upstream lacks must be cited or tracked, not waved through.

- [x] No new top-level `WORKFLOW.md`/`Config` key and no new worker/orchestrator phase/gate/artifact.
- [ ] Adds one, justified by an upstream **Elixir reference** — cite the matching file and line from the openai/symphony Elixir tree in the body below (a bare module name does not count; the gate looks for a concrete source path).
- [ ] Adds one, tracked as a **DEVIATIONS.md row** + an `area:spec-alignment` issue, updated in this PR.

Docs-only — touches `docs/validation/` exclusively; no code, no config keys, no worker/orchestrator phases.

## Size budget (AGENTS.md)
Classify into exactly one (review discipline; the size-gate is human-signed, not CI-blocked):

- [x] `within budget` — ≤12 production files / ≤300 production LOC (test files and generated code excluded).
- [ ] `size-gated: justified overage` — production diff over budget for correctness / regression / race-safety coverage that cannot be split without losing atomicity. **Needs human size-gate sign-off.**
- [ ] `size-gated: split recommended` — over budget for scope creep / separable concerns. Split instead.

0 production LOC (docs + assets only).

[...]

PR #628, opened 2026-06-04
[codex] Bound clean continuation loops

### LINKED ISSUE #621: Unfinishable issues degenerate into an unbounded continuation loop; blocked/failure-retry never reached
## Summary

In an autonomous binary + Linear + real-Codex run, **every issue whose task cannot actually be finished degenerates into an unbounded continuation loop** rather than reaching `blocked` or a `failure`-retry. The codex-app-server turn exits `Succeeded` regardless of whether the task is impossible; the agent keeps the issue in `In Progress` (an `active_state`); the SPEC §16.5 continuation loop re-dispatches immediately; `completed_total` increments per turn and climbs without bound.

Observed on `main` @ `c42c33d` (`aiops-platform/0.136.0`), `agent.default: codex-app-server`, `max_concurrent_agents: 2`, `max_turns: 12`, `active_states: [Todo, In Progress]`.

## Repro

4 disposable Linear issues whose acceptance is unsatisfiable: 2 needing an absent secret (`SENDGRID_API_KEY`, `FLY_API_TOKEN`), 2 with impossible/contradictory tests (self-contradictory assertions; a test that pings an unreachable `*.invalid` host). Each was moved to In Progress by the agent and then never escaped.

## Evidence

`runner_start` (re-dispatch) counts in one ~13-minute window:

```
feature issues:            1 each   (dispatch once, hand off, done)
AIS-67 (needs secret):     8
AIS-68 (needs secret):     5
AIS-69 (impossible tests): 2
AIS-70 (unreachable host): 0   ← never ran: starved by the loopers holding both concurrency slots
total runner_start: 21   total "Succeeded": 12  (== completed_total=12)
```

One cycle, repeating indefinitely:
```
event=runner_start … → run_phase_transition msg="Succeeded" → PreparingWorkspace → BuildingPrompt → event=runner_start …
```
The momentary `retrying` rows during this are `kind=continuation` (the 1s continuation delay), **never** `kind=failure`.

## Impact

- **Unbounded codex quota burn** — 13.79M tokens / 68 agent-minutes and still climbing when the run was force-stopped; on a metered plan this is open-ended.
- **Concurrency starvation** — the loopers permanently hold both `max_concurrent_agents` slots, so other Todo issues (AIS-70) get **zero** dispatches.
- **`blocked` is never reached** — the documented home for "agent waiting on an external dependency" stays empty even with two issues whose acceptance is explicitly unsatisfiable.
- **`completed_total` is meaningless** — it counts turn-completions of stuck issues, not delivered work (see companion issue on the misleading KPI).

## Notes / SPEC angle

The config is already correct per `active_states` excluding handoff states (`In Review` is inactive). The gap is structural:
1. A genuinely-blocked agent has **no inactive tracker state** to move the issue to — `In Review` means "done, awaiting human", and there is no `Blocked` state in the active/inactive config — so it can only leave the issue `In Progress`, which guarantees re-dispatch.
2. The worker has **no per-issue forward-progress / cumulative-turn budget across re-dispatches**; `max_turns` caps a single dispatch, not the continuation chain (cf. the `max_retry_attempts` removal in #577 — but that was for `failure` retries, not the continuation loop).

Please confirm the SPEC §16.5 intent against the Elixir reference: does upstream bound continuation for an issue that completes turns but makes no progress, and/or expect a distinct blocked-handoff state? This was observed end-to-end, not theorized.

Full writeup + evidence: `docs/validation/2026-06-03-binary-linear-codex-autonomous-e2e.md` (Finding 1).

hey-yvan:
Closes #621

## Summary
- Add `agent.max_continuation_turns` and carry clean continuation turn counts across continuation dispatch/retry boundaries.
- Enforce D34 as an issue-level clean-turn budget: every fresh/clean-continuation dispatch receives the remaining clean-turn budget, and the codex app-server runner stops cleanly at that boundary so the orchestrator can park the still-active issue in `Blocked` with `method=continuation_budget`.
- Keep failure retry and quota backoff independent from the D34 budget; those paths remain SPEC §8.4 / §16.6 backoff behavior and do not receive the clean-turn budget.
- Preserve SPEC §16.5 self-stop semantics by carrying `IssueLeftActiveSet` from runner → worker → orchestrator; D34 only blocks clean exits that still need another active continuation.
- Document accepted SPEC deviation D34, the #621 / PR #625 upstream reproduction context, blocked-claim terminology, runtime event `continuation_budget_blocked`, and the no-auto-redrive hot-reload semantics.

- [x] Adds one, tracked as a **DEVIATIONS.md row** + an `area:spec-alignment` issue, updated in this PR.

D34 is the explicit accepted deviation for #621. PR #625 reproduced the same clean continuation loop against upstream Symphony behavior, so this PR treats the fix as harness hardening rather than an upstream-alignment patch.

## Size budget (AGENTS.md)
Classify into exactly one (review discipline; the size-gate is human-signed, not CI-blocked):

- [ ] `within budget` — <=12 production files / <=300 production LOC (test files and generated code excluded).
- [x] `size-gated: justified overage` — production diff over budget for correctness / regression / race-safety coverage that cannot be split without losing atomicity. **Needs human size-gate sign-off.**

Production code is 22 files and 459 lines of churn. The overage is the atomic surface needed to thread one accepted deviation through workflow config/defaults/validation, hot reload, dispatch, retry carry, runner clean-budget stopping, finalization, state events, print-config, dashboard/runbooks, and documentation.

## Testing
- [x] `go vet ./...`
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --config=.golangci.yml --issues-exit-code=0` (`0 issues`)
- [x] `gofmt -l $(git ls-files '*.go')`
- [x] `go mod tidy && git diff --exit-code -- go.mod go.sum`
- [x] `git diff --check`
- [x] `npm test -- --run` in `cmd/worker/dashboard`
- [x] `go test -tags e2e -race -timeout 15m ./test/e2e/...`

Additional focused validation:
- `go test ./internal/orchestrator ./internal/worker ./internal/runner ./internal/workflow ./cmd/worker`
- focused red/green tests for workflow config, print-config, orchestrator finalize, dispatch-time hot-reload budget blocking, retry carry, runner clean-budget stopping, failure/quota isolation, and dashboard/runbook state surfaces
- self-stop regression: budget-reaching §16.5 clean exits stay on the continuation/reconcile cleanup path, while active rechecks remain budget-blocked
- mutation-style checks for the `>=` budget boundary and missing-turn-event fallback counting

## Notes
- `agent.max_turns` still caps a single app-server session and is never expanded by `agent.max_continuation_turns`; the continuation budget only tightens fresh/clean-continuation dispatches.


hey-yvan:
@codex review

Issue #595, opened 2026-06-02
security(workspace): clone_url basic-auth leaks via mirror clone/fetch error paths


hey-yvan:
## Summary

The `repo.clone_url` basic-auth userinfo (the agent's push + PR credential) can leak into worker logs / error output on a clone or fetch failure, violating the AGENTS.md secret-masking convention ("Secret-bearing values that arrive via config or CLI (e.g. `clone_url` basic-auth userinfo) must be masked before they reach any log, error string, or state output … Mask clone URLs with `workflow.MaskCloneURL`", earned by #469/PR #483).

Surfaced by Codex review on #591 (the runbook follow-up for #579), which has been caveated to not claim full masking pending this fix.

## Leak vectors (verified)

1. **Error wrapping with the raw clone URL** — `internal/workspace/mirror.go`:
   ```go
   if err := runGit(ctx, filepath.Dir(mirror), "clone", "--bare", cloneURL, mirror); err != nil {
       return "", fmt.Errorf("clone mirror %s: %w", cloneURL, err)
   }
   ```
   `cloneURL` carries the embedded `user:token@` userinfo and is interpolated verbatim into the returned error string, which propagates up to worker logs / task error output. (The sibling wraps at the same site use `mirror` (a path) or carry no URL, so this is the one credentialed `%s`.)

2. **Verbatim git stderr forwarding** — `internal/workspace/manager.go`:
   ```go
   func runGit(ctx context.Context, dir string, args ...string) error {
       cmd := exec.CommandContext(ctx, "git", args...)
       cmd.Dir = dir
       cmd.Stdout = os.Stdout
       cmd.Stderr = os.Stderr
       return cmd.Run()
   }
   ```
   A failed `git clone --bare <cloneURL>` / `git fetch` can print the remote URL (with userinfo, depending on git version) on stderr, which is forwarded straight to the worker's `os.Stderr`.

## Acceptance criteria

- [ ] `internal/workspace/mirror.go` clone/fetch error paths mask the clone URL with `workflow.MaskCloneURL` before it enters any error string (fix vector 1 — apply at every site that interpolates a clone URL).
[...]

PR #691, opened 2026-06-06
security: contain git mirror paths against clone URL traversal (#665)

### LINKED ISSUE #665: security: harden git mirror path generation against clone URL path traversal
## Summary

`internal/workspace/mirror.go` maps a clone URL into a deterministic local
bare-mirror path. The component sanitizer (`sanitizeMirrorComponent`) only
rewrites `/` and spaces; it does **not** reject or rewrite path-special
components such as `.`, `..`, or backslashes. As a result a malformed clone
URL path can preserve `..` components through `mirrorPathFor`, and
`filepath.Join(root, host, repoPath+".git")` then normalizes the result
*outside* the intended mirror root.

Worked example for `https://host/../../evil.git`:
- `splitCloneURL` → path `../../evil.git`
- `strings.TrimSuffix(..., ".git")` → `../../evil`
- split + `sanitizeMirrorComponent` keeps `["..", "..", "evil"]` (`..` is
  non-empty and contains no `/` or space, so it passes through unchanged)
- `filepath.Join(root, host, "../../evil.git")` collapses the `..` and escapes
  `root`.

## Trust model (why this is P2, not P1)

`clone_url` normally arrives from operator-controlled workflow config, **not**
from an attacker-controlled channel, so this is defense-in-depth path
containment hardening rather than an externally-triggerable vulnerability. It
is still worth fixing: workspace/mirror code writes to the filesystem from a
derived path and should guarantee containment under the configured mirror root
regardless of input shape. This also matches the repo's existing posture of
masking/validating `clone_url`-derived values (`workflow.MaskCloneURL`, #595).

## Current code

- `internal/workspace/mirror.go`
  - `mirrorPathFor` (`filepath.Join(root, host, repoPath+".git")`)
  - `splitCloneURL`
  - `sanitizeMirrorComponent` (only rewrites `/` and spaces)

## Proposed fix

- Strengthen component sanitization:
  - map `""`, `"."`, `".."` to a safe fallback (e.g. `"unknown"`);
  - rewrite path separators, backslashes, and control characters;
  - reuse the existing workspace component sanitizer if it already covers this.
- Add a final root-containment check before returning:
  - compute absolute root and mirror path;
  - verify `filepath.Rel(rootAbs, mirrorAbs)` is neither `..` nor prefixed by
    `".."+string(filepath.Separator)`, and is not absolute;
  - on violation, fall back to a contained deterministic path.
- Preserve the deterministic host/owner/repo layout for normal URLs.

## Acceptance criteria

- Regression tests for: `https://host/../../evil.git`, `git@host:../../evil.git`,
  `https://host/a/../evil.git`, `https://../owner/repo.git`,
  Windows-style `..\evil.git`, and empty/malformed clone URL fallback.
- Tests assert the returned path is always contained under the mirror root.
- No behavior change for valid `https://host/owner/repo.git` and
  `git@host:owner/repo.git` — paths stay stable and readable.

## Related

- #595 (clone_url redaction/masking), `workflow.MaskCloneURL`.

hey-yvan:
Closes #665

## Summary
- `mirrorPathFor` mapped a clone URL into a local bare-mirror path with `sanitizeMirrorComponent`, which only rewrote `/` and spaces — so `.`/`..`/`\`/control chars survived and `filepath.Join(root, host, repoPath+".git")` could normalize a path *outside* the mirror root (e.g. `https://host/../../evil.git`). `clone_url` is operator-controlled config, so this is **defense-in-depth containment hardening** (matching #595's `clone_url` masking posture), not an externally-triggerable escape.
- **Layer 1 — stronger sanitizer:** replace the weak duplicate `sanitizeMirrorComponent` with the existing shared `sanitizeComponent` (maps `""`/`"."`/`".."`→`"unknown"`, rewrites every char outside `[A-Za-z0-9._-]`). Removes a duplicate sanitizer (one source of truth) and neutralizes traversal at the segment level. Valid URLs are unaffected — `gitea.example.com`/`acme`/`demo` sanitize to themselves, so the host/owner/repo layout stays stable and readable.
- **Layer 2 — containment backstop:** `pathContainedUnder` (`filepath.Abs`+`filepath.Rel`, **not** a string prefix, so a `<root>-evil` sibling is correctly rejected) routes any escaping candidate to a deterministic URL-keyed fallback under root — guaranteeing mirror writes stay under the configured root regardless of input shape, even if sanitization is later weakened or `splitCloneURL` grows a new parsing branch.

## Testing notes
- Regression tests for every acceptance-criteria URL (https/scp `..`, mid-path `..`, authority `..`, windows `..\`, deep `../../etc/passwd`, empty/whitespace/garbage), asserting **exact structured paths** + an independent `filepath.Rel` containment cross-check. Exact-path assertions matter: with the backstop present, a containment-only test would still pass if layer-1 sanitization were reverted (the backstop silently catches it). Pinning the structured path makes layer 1 mutation-detectable.
- Both layers were **mutation-verified against the committed artifact**: reverting per-segment sanitization fails the exact-path test; making the backstop a no-op fails the fallback test.

## Adversarial review
Ran a 4-lens adversarial-review workflow (security / spec-alignment / go-correctness / test-placebo) with per-finding verification — substituting for origin `@codex review`, currently unavailable. 9 findings, **2 confirmed (both LOW)**, folded into this commit:
- **Removed** the now-unreachable `if repoPath == ""` branch (`sanitizeComponent` maps `""`→`"unknown"`, so `repoPath` is always non-empty; empty clone URLs are handled in `splitCloneURL`). Clean-code rule 5.
- **Corrected the `mirrorPathFor` doc comment**: the 120-rune-capped sanitizer is *best-effort* de-confliction (consistent with every other `SanitizeComponent`-derived path in the repo), not a "two repos never collide" guarantee. The only residual collision is two distinct **>120-char** segments sharing a mirror dir — cache confusion, **not** a traversal escape, unreachable with real forge name limits (≤100 chars), and the same trade-off the repo already accepts for issue-id workspace paths (documented in `workspace-cache.md`). A hash-disambiguator only for mirrors would be unearned scaffolding inconsistent with repo norms, so it was deliberately not added.
- **Pinned intentional non-hostile behavior** with a new test: `host:port`→`host_port` (colon sanitized — Windows-safe), dots/uppercase in owner/repo preserved, empty path components → `unknown`.

The verifier independently confirmed containment holds for every hostile input tried (percent-encoded `%2e%2e`, `..%2f`, embedded `/etc/passwd`, UNC `\\server\share`, NUL/control bytes, `file://`, deep `../../`, relative roots); the URL-keyed fallback can itself never escape (separators are stripped into a single child segment).

Internal workspace path-derivation hardening: no SPEC config key, no worker/orchestrator phase, no tracker-write / post-push gate. Stays strictly inside the Symphony scheduler/runner boundary; `internal/workspace/mirror.go` is not a SPEC-sensitive path.

- [x] `within budget` — 1 production file, ~24 net production LOC (test files excluded).

## Testing
- [x] `go vet ./...`

### DIFF
--- a/internal/workspace/mirror.go
+++ b/internal/workspace/mirror.go
@@ -78,23 +78,76 @@ func MirrorRoot(override string) string {
 
 // mirrorPathFor maps a clone URL to a stable on-disk location under root.
 // We do not need cryptographic uniqueness, just a deterministic, filesystem-
-// safe path keyed by host + path so two repos never collide. The trailing
-// ".git" is preserved to keep the layout recognisable when inspected by hand.
+// safe path keyed by the sanitized host + path. De-confliction is best-effort
+// and matches how sanitizeComponent is used for every other derived workspace
+// path: the shared sanitizer rewrites unsafe characters and caps each segment,
+// so two repos whose sanitized, length-capped segments coincide would share a
+// mirror directory. That is an accepted trade-off because clone_url is
+// operator-controlled config and real forge host/owner/repo names stay well
+// under the cap — not a "never collides" guarantee. The trailing ".git" is
+// preserved to keep the layout recognisable when inspected by hand.
 func mirrorPathFor(root, cloneURL string) string {
 	host, repoPath := splitCloneURL(cloneURL)
-	host = sanitizeMirrorComponent(host)
+	host = sanitizeComponent(host)
 	repoPath = strings.TrimSuffix(repoPath, ".git")
 	// Preserve owner/name structure so multiple repos under the same owner
-	// share a parent directory, which makes manual cleanup easier.
+	// share a parent directory, which makes manual cleanup easier. Each segment
+	// goes through the shared workspace component sanitizer, which maps
+	// ""/"."/".." to "unknown" and rewrites separators, backslashes, and control
+	// characters — so no segment can smuggle a path-traversal hop into the
+	// joined path (#665).
 	parts := strings.Split(repoPath, "/")
 	for i, p := range parts {
-		parts[i] = sanitizeMirrorComponent(p)
+		parts[i] = sanitizeComponent(p)
 	}
+	// strings.Split never yields an empty slice and sanitizeComponent maps the
+	// empty string to "unknown", so repoPath is always non-empty here; no
+	// empty-string fallback is needed (an empty clone URL is handled upstream in
+	// splitCloneURL, which returns "repo").
 	repoPath = strings.Join(parts, string(filepath.Separator))
-	if repoPath == "" {
-		repoPath = "repo"
+	candidate := filepath.Join(root, host, repoPath+".git")
+	return mirrorPathUnderRoot(root, candidate, cloneURL)
+}
+
+// mirrorPathUnderRoot returns candidate when it is contained under root, and a
+// deterministic, URL-keyed fallback path under root otherwise. The per-segment
+// sanitization in mirrorPathFor already neutralizes "."/".."/separators, so a
+// sanitized candidate cannot climb above root and this guard is a
+// defense-in-depth backstop: it guarantees mirror writes stay under the
+// configured root regardless of input shape, even if the sanitizer is later
+// weakened or splitCloneURL grows a new parsing branch (#665). clone_url is
+// operator-controlled config rather than an attacker channel, so this is
+// hardening, not a fix for an externally-triggerable escape.
+func mirrorPathUnderRoot(root, candidate, cloneURL string) string {
+	if pathContainedUnder(root, candidate) {
+		return candidate
 	}
-	return filepath.Join(root, host, repoPath+".git")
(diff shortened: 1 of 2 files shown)


hey-yvan:
Independent Codex review (origin `@codex` quota unavailable → ran on the durably-authorized bytevane fork, [bytevane/aiops-platform#1](https://github.com/bytevane/aiops-platform/pull/1)) came back **clean**: _"Didn't find any major issues. Breezy!"_

Review convergence on this diff:
- Adversarial workflow (13 agents, 4 lenses: security / spec-alignment / go-correctness / test-placebo) → 2 LOW findings, both fixed (dead branch removed; over-promising collision comment corrected).
- `pr-review-toolkit:code-reviewer` → clean (confirmed `pathContainedUnder`'s error path is correct fail-closed).
- Independent Codex (different model) → clean.
- Both sanitization + containment layers mutation-verified against the committed artifact.

Issue #666, opened 2026-06-05
reliability: avoid per-read goroutine lifecycle issues in Codex app-server stdout reader


hey-yvan:
## Summary

`internal/runner/codex_app_server.go` starts a **new goroutine per scanner
read** inside `readLine` (`go func()` at codex_app_server.go:674, wrapping
`c.scanner.Scan()`). If the caller returns first — context cancellation or the
app-server read timeout (`readLineOnce`) — that goroutine remains blocked inside
`scanner.Scan()` until stdout later produces a line or the process exits.

The send channel is buffered (`make(chan readResult, 1)`), so the goroutine can
send-and-exit even with no reader; the residual concern is the goroutine parked
in `Scan()` after its caller has moved on. The process is usually killed on
timeout, so this is not a guaranteed production leak on every path, but the
lifecycle is hard to reason about and the timeout/cancel paths deserve focused
leak coverage.

This is exactly the failure class AGENTS.md "cross-cutting checklist" item 2
warns about (replicate Elixir's implicit runtime guarantees explicitly in Go;
goroutine lifetimes must be obvious) and the Go Code Review Comments guidance
that goroutine lifetimes be obvious or documented with clear exit conditions.

## Proposed fix

Replace per-read goroutines with a single long-lived reader loop per
app-server process:

- one stdout reader goroutine continuously scans lines and sends `readResult`
  values on a channel;
- request/response waiting (`readLineOnce`) selects on: reader channel,
  context cancellation, read timeout, process exit;
- on timeout/cancel, terminate the app-server or close the pipe so the reader
  goroutine exits deterministically;
- guard the goroutine per the repo's `safeGo`/`recoverPanic` convention.

## Acceptance criteria

- Tests simulate: scanner produces no line and read timeout fires; context
  cancellation while scanner is blocked; app-server exits and reader returns
  EOF; scanner line-too-long still maps to the existing response-error
  behavior.
- A goroutine-leak regression test (or equivalent lifecycle assertion).
- Transcript writing behavior preserved.
- Existing timeout categories (`CategoryResponseError` etc.) and
  operator-facing error messages preserved unless intentionally updated.

## Related

- #661 burn-down comment already flags reader-lifecycle as the priority split
  for this file; this issue tracks the reliability fix itself (file-size
  decomposition stays in #661).

PR #693, opened 2026-06-06
reliability: bound terminal-cleanup state-refresh retry with backoff + cap (#675)

### LINKED ISSUE #675: reliability: bound terminal-cleanup state-refresh retry with backoff and attempt cap
## Summary

The §18.1 terminal-workspace-cleanup path re-checks the issue's current tracker
state before deleting the deterministic workspace. When that state refresh
**fails** (or the issue is absent from the response), `verifyReconciledWorkspaceStillTerminal`
returns `known=false` (`internal/orchestrator/actor_cleanup.go:240-247`), and
`runReconciledWorkspaceCleanup` calls `retryReconciledWorkspaceCleanup`
(actor_cleanup.go:188-190). That retry simply re-arms a fixed-delay timer and
re-runs the whole check:

```go
func (o *Orchestrator) retryReconciledWorkspaceCleanup(w ReconciledWorkspace, continuation *continuationAfterSkippedCleanup) {
	safeGo("orchestrator.reconcile_cleanup_retry", func() {
		timer := time.NewTimer(terminalCleanupStateRetryDelay) // = continuationRetryDelay = 1s
		defer timer.Stop()
		select {
		case <-timer.C:
			o.runReconciledWorkspaceCleanup(w, continuation)
		case <-o.runCtx.Done():
		}
	})
}
```

There is **no attempt counter, no backoff growth, and no give-up bound**. The
only exits are: the state refresh eventually succeeds, the issue gets
re-claimed (`beginReconcileWorkspaceCleanup` returns false), or the process
shuts down (`runCtx.Done()`). While the tracker is unavailable, every terminal
issue awaiting cleanup spins a fresh goroutine + one extra tracker state fetch
**every 1s, forever** — adding load to an already-unhealthy tracker and holding
the workspace path un-reclaimable for the whole outage.

## Why this is a problem

This contradicts the project's own backoff discipline: every other reschedule
path (failure retry, continuation retry) routes through
`RetryScheduler.NextDelay` with an attempt count and exponential backoff
(§8.4/§16.6). The cleanup retry is the one reschedule loop that uses a fixed 1s
delay with no ceiling. Only `flakyStateRefresher` (fails once, then succeeds) is
covered in tests; the persistent-failure path is untested.

## Proposed fix

- Thread an attempt count + exponential backoff through
  `retryReconciledWorkspaceCleanup` (reuse `RetryScheduler.NextDelay`).
- Cap total attempts; on exhaustion, log and stop — the startup reconcile sweep
  re-discovers the orphaned workspace on the next worker start, so giving up is
  safe.
- Stop the per-retry goroutine churn from compounding (one timer per issue, not
  a new goroutine per tick beyond what backoff dictates).

## Acceptance criteria

- A regression test with a resolver that fails persistently asserts the retry
  loop backs off and terminates within a bounded number of attempts (not just
  the existing fail-once `flakyStateRefresher`).
- Backoff schedule matches the failure-retry path.
- Successful-refresh and re-claim exits unchanged.

## Related

- `RetryScheduler.NextDelay`, `continuationRetryDelay` (`actor_retry.go:42`).
- Same backoff-discipline class as #287/#405 (timeout-bounded I/O) and the
  §8.4/§16.6 retry machinery.

hey-yvan:
Closes #675

## Summary
The SPEC §18.1 terminal-workspace-cleanup path rechecks the issue's tracker state before deleting the deterministic workspace. When that refresh **failed** (or the issue was absent), `retryReconciledWorkspaceCleanup` re-armed a **fixed 1s timer** and re-ran forever — no attempt counter, no backoff growth, no give-up bound. While a tracker was unavailable, every terminal issue awaiting cleanup spun a fresh goroutine + one extra state fetch **every 1s indefinitely**, hammering an already-unhealthy tracker and pinning the workspace un-reclaimable for the whole outage. It was the one reschedule loop that bypassed the §8.4/§16.6 backoff discipline every other retry path uses.

**Fix:**
- Thread a 1-based attempt count through the retry. The delay now comes from `RetryScheduler.NextDelay(RetryKindFailure, attempt)` — the **same exponential backoff as the failure-retry path** — instead of a fixed 1s interval.
- Cap retries at `maxReconciledCleanupStateRetries` (8); on exhaustion log `event=reconcile_workspace_cleanup_giveup` and stop. The next worker start's reconcile sweep re-discovers the orphaned workspace, so giving up is safe and bounds both the goroutine churn and the tracker load.
- Split `runReconciledWorkspaceCleanup` into a thin attempt-0 entry + `runReconciledWorkspaceCleanupAttempt(attempt)` so existing call sites are unchanged and only the retry path carries the counter. Removed the now-unused `terminalCleanupStateRetryDelay`.

## Adversarial review
Ran a 4-lens workflow (concurrency / logic-edge / spec-clean / test-placebo) with per-finding verification (origin `@codex` quota down → independent Codex on the bytevane fork). **1 confirmed MEDIUM**, adjudicated and folded in:

> The give-up path can carry a non-nil §16.6 continuation; dropping it resets the D34 clean-turn budget on tracker recovery.

I rejected the finding's *primary* fix (reschedule the continuation on give-up) as **unsafe** — resuming would bypass the deletion-time state recheck this loop could not complete, defeating the §18.1 stale-terminal-deletion guard and the D35 operator-stop fail-closed gate. The drop itself is correct: the D34 budget is process-local/non-durable by design (SPEC §14.3), so re-derivation on the next poll re-dispatch is the same recovery as a worker restart. I applied the finding's **safe** recommendation instead: document the deliberate no-reschedule, and **log the drop** (`continuation_dropped` + `continuation_turns`) so it is no longer silent (AGENTS.md "no silent caps"). (Two other lenses' takes on the same behavior were verifier-dismissed as either an accepted §14.3 semantic or an unsafe fix; this synthesis matches the SPEC-grounded reading.)

## Testing notes
- **Give-up test** (persistent-failure resolver): asserts the loop backs off and terminates at exactly `cap+1` fetches, then stops probing, and never cleans the workspace.
- **Schedule test** (recording scheduler): asserts each retry asks for `RetryKindFailure` attempts `1..cap` (matching the failure-retry path) **and** that the give-up schedules **no** continuation retry — guarding against the unsafe resume. The flow carries a non-nil continuation (verified: the giveup log reports `continuation_dropped=true`).
- The existing fail-once `flakyStateRefresher` test covers the success exit, unchanged.
- **Mutation-verified against the committed artifact:** removing the cap → give-up test fails; reverting `RetryKindFailure`→`RetryKindContinuation` (the original fixed-1s bug) → schedule test fails; adding an unsafe reschedule-on-give-up → the no-continuation guard fails.
- A `-race`-caught bug during development: mutating the package-global cap from the test raced the off-actor retry goroutine reading it → the test reads the production default instead (concurrent reads don't race).

Brings the one outlier reschedule loop onto the existing §8.4/§16.6 backoff seam (`RetryScheduler.NextDelay`); no new config key, no new phase/gate — modifies existing `internal/orchestrator` cleanup files. Reinforces the backoff discipline of #287/#405.

- [x] `within budget` — 2 production files, ~50 net production LOC (test files excluded).

## Testing
- [x] `go vet ./...`

### DIFF
--- a/internal/orchestrator/actor.go
+++ b/internal/orchestrator/actor.go
@@ -104,7 +104,6 @@ type Orchestrator struct {
 
 	state      *OrchestratorState
 	dispatcher Dispatcher
-	scheduler  Scheduler
 	retryWake  chan struct{}
 
 	// workspaceCleaner removes the workspace of a terminal-state run after its
@@ -130,6 +129,15 @@ type Orchestrator struct {
 	retryTerminalResolverMu sync.Mutex
 	retryTerminalResolver   IssueStateRefresher
 	retryTerminalStates     map[string]struct{}
+
+	// scheduler computes retry backoff (SPEC §8.4/§16.6). It is read on every
+	// retry fire — including the off-actor terminal-cleanup state-refresh retry
+	// (#675) — and swapped on-actor by UpdateRetryScheduler when the runtime
+	// poller reloads workflow timing, so a mutex guards the swap (same discipline
+	// as candidateLister / retryTerminalResolver). currentScheduler is the only
+	// read path.
+	schedulerMu sync.Mutex
+	scheduler   Scheduler
 
 	// runCtx is captured by Run so followup goroutines can cancel
 	// their work when the actor stops. Set once at the top of Run
@@ -265,6 +273,15 @@ func (o *Orchestrator) currentRetryTerminalResolver() (IssueStateRefresher, map[
 	o.retryTerminalResolverMu.Lock()
 	defer o.retryTerminalResolverMu.Unlock()
 	return o.retryTerminalResolver, o.retryTerminalStates
+}
+
+// currentScheduler returns the live retry scheduler under schedulerMu so the
+// off-actor terminal-cleanup retry can read it without racing an on-actor
+// UpdateRetryScheduler swap (#675).
+func (o *Orchestrator) currentScheduler() Scheduler {
+	o.schedulerMu.Lock()
+	defer o.schedulerMu.Unlock()
+	return o.scheduler
 }
 
 func (o *Orchestrator) LookupOperatorTerminalStop(ctx context.Context, id IssueID) (OperatorTerminalStopEntry, bool, error) {
--- a/internal/orchestrator/actor_cleanup.go
+++ b/internal/orchestrator/actor_cleanup.go
@@ -171,6 +171,15 @@ func terminalWorkspaceForCleanup(id IssueID, identifier, path, root, state strin
 // hardening). Callers must invoke it from a followup goroutine, never inside
 // an apply.
 func (o *Orchestrator) runReconciledWorkspaceCleanup(w ReconciledWorkspace, continuation *continuationAfterSkippedCleanup) {
+	o.runReconciledWorkspaceCleanupAttempt(w, continuation, 0)
+}
+
+// runReconciledWorkspaceCleanupAttempt is one pass of the deletion-time recheck.
+// attempt is 0 for the initial pass and N for the Nth backed-off retry after a
+// failed or absent tracker state refresh; on an unknown refresh it schedules the
+// next attempt through retryReconciledWorkspaceCleanup (exponential backoff +
+// give-up bound) instead of re-probing on a fixed interval forever (#675).
+func (o *Orchestrator) runReconciledWorkspaceCleanupAttempt(w ReconciledWorkspace, continuation *continuationAfterSkippedCleanup, attempt int) {
 	if o.workspaceCleaner == nil || strings.TrimSpace(w.Path) == "" {
 		return
 	}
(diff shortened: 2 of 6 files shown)


hey-yvan:
Independent Codex review (bytevane fork) flagged a **P2 data race**: this off-actor cleanup retry read `o.scheduler` while `UpdateRetryScheduler` swaps it on the actor goroutine — every prior scheduler read was on-actor, so this path introduced the first off-actor read. **Fixed** by guarding the field with `schedulerMu` and reading via a new `currentScheduler()` accessor (the same discipline `candidateLister` / `retryTerminalResolver` already use for off-actor seams). Added a scheduler-swap race regression test that drives the cleanup loop while continuously swapping the scheduler — it fails under `-race` without the guard (mutation-verified). My Claude reviews missed this; the independent Codex pass caught it (the #666 lesson).

hey-yvan:
Independent Codex re-review (bytevane fork, on the fixed commit `3b8e5ca`) came back **clean**: _"Didn't find any major issues. 🎉"_ — the P2 scheduler race is resolved and nothing new surfaced.

Review convergence:
- Adversarial workflow (4 lenses) → 1 confirmed MEDIUM (continuation-drop), adjudicated: rejected the unsafe reschedule, applied observability (continuation_dropped log).
- `pr-review-toolkit:code-reviewer` → clean.
- Independent Codex → caught a P2 off-actor scheduler data race my Claude reviews missed → fixed (schedulerMu + currentScheduler) + race regression test → re-review clean.
- All behaviors (cap, backoff schedule, no-reschedule guard, scheduler-swap race) mutation-verified against the committed artifact.

PR #694, opened 2026-06-06
security: cover MaskCloneURL + fail closed on unparseable credentialed URLs (#676)

### LINKED ISSUE #676: testing: add direct unit coverage for MaskCloneURL credential masking
## Summary

`workflow.MaskCloneURL` (`internal/workflow/config.go:19`) strips embedded
basic-auth from a clone URL before it reaches a log, error string, or
`--print-config` output. It is the credential-masking primitive guarding
`mirror.go:190`, `print_config.go:263`, and `doctor.go:1040/1061`. It has **no
direct unit test**.

The only test file that names it (`internal/workspace/redact_test.go`) does not
test it as a subject:
- line 44 is a comment referencing it,
- line 145 calls `MaskCloneURL(cloneURL)` only to format the *expected* failure
  message inside `TestEnsureMirror_FailedCloneMasksErrorString` — i.e. the
  expected value is computed by the very function under question, so the test
  passes even if `MaskCloneURL` regresses (both sides change together).

## Why this matters

This is exactly the failure class that earned the secret-masking rule
(#469/#483): a masking helper that is silently under-covered ships a credential
leak on green CI. The current implementation relies on `url.Parse` semantics
(`u.User = nil`); a refactor that, say, returned `u.Redacted()` (which renders
`xxxxx` userinfo) or mishandled the no-password / unparseable-SSH branches would
leak or malform with no failing test.

## Current code

```go
func MaskCloneURL(raw string) string {
	if raw == "" { return raw }
	u, err := url.Parse(raw)
	if err != nil || u.User == nil { return raw } // SSH git@host:path has no userinfo -> unchanged
	u.User = nil
	return u.String()
}
```

## Proposed fix

Add a table-driven test in `internal/workflow` asserting each branch:
- `https://user:token@host/r.git` → `https://host/r.git` (userinfo stripped)
- password containing `@` (`https://u:p@ss@host/r.git`) masked correctly
- bare-username token alias (`https://oauth2@host/r.git`) → stripped
- SSH form `git@host:owner/repo.git` → returned unchanged
- empty string → unchanged
- non-URL / unparseable input → unchanged (no panic)
- assert the result never contains the original userinfo substring

Use `Foo(%q) = %q; want %q` failure messages (clean-code rule 9).

## Related

- #469 / #483 (credential masking under-coverage), #595 (clone_url redaction),
  #665 (mirror path traversal — same `clone_url` trust surface).

hey-yvan:
Closes #676

## Summary
`workflow.MaskCloneURL` strips embedded basic-auth from a clone URL before it reaches a log, error string, or `--print-config` output (`mirror.go`, `print_config.go`, `doctor.go`). It had **no direct unit test** — the only file naming it computed its expected value *with* `MaskCloneURL`, so a regression passed both sides together (the #469/#483 under-coverage class the issue cites).

Adding direct table-driven coverage **exposed a real leak**: `MaskCloneURL` returned the raw string whenever `url.Parse` failed, so a **credentialed URL that does not parse** (a malformed port — `https://user:token@host:notaport/r` — or a space in the userinfo) leaked the token verbatim into the masked output, and thus into the clone-failure error logged at `mirror.go`.

**Fix — fail closed.** Keep `url.Parse` as the primary path (parse-success behavior is byte-identical), but on the `url.Parse` *error* path run a conservative string strip of `scheme://userinfo@` (`stripCloneURLUserinfo`) that mirrors url.Parse's last-`@`-in-authority rule and leaves an `@` in the path/query untouched. The SSH scp form (`git@host:path` — url.Parse rejects it, no `://` authority) and plain paths carry no basic-auth userinfo and stay unchanged.

## Testing notes
- 13-case table asserting exact masked output + a sensitive-substring absence check across every branch: userinfo / password-with-`@` / bare-token stripped, ssh url-form stripped, scp-form unchanged, empty / non-url / no-cred-unparseable unchanged, `@`-in-path not over-stripped, and the fail-closed cases that previously leaked — malformed port, space in userinfo, `@` in the path tail (authority/tail split honored), no-tail (`end<0` branch), and password-with-`@` + bad port (the fallback's **last-@** split).
- **Mutation-verified against the committed artifact:** reverting the fail-closed fallback → the fail-closed cases fail; not stripping on parse-success → the masking cases fail; first-`@` instead of last → the last-@ case fails; dropping the path tail → the `@`-in-path/bad-port cases fail.

## Adversarial review
4-lens workflow (leak-bypass / correctness / spec-clean / test-quality) with per-finding verification — **6 confirmed, all folded in**:
- **MEDIUM (test):** the fallback's authority/tail split was only exercised via parse-success inputs → added the `@`-in-path, no-tail, and last-@ fail-closed cases.
- **MEDIUM (gate):** `config.go` is a SPEC-sensitive path → added **DEVIATIONS.md D36** tracking clone-URL masking as the aiops-platform logging-safety extension it is (previously untracked), and checked the DEVIATIONS option below.
- **LOW (doc):** corrected the `MaskCloneURL` comment — url.Parse *rejects* the scp form (it is not "kept as an opaque path"); the no-`://` fallback returns it unchanged.
- **LOW (class):** kept `workspace.redactCredentials`/`credentialURLRe`'s "agrees with MaskCloneURL" comment honest — it has no fail-closed gap because a free-text URL ends at whitespace (`\s` is a correct token boundary), so its splitting agreement holds.

A leak-bypass claim (scheme-relative `//user:token@host:badport`) was verifier-dismissed: it is not a valid git clone URL (git requires a scheme), so it cannot reach `MaskCloneURL`.

- [x] Adds one, tracked as a **DEVIATIONS.md row** + an `area:spec-alignment` issue, updated in this PR.

The change is to `internal/workflow/config.go` (a SPEC-sensitive path). It adds no Config key — it hardens the existing `MaskCloneURL` credential-masking utility, an aiops-platform extension with no upstream Symphony equivalent. **DEVIATIONS.md D36** now tracks that extension (previously untracked) and records the fail-closed hardening; umbrella spec-alignment tracker [#67](https://github.com/xrf9268-hue/aiops-platform/issues/67).

- [x] `within budget` — 3 production files (config.go, redact.go comment, DEVIATIONS.md), ~53 net production LOC (test files excluded).

## Testing
- [x] `go vet ./...`


hey-yvan:
Independent Codex review (bytevane fork, [bytevane/aiops-platform#4](https://github.com/bytevane/aiops-platform/pull/4)) came back **clean**: _"Didn't find any major issues."_

Review convergence:
- Adversarial workflow (4 lenses) → 6 confirmed, all folded in (the fail-closed coverage gap, the merge-blocking metadata gate → D36, the scp-form doc inaccuracy, the redactCredentials cross-reference, plus two test-completeness cases).
- `pr-review-toolkit:code-reviewer` → clean (stress-tested the fallback with malformed IPv6 / percent-escapes / compound schemes — all strip; parse-success byte-identical).
- Independent Codex → clean.
- The fail-closed fix + last-@/tail-split logic mutation-verified against the committed artifact (4 distinct mutations each fail a test).

Issue #661, opened 2026-06-05
refactor: burn down oversized Go file baselines


hey-yvan:
## Context

PR #657 adds a machine-enforced production Go file-size budget for #647, with an explicit grandfathered baseline for existing oversized non-test, non-generated Go files. That is the right ratchet for preventing new regressions, but the baseline itself is still known technical debt and should be paid down separately from the enforcement PR.

This issue tracks the burn-down work only; it should not block #657 unless the implementation of the gate itself changes.

Related:

- #647 — adds the file-size enforcement signal.
- #657 — current PR that introduces the baseline ratchet.
- #660 — audits the policy and docs against official Go guidance.

## Current oversized production Go file baseline from #657

- `internal/doctor/doctor.go` — 1294 lines
- `internal/orchestrator/state.go` — 1020 lines
- `cmd/tui/main.go` — 904 lines
- `internal/workflow/loader.go` — 852 lines
- `internal/runner/codex_app_server.go` — 840 lines

## Acceptance criteria

- Decompose the oversized files in focused, behavior-preserving PRs until each either falls under the repository's 800-line production-file budget or has a freshly justified, reviewed exception.
- Remove each file from `scripts/file_size_budget_test.go`'s baseline as it is paid down; the gate must continue to fail for any new unbaselined oversized production Go file.
- Follow Go's official guidance in spirit: split around cohesive responsibilities and readable package boundaries, not arbitrary line-count slicing. Do not introduce package churn, indirection, or public API surface solely to satisfy the budget.
- Preserve existing behavior with characterization/regression tests before moving code, especially for doctor output, orchestrator state transitions, workflow loading, runner protocol payloads, and TUI flows.
- Keep each cleanup PR small enough to review independently; one file or one cohesive subsystem per PR is preferred.
- If #660 changes the file-size policy or enforcement strategy, update this issue's baseline/criteria before starting the burn-down.

## Notes

This is intentionally separate from the function-complexity cleanup history: #521 handled the heaviest `gocognit` hotspots and is already closed. This issue is for file-level baseline repayment introduced by #657's ratchet.

hey-yvan:
### Suggested prioritization from follow-up code review

The current oversized-file baseline is accurate under the raw-line metric used by the file-size gate. Do not replace these numbers with nonblank/noncomment LOC from external reviews.

Suggested burn-down order should be risk-weighted, not just largest-file-first:

1. `internal/runner/codex_app_server.go`
[...]

PR #699, opened 2026-06-06
fix: validate Codex app-server JSON-RPC ids before int conversion (#671)

### LINKED ISSUE #671: protocol: validate Codex app-server numeric IDs before converting to int
## Summary

`numberID` (`internal/runner/codex_app_server.go:819`) accepts `float64` values
decoded from JSON and converts them with `int(x)` (codex_app_server.go:822),
silently truncating non-integer values and not guarding NaN/Inf or
out-of-`int`-range magnitudes. JSON-RPC ids from the Codex app-server should be
exact integers; a malformed id should be rejected, not truncated to a
plausible-but-wrong int that then matches the wrong pending request.

## Proposed fix

- For the `float64` case:
  - require a finite value (reject NaN/Inf);
  - require `math.Trunc(x) == x` (reject fractional ids such as `1.5`);
  - require the value fits in `int` (reject out-of-range magnitudes).
- Keep the existing `int` case for tests/internal callers.
- Consider decoding with `json.Decoder.UseNumber()` if broader protocol parsing
  benefits from exact numeric validation.

## Acceptance criteria

- Tests reject `1.5`, very large out-of-`int`-range values, and (if invalid for
  the field) negative ids.
- Tests preserve current valid integer ids.
- The resulting failure stays in error category `CategoryResponseError`.

## Related

- Touches `codex_app_server.go`; coordinate with the #661 decomposition of this
  file but does not depend on it.

hey-yvan:
Closes #671

## Summary
- `numberID` (the JSON-RPC id extractor for the Codex app-server) converted a
  JSON-decoded `float64` id with `int(x)`, **silently truncating** non-integers
  (`1.5` → `1`) and not guarding `NaN`/`Inf` or out-of-`int`-range magnitudes. A
  truncated id could match the **wrong pending request** in `responseForID`.
- **Fix (`numberID`):** for the `float64` case, reject `NaN`/`Inf`, fractional
  values (`math.Trunc(x) != x`), and out-of-`int`-range magnitudes; otherwise
  return the exact `int`. The range bounds account for float rounding
  (`float64(math.MaxInt)` rounds up to 2^63 → strict upper bound; `math.MinInt`
  is exactly representable → inclusive lower bound). The `int` case is unchanged.
- **Fix (`responseForID`):** a **present-but-non-integer** id is now surfaced as
  a `CategoryResponseError` (we can't reliably tell which request it answers, so
  reject rather than skip-and-wait — also removes a latent hang where the old
  skip looped to EOF). An **absent** id stays an interleaved notification (skip),
  so notification handling is unchanged.

### Scope: why these helpers are safe to harden
`numberID`/`responseForID` are used **only** by the synchronous
`(*appServerClient).request()` loop to match a client→server **response** by id.
Server→client requests (approvals) flow through a **separate** turn loop
(`handleServerRequest`, `codex_app_server_turn.go`) that echoes `msg["id"]`
directly and never calls `numberID`. The client only ever **sends** non-negative
integer ids, so negative integral ids remain valid JSON-RPC ids (they simply
mismatch and skip) rather than being rejected.

### File-size: extraction, not growth
The fix would have grown the already-oversized `codex_app_server.go` (over the
800-line budget). Per AGENTS rule 7 (decompose by responsibility; a reduction
must lower the baseline), I **extracted** the response-matching helpers
(`responseForID` + `numberID`) into a new `internal/runner/codex_app_server_response.go`,
lowering `codex_app_server.go`'s baseline **907 → 877** — a first step toward the
#661/#499 response-matching decomposition the budget comment cites.

- [ ] Adds one, tracked as a **DEVIATIONS.md row** + an `area:spec-alignment` issue.

Sub-SPEC hardening of the Codex runner path. The new file is `internal/runner/*`
(not `internal/orchestrator|worker`), so the PR-metadata SPEC gate does not apply;
no Codex turn/start payload changed (cross-cutting item 3 not engaged).

- [x] `within budget` — production change is a `codex_app_server.go` → `codex_app_server_response.go` move plus ~10 LOC of validation; the test additions are excluded.

## Testing — head `122cd84`
- [x] `gofmt -l` clean; `go vet ./...` clean; `go mod tidy` clean; golangci-lint `0 issues` on `internal/runner/...`
- [x] `go test -run '^TestProductionGoFilesStayWithinSizeBudget$' -count=1 ./scripts` (baseline ratchet 907→877)
- [x] `go test -race -covermode=atomic ./...`; python unittests; `go build ./cmd/worker ./cmd/tui`
- [x] `TestNumberID` (table) — rejects `1.5`, `NaN`, `±Inf`, `1e19`; preserves `0`/`7`/`-3`/`int`. **Mutation-verified.**
- [x] `TestResponseForID` (table) — matching id → result / `CategoryResponseError`; mismatched-int & absent id → skip; present non-integer id → `CategoryResponseError`. **Mutation-verified.**
- [x] `TestRequest_FractionalIDRejectedNotTruncated` — the exact bug: id `0.5` (old code truncated → matched request `0`) now → `CategoryResponseError`. **Mutation-verified.**
- [x] `TestRequest_OutOfRangeIDReturnsResponseError`. Mutations restored via `git checkout`, tree==HEAD.

## Review — independent triple review, MERGE-READY
- **4-lens adversarial workflow** (Claude-family): numeric-correctness, protocol-semantics, scope/decomposition, test-quality — all MERGE-READY, `confirmedActionable: []`.
- **`codex:codex-rescue`** (Codex-family): MERGE-READY — no defects; confirmed range/rounding edges, no hang risk, clean extraction, mutation-verified non-placebo tests.
- Both flagged the same **optional** LOW: add direct `responseForID` unit tests for the unchanged skip branches → done (`TestResponseForID`).
- **Fresh `codex:codex-rescue` on the pushed head `122cd84`: MERGE-READY**, all NIT, no defects (numeric edges, no-hang, server-request isolation, baseline, SPEC all confirmed). (An earlier confirmation pass returned a *procedural* NEEDS-CHANGES from a wrong premise — it diffed vs main and re-flagged the production change as unreviewed; the production diff is byte-identical to the prior MERGE-READY head `2812b30`, and this fresh full-diff pass supersedes it.)
- `@codex` GitHub bot is externally unavailable batch-wide (origin "unknown error"; bytevane fork has no Codex env); merging on the compensating independent triple review per protocol §4 (operator-authorized this batch).

### Review threads
GraphQL `reviewThreads` count = 0.

### DIFF
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -621,26 +621,6 @@ func (c *appServerClient) request(ctx context.Context, method string, params any
 		c.handleNotification(msg)
 	}
 }
-
-// responseForID reports whether msg is the JSON-RPC response to request id. When
-// it is (matched=true), it returns the result map — defaulting a missing or null
-// result to an empty map — or a CategoryResponseError carrying the `error`
-// member. A non-matching msg (matched=false) is an interleaved notification the
-// caller must dispatch and keep reading past.
-func responseForID(msg map[string]any, id int) (result map[string]any, matched bool, err error) {
-	gotID, ok := numberID(msg["id"])
-	if !ok || gotID != id {
-		return nil, false, nil
-	}
-	if e, ok := msg["error"]; ok {
-		return nil, true, NewError(CategoryResponseError, fmt.Sprintf("rpc error: %v", e), nil)
-	}
-	result, _ = msg["result"].(map[string]any)
-	if result == nil {
-		result = map[string]any{}
-	}
-	return result, true, nil
-}
 func (c *appServerClient) notify(method string, params map[string]any) error {
 	return c.send(map[string]any{"jsonrpc": "2.0", "method": method, "params": params})
 }
@@ -882,16 +862,6 @@ func extractString(m map[string]any, outer, inner string) (string, error) {
 		return "", NewError(CategoryResponseError, fmt.Sprintf("missing %s.%s", outer, inner), nil)
 	}
 	return v, nil
-}
-func numberID(v any) (int, bool) {
-	switch x := v.(type) {
-	case float64:
-		return int(x), true
-	case int:
-		return x, true
-	default:
-		return 0, false
-	}
 }
 func writeAppServerArtifact(workdir string, buf *cappedWriter) {
 	dir := filepath.Join(workdir, ".aiops")
--- a/internal/runner/codex_app_server_request_test.go
+++ b/internal/runner/codex_app_server_request_test.go
@@ -9,6 +9,7 @@ package runner
 
 import (
 	"context"
+	"math"
 	"testing"
 )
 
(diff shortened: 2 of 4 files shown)


hey-yvan:
@codex review

Issue #672, opened 2026-06-05
performance: reduce N+1 Linear blocker lookups during polling


hey-yvan:
## Summary

Linear polling maps issues, then fetches blockers per Todo-state issue:
`mapLinearIssueNode` (`internal/tracker/linear.go:287`) calls
`linearBlockersForIssue` (linear.go:392) for each Todo issue individually. This
is simple and correct, but becomes N+1 GraphQL calls when many Todo issues are
active on each poll cycle.

## Proposed options (pick what keeps pagination/rate-limit behavior safe)

- Fetch blocker (inverse-relation) data in the main `ListIssues` GraphQL query
  when practical.
- Batch blocker queries by issue id.
- Only resolve blockers for the candidate set that may actually be dispatched.
- Cache blocker info across a single poll cycle.

## Acceptance criteria

- Tests with multiple Todo issues + blockers.
- Existing blocked/ready semantics preserved (including pagination of
  `inverseRelations`).
- A mocked-call-count regression (or benchmark) showing fewer Linear calls for
  many Todo issues.
- Rate-limit / pagination behavior stays safe.

PR #700, opened 2026-06-06
perf: cache Gitea blocker lookups per poll tick (#677)

### LINKED ISSUE #677: performance: reduce N+1 Gitea blocker lookups during polling
## Summary

The Gitea analogue of #672 (Linear N+1 blocker lookups). When resolving the
§8.2 Todo-blocker rule, `buildBlockedBy` (`internal/gitea/tracker_client.go:546`)
iterates the distinct `Depends on #N` refs in an issue body and calls
`getIssueByNumber` (tracker_client.go:417) **once per ref**, each a fresh
`GET /api/v1/repos/{owner}/{repo}/issues/{n}` with no caching of the issue
**body**:

```go
for _, m := range matches {
	...
	issue, found, err := c.getIssueByNumber(ctx, n) // fresh HTTP GET, no body cache
	...
}
```

`getIssueByNumber` consults no cache, and the existing `cacheIssueNumber` only
memoizes issue *numbers*, not fetched issues. So within a single poll tick:

- blocker fetches are O(distinct refs) **per source issue** (the code comment
  acknowledges this), and
- a blocker referenced by two different source issues is fetched **twice** in
  the same tick.

The poll loop is single-threaded, so each serial round-trip directly delays
dispatch of newly-eligible issues and adds load under Gitea rate limits.

## Proposed fix

- Add a per-poll-tick memoization cache keyed by issue number in front of
  `getIssueByNumber`, so each blocker is fetched at most once per tick across
  all source issues.
- Invalidate/rebuild the cache each tick so state stays fresh between ticks.
- Keep the best-effort skip-on-error semantics unchanged.

## Acceptance criteria

- A mocked-call-count regression: N source issues sharing a blocker fetch that
  blocker once per tick, not N times.
- Existing blocked/ready semantics and best-effort error skipping preserved.
- Cache does not leak state across ticks (a blocker that changes state between
  ticks is re-read).

## Related

- Sibling of #672 (Linear N+1, `linear.go:287/392`). Same performance class,
  different client/code path; track separately so each keeps its own
  pagination/rate-limit constraints.

hey-yvan:
Closes #677

## Summary
- `buildBlockedBy` (`internal/gitea/tracker_client.go`) resolved each
  `Depends on #N` reference by calling `getIssueByNumber` **once per distinct ref
  per source issue with no caching**. So within a single poll tick a blocker
  referenced by multiple source issues was fetched once **per source issue**
  (O(distinct refs) per source issue). Each serial `GET /repos/{o}/{r}/issues/{n}`
  delays dispatch of newly-eligible issues and adds Gitea rate-limit load.
- **Fix:** install a **per-poll-tick blocker cache** — created once at the top of
  `ListIssuesByStates` (the per-tick entrypoint) and carried on the call
  `context` (via a typed `giteaTrackerContextKey`, matching the
  `internal/runner/tools.go` precedent) — and consume it in `buildBlockedBy` via a
  new `cachedIssueByNumber` helper. Each distinct blocker is now fetched **at most
  once per tick** across all source issues.
- The cache lives only for the tick's `context` (re-created each call), so a
  blocker's state is **re-read on the next tick** (no cross-tick staleness).

### Best-effort semantics preserved (the key design choice)
`cachedIssueByNumber` caches only **definitive** results — a successful fetch or a
404 not-found (both `err == nil`). A **transient** fetch error (`5xx`/network) is
**not** cached, so a later source issue referencing the same blocker can still
retry it within the tick. This preserves the original `err != nil || !found →
skip` semantics exactly, and avoids the dangerous failure mode of caching a
transient error and **dropping a still-valid blocker** for the rest of the tick
(which could let a blocked issue dispatch).

### Scope / adjacency
`getIssueByNumber` is also called by `FetchIssueStatesByRefs` (the reconcile
entrypoint), which does **not** install the cache and does not call
`buildBlockedBy`, so it is unaffected (falls back to a direct fetch). Sibling of
#672 (Linear N+1), tracked separately per the issue.

- [ ] Adds one, tracked as a **DEVIATIONS.md row** + an `area:spec-alignment` issue.

Scheduler/runner-internal read-path memoization in the tracker reader (deduping
redundant GETs within one poll tick) — squarely the scheduler's job, invisible to
the agent, no SPEC boundary concern; no SPEC-sensitive path touched.

- [x] `within budget` — production diff is `internal/gitea/tracker_client.go` (~+62 LOC); test additions excluded.

## Testing — head `79ec47d`
- [x] `gofmt -l` clean; `go vet ./...` clean; `go mod tidy` clean; golangci-lint `0 issues` on `internal/gitea/...`
- [x] `go test -run '^TestProductionGoFilesStayWithinSizeBudget$' -count=1 ./scripts`; `go test -race -covermode=atomic ./...`; python unittests; `go build ./cmd/worker ./cmd/tui`
- [x] **AC1 call-count:** `…FetchesSharedBlockerOncePerTick` — 2 source issues sharing blocker #3 fetch it **once**. **Mutation-verified** (bypass cache → 2 fetches → FAIL).
- [x] **AC3 no cross-tick leak:** `…RereadsBlockerOnNextTick` — blocker fetched twice across two ticks and its `Done` transition is re-read on tick 2.
- [x] **AC2 best-effort error skip:** `…RetriesBlockerAfterTransientError` — 500-then-200 → fetched twice, second source issue retries and gets the blocker. **Mutation-verified** (cache errors → 1 fetch, blocker dropped → FAIL).
- [x] `TestCachedIssueByNumberNilCacheFallsBackToDirectFetch` — defensive nil-cache path does a direct fetch (no memoization, no nil-map panic).

## Review — independent triple review, MERGE-READY
- **4-lens adversarial workflow** (Claude-family): correctness/lifecycle, error-semantics, design/idiom, test-quality — all MERGE-READY, `confirmedActionable: []`.
- **`codex:codex-rescue`** (Codex-family): MERGE-READY — no defects across lifecycle, concurrency, error-handling, design, and test axes.
- Addressed the two concrete NITs: fixed a stale test-helper doc-comment name (clean-code rule 5) and added the nil-cache fallback test. (Left the `iota` vs `iota+1` key NIT as-is — harmless; distinct typed key, no collision.)
- `@codex` GitHub bot is externally unavailable batch-wide (origin "unknown error"; bytevane fork has no Codex env); merging on the compensating independent review per protocol §4 (operator-authorized this batch).

### Review threads
GraphQL `reviewThreads` count = 0.

### DIFF
--- a/internal/gitea/tracker_client.go
+++ b/internal/gitea/tracker_client.go
@@ -106,6 +106,9 @@ func (c *TrackerClient) ListIssuesByStates(ctx context.Context, states []string)
 	if len(wantedStates) == 0 {
 		return nil, nil
 	}
+	// Install a per-poll-tick blocker cache so buildBlockedBy fetches each
+	// `Depends on #N` blocker at most once per tick across all source issues (#677).
+	ctx = withBlockerCache(ctx)
 	labelNames := StateLabelNamesForStates(states, DefaultStateLabelMappings())
 	issueState := giteaAPIStateForWorkflowStates(states, c.Config.TerminalStates)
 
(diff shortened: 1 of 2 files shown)


hey-yvan:
@codex review

Issue #10, opened 2026-04-28
[M2][P0] Define workflow discovery and fallback behavior


hey-yvan:
## Goal

Make worker behavior deterministic when a repo has or does not have `WORKFLOW.md`.

## Acceptance criteria

- Worker loads repo `WORKFLOW.md` when present.
- Worker uses safe defaults when absent.
- Task event records which workflow file/default was used.
- README documents where `WORKFLOW.md` should live.

## Notes

The repo-owned workflow contract is central to the Symphony-style design.

PR #743, opened 2026-06-11
fix: map Gitea aiops/todo to SPEC state "Todo" so the blocker gate fires

### LINKED ISSUE #739: Gitea tracker: `Depends on #N` blocker gate never fires (gate keys on literal "Todo" state, Gitea maps aiops/todo to "AI Ready")
## Observed (v0.1.0 packaged binary, live lifecycle test 2026-06-11)

With `tracker.kind: gitea` and `active_states: [AI Ready, Rework]`, the worker dispatched issue #10 (body ends with `Depends on #4`) while its blocker #4 was still `aiops/todo` ("AI Ready", non-terminal). Reproduced at 12:06:50 in the test log; the second dispatch slot picked another blocked issue (#11) the same way.

## Root cause

The blocker gate applies only to issues whose normalized state is literally `"Todo"` — pinned by `internal/orchestrator/poller_test.go::TestPollOnceIgnoresBlockersForNonTodoStates` ("want non-Todo issue dispatched despite blockers"). Gitea's `DefaultStateLabelMappings()` maps the pre-dispatch label `aiops/todo` to state **"AI Ready"** — never "Todo" — so on Gitea the gate is structurally dead. `internal/gitea/tracker_client.go::buildBlockedBy` (SPEC §11.3 parsing, plus the #677 per-tick blocker cache) produces `BlockedBy` data that the poller then ignores for every Gitea issue.

This is AGENTS.md cross-cutting checklist item 1: the Gitea adapter (producer) and the poller gate (consumer) of the same SPEC concept diverged on an adjacent path.

## Suggested direction

Check SPEC §11.3 / the Elixir reference for whether the blocker rule keys on the literal "Todo" state name or on "not-yet-started" semantics. If the latter, the gate should apply to any state in `active_states` that precedes dispatch (for Gitea: "AI Ready"), or key off the tracker adapter's own notion of the initial state. Either fix the gate or delete the dead `buildBlockedBy` path for Gitea (one source of truth; no dead producers).

## Workaround used in the field

Operator-side label gating: only label an issue `aiops/todo` once all its dependencies are `aiops/done`.

Full test report: `.trellis/tasks/06-11-web-todo-app-full-lifecycle-test-with-packaged-release-binary/report.md` (F1).

hey-yvan:
Closes #739

## Summary
- On Gitea trackers the SPEC §8.2 blocker rule ("If the issue state is `Todo`, do not dispatch when any blocker is non-terminal") was structurally dead: the Gitea adapter mapped its pre-dispatch label `aiops/todo` to the invented state name **"AI Ready"**, while the poller gate `todoIssueBlockedByOpenDependency` (`internal/orchestrator/poller.go:487`) keys on the literal `"Todo"` — a faithful port of upstream `elixir/lib/symphony_elixir/orchestrator.ex:864` (`todo_issue_blocked_by_non_terminal?`, which compares `normalize_issue_state(issue_state) == "todo"`). Field repro: v0.1.0 lifecycle test dispatched issue #10 (`Depends on #4`) while blocker #4 was non-terminal.
- Root-cause fix is producer-side, one line: `DefaultStateLabelMappings()` now maps `aiops/todo → "Todo"` (SPEC vocabulary, clean-code rule 4), with the constraint documented at the mapping. The gate is untouched — it already matches SPEC §8.2 and upstream. No new "adapter initial state" concept was added (upstream has no equivalent; principle 6).
- `buildBlockedBy` (§11.3) keeps its live consumer instead of being deleted; one source of truth restored across producer/consumer (cross-cutting checklist item 1).
- Atomic rename swept across tests, e2e fixtures, examples, README/runbook tables, per clean-code rule 3/10. Historical records (`DECISION.md` "Was" column, `docs/research/`, `docs/superpowers/plans/`, Trellis archives) intentionally untouched.
- Side benefit: Gitea now works under the SPEC §6.4 default `active_states: [Todo, In Progress]` without an operator override.
- New seam regression test `internal/orchestrator/poller_gitea_seam_test.go` pins the producer/consumer vocabulary: states derived from `gitea.DefaultStateLabelMappings()` are fed through `filterEligibleCandidates`, reproducing the field scenario (#10 blocked by #4). Mutation-verified: reverting the mapping to "AI Ready" fails this test and the gitea package tests.

- [x] Adds one, justified by an upstream **Elixir reference** — cite the matching file and line from the openai/symphony Elixir tree in the body below (a bare module name does not count; the gate looks for a concrete source path).

No new key/phase/gate is added; `internal/workflow/config.go` is touched only to update a stale comment that still named "AI Ready" as example vocabulary. The change itself is dictated by the upstream reference: `elixir/lib/symphony_elixir/orchestrator.ex:864` (`todo_issue_blocked_by_non_terminal?` keys on the literal normalized `"todo"` state), which is why the Gitea pre-dispatch state must be named `Todo`, and by SPEC §8.2 / §6.4 (default `active_states` `["Todo", "In Progress"]`).

Production diff: `internal/gitea/label_state.go` (1 mapping line + why-comment) and `internal/workflow/config.go` (comment only). Everything else is tests, e2e fixtures, examples, and docs.

## Testing
- [x] `go vet ./...`
- [x] `gofmt -l` clean + blocking golangci gate (`staticcheck,unparam,unused`) on changed packages: 0 issues

Additional validation:
- Mutation verification: with the mapping reverted to "AI Ready", `TestGiteaPreDispatchStateIsBlockerGated` fails (`want 0, got 1 dispatched`) and `internal/gitea` package tests fail; restored fix is green.
- `go vet -tags e2e ./test/e2e/...` compiles; the Docker-backed e2e suite is validated by the CI `e2e` job (mechanical state-string updates only).
- `go build ./cmd/worker ./cmd/tui` clean; `go mod tidy` no-op.

## Notes
**Pre-push dual review (protocol §3), head `eacf6d3`:**
- Claude family (`feature-dev:code-reviewer` on b7791fc, then hardened `claude -p` diff-only on eacf6d3): the one MEDIUM (README/local-dev Gitea label tables missing the live `Human Review` / `aiops/human-review` row) was validated and fixed in the amended head. Remaining finding is one LOW, recorded below.
- Codex family: the `codex:codex-rescue` subagent run died mid-flight (rollout stalled, process gone, no job record) — fell back per protocol to `codex exec --sandbox read-only --ephemeral --output-schema` with the diff on stdin; result `{"blocking_findings":[]}` on both b7791fc and the amended head eacf6d3.

**Recorded LOW (no change, rationale):**
- *Claude LOW:* deployments with a stale `active_states: ["AI Ready"]` override now silently match no Gitea label (zero dispatch) and there is no startup diagnostic for unmappable active states. Accepted: pre-release (no users to migrate; the only field deployment is our own lifecycle test), upstream/SPEC have no active_states validation equivalent (principle 6 — upstream absence is an over-design signal), and `go run ./cmd/worker --print-config` already surfaces the effective states for self-diagnosis.
- *Claude LOW (seam test breadth):* the seam test pins vocabulary with a single `Done` terminal state; gate interaction with the SPEC default 5-state terminal set is already covered by `TestPollOnceTreatsDefaultSpecTerminalBlockersAsUnblocked` and `TestPollOnceTodoBlockerHonorsOperatorConfiguredTerminalStates`.

**Acceptance vs issue #739 suggested direction:** SPEC §8.2 keys the rule on the literal `Todo` state name (not "not-yet-started" semantics) — confirmed against upstream; therefore the fix aligns the tracker adapter's pre-dispatch state name instead of changing the gate, and the `buildBlockedBy` producer stays (no dead path remains).

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,8 +367,9 @@ label:
 
 | Workflow state | Gitea label |
 | --- | --- |
-| `AI Ready` | `aiops/todo` |
+| `Todo` | `aiops/todo` |
 | `In Progress` | `aiops/in-progress` |
+| `Human Review` | `aiops/human-review` |
 | `Rework` | `aiops/rework` |
 | `Done` | `aiops/done` |
 | `Canceled` | `aiops/canceled` |
--- a/cmd/worker/main_test.go
+++ b/cmd/worker/main_test.go
@@ -1712,7 +1712,7 @@ func TestWorkerEntrypointDoesNotRequirePostgresQueue(t *testing.T) {
 func TestLoadWorkflowForStartupReconcileUsesConfiguredWorkflowPath(t *testing.T) {
 	dir := t.TempDir()
 	workflowPath := filepath.Join(dir, "linear-workflow.md")
-	body := "---\nrepo:\n  owner: o\n  name: r\n  clone_url: git@example.com:o/r.git\ntracker:\n  kind: linear\n  project_slug: platform\n  active_states: [\"AI Ready\"]\n  terminal_states: [\"Done\"]\n---\nprompt\n"
+	body := "---\nrepo:\n  owner: o\n  name: r\n  clone_url: git@example.com:o/r.git\ntracker:\n  kind: linear\n  project_slug: platform\n  active_states: [\"Todo\"]\n  terminal_states: [\"Done\"]\n---\nprompt\n"
 	if err := os.WriteFile(workflowPath, []byte(body), 0o644); err != nil {
 		t.Fatalf("write workflow: %v", err)
 	}
@@ -2041,14 +2041,14 @@ func TestWorkerReconciliationConfigIncludesInactiveStates(t *testing.T) {
 	cfg := workflow.DefaultConfig()
 	cfg.Repo.CloneURL = "git@example.com:o/r.git"
 	cfg.Tracker.Kind = "linear"
-	cfg.Tracker.ActiveStates = []string{"AI Ready", "In Progress", "Rework"}
+	cfg.Tracker.ActiveStates = []string{"Todo", "In Progress", "Rework"}
 	cfg.Tracker.TerminalStates = []string{"Done", "Canceled"}
 
 	reconcile := reconciliationConfigForWorkflow(cfg)
 	if len(reconcile.InactiveStates) == 0 {
 		t.Fatalf("inactive reconciliation states = %v, want non-empty states for explicit inactive tracker observations", reconcile.InactiveStates)
 	}
-	if containsState(reconcile.InactiveStates, "AI Ready") || containsState(reconcile.InactiveStates, "In Progress") || containsState(reconcile.InactiveStates, "Rework") {
+	if containsState(reconcile.InactiveStates, "Todo") || containsState(reconcile.InactiveStates, "In Progress") || containsState(reconcile.InactiveStates, "Rework") {
 		t.Fatalf("inactive reconciliation states = %v, must not include configured active states", reconcile.InactiveStates)
 	}
 	if containsState(reconcile.InactiveStates, "Done") || containsState(reconcile.InactiveStates, "Canceled") {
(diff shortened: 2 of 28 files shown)


hey-yvan:
@codex review

PR #766, opened 2026-06-12
test(orchestrator): replace fixed-sleep synchronization with deterministic waits

### LINKED ISSUE #760: test-debt: replace fixed time.Sleep synchronization in orchestrator tests with deadline-based polling
## Summary

Orchestrator tests synchronize with async behavior via fixed `time.Sleep` waits: 9 occurrences in `internal/orchestrator/poller_test.go`, 7 in `internal/orchestrator/actor_test.go` (31 across the repo including `test/e2e`). A fixed sleep is the classic flaky-test pattern: too short on a slow CI runner → intermittent failure; padded long → wasted wall time on every run. The suite currently passes under `-race`, but every one of these is a latent intermittent failure waiting for a loaded runner.

## Evidence

```
$ grep -c 'time.Sleep' internal/orchestrator/poller_test.go internal/orchestrator/actor_test.go
internal/orchestrator/poller_test.go:9
internal/orchestrator/actor_test.go:7
```

Pattern: sleep a fixed 1–50ms, then assert dispatcher counts / orchestrator state, hoping the async followup has run.

## Why this matters

- CI runs `go test -race` on shared runners; race-detector overhead is 2–20×, so a 1ms sleep that works locally is exactly the kind of budget that breaks under load.
- Flaky CI erodes the value of the otherwise-strict blocking gate: a red that might be timing noise trains people to re-run instead of investigate.
- The repo already has the right idiom in places (channel/WaitGroup-driven fakes like `blockingDispatcher`); this is about finishing the class (clean-code rule 10), not inventing new machinery.

## Suggested fix

For each sleep, prefer in this order:

1. **Deterministic signal**: have the fake (dispatcher/tracker) close a channel or send on completion, and block the test on it with a `select` + generous timeout. Most sites can use the existing fake types.
2. **Poll-until-deadline**: where a state transition has no observable hook, replace `time.Sleep(N)` + assert with a polling loop bounded by a deadline (e.g. retry the assertion every 5ms for up to ~5s), failing with input/actual/expected per clean-code rule 9.

Do not change production code to accommodate tests unless a genuinely missing observation point is found (then add the smallest hook, e.g. an exported-for-test notification, with justification).

## Acceptance criteria

- [ ] No fixed-sleep synchronization remains in `internal/orchestrator` tests (sleeps that *are* the behavior under test — e.g. real backoff timing assertions — may stay, each with a comment saying why).
- [ ] Suite still passes under `go test -race -count=10 ./internal/orchestrator/` (or similar repetition) to demonstrate stability.
- [ ] No net loss of assertion strength: each converted test still asserts the same observable state.
- [ ] `test/e2e` sleeps reviewed in the same pass; convert where the same idiom applies, leave container-startup waits that poll with deadlines as-is.

## Refs

- `internal/orchestrator/poller_test.go`, `internal/orchestrator/actor_test.go`
- AGENTS.md §Clean code rules 9, 10

_Found during the 2026-06-11 full-project review._

hey-yvan:
Closes #760

## Summary
- Fixed `time.Sleep` synchronization in orchestrator/e2e tests is the classic flaky pattern under `-race` on loaded CI runners (race overhead 2–20×). Each of the 16 flagged sites was judged individually rather than blanket-converted.
- **Converted (5 sites):** three sleeps deleted outright because `PollOnce` dispatches synchronously — `RequestDispatchAfterTrackerRecheck` blocks on its reply channel and the accepted-dispatch followup calls `dispatcher.Spawn` before replying, so spawn counts are final the moment `PollOnce` returns (verified independently by both pre-push reviewers against `actor_dispatch.go`; comments added at each site). The continuation-retry-fire test now blocks on the orchestrator poll-wake channel (`fireWakeSignal` is the only reachable wake producer in that scenario — the single-source assumption is documented at the site). The capacity-respecting retry-fire test waits for the deterministic `capacityDeferRetry` attempt bump, with a timeout message that reports last-observed spawn count + retrying state. The `safeGo` panic-confinement test polls the captured log with a deadline (`captureOrchestratorLog` split into a shared snapshot reader — one source of truth, both callers reuse it).
- **Intentionally kept (4 sites, each commented):** the give-up and sticky-operator-stop tests assert the *absence* of further activity, where a bounded quiescence window is the only possible observation without adding a production seam (the PRD forbids that absent genuine need); the claim-gap race test uses its sleep to inject scheduling pressure, not for synchronization.
- The remaining grep hits live inside the sanctioned `waitFor*` / `waitForE2E` poll-until-deadline helpers and are not fixed-sleep synchronization.
- **Zero production code changes.**

Test-only change; no SPEC-sensitive path touched.

Production diff: **0 files / 0 LOC** (all five changed files are tests).

## Testing
- [x] `go vet ./...` and `go vet -tags e2e ./test/e2e/...`
- [x] `gofmt -l` clean
- [x] **Core deliverable:** `go test -race -count=10 ./internal/orchestrator/` stable at ~26s (run three times across implementation and review-fix rounds); the touched tests additionally stable at `-count=20`.

**Behavior-preservation verification:** with the production retry-fire capacity gate temporarily removed, `TestRetryFire_RespectsCapacityBeforeSpawning` fails; with the continuation fire mutated to spawn instead of wake, `TestContinuationRetryTimerRequiresTrackerRecheckedDispatch` fails — the converted waits still detect the regressions the original sleeps guarded (production file restored byte-identical afterwards).

## Review ledger (pre-push protocol §3–§5)
- Head: `ad450de` (test commit; `bdbe1b3` adds Trellis task state only). CI: **5/5 green** on this head.
- Codex CLI unavailable in this remote container; compensated per protocol §3 with **two independent diff-only Claude-family reviewers**, briefed without the author's conclusions.
- Verdicts: **MERGE-READY ×2.** Both independently verified the load-bearing "PollOnce dispatches synchronously" claim against `actor_dispatch.go:48-60,135-138,267-299` and confirmed no placebo conversions and no weakened negative assertions.
- Reviewer findings applied before push (the only delta vs the reviewed diff): the MEDIUM diagnostic regression at the capacity-defer wait (timeout now reports last-observed spawn count + retrying state) and the single-wake-source assumption comment at the wake-select site. Remaining LOWs recorded: a signaling `staticStateRefresher` wrapper could make the operator-stop quiescence window start deterministically (pre-existing idiom); the wake-select is forward-fragile if a future `queuePollWake` producer lands earlier in the flow (documented at the site).
- **`@codex review` gate: ✅ clean.** Operator-triggered with the workspace-bound `bytevane` identity; `chatgpt-codex-connector` reviewed commit `bdbe1b38c3` (this head) and reported "Didn't find any major issues." Zero unresolved review threads.
- **All merge gates are satisfied** (CI green on head, dual local review converged, Codex clean, acceptance criteria met, `within budget` — 0 production LOC). Awaiting human merge permission per protocol §8.


bytevane:
@codex review

PR #824, opened 2026-06-13
test/e2e: clear the 5 pre-existing golangci e2e-tagged findings + lint e2e in CI

### LINKED ISSUE #802: test/e2e: clear the 5 pre-existing golangci blocking-set findings hidden by the missing e2e build tag in CI lint
## Context

While reviewing the #782 scripted-agent e2e test (PR for `fix/782-e2e-scripted-agent-loop`), running the golangci blocking set WITH `--build-tags=e2e` surfaced 5 findings in pre-existing e2e files that CI never sees (CI lints without the e2e tag, so build-tagged files are skipped):

- `test/e2e/happypath_test.go:233` — contextcheck
- `test/e2e/gitea.go:74` — errorlint
- `test/e2e/gitea.go:148` — errcheck
- `test/e2e/gitea.go:205` — unparam
- `test/e2e/gitea.go:246` — unused (`commentIssue`)

None are in the #782 diff; fixing them there would have expanded the PR beyond its issue scope.

## Acceptance criteria

- [ ] The 5 findings above are fixed (or individually justified with a nolint + reason).
- [ ] Decide whether the CI golangci gate should run a second pass with `--build-tags=e2e` over `test/e2e/` so this class cannot silently accumulate; implement or record the decision in this issue.

Found during the 2026-06-12 release-readiness batch (#780/#781/#782).

hey-yvan:
## Summary

CI's golangci-lint gate runs **without** the `e2e` build tag, so `test/e2e/*.go` (`//go:build e2e`) is never type-checked and that class of lint finding silently accumulates. The 2026-06-12 release-readiness batch surfaced **5 findings** under the blocking set with `--build-tags=e2e` (#802). This fixes all 5 and adds a CI pass so they can't recur.

## The 5 findings (all fixed)

| Linter | Location | Fix |
|---|---|---|
| contextcheck | `poll.go` `pollUntil` | Thread the caller's `ctx` through `pollUntil` and derive the poll deadline from it (`WithTimeout(ctx, …)`), so a cancelled/expired parent stops the poll instead of being ignored (was `context.Background()`). 4 call sites updated. |
| errorlint | `gitea.go:74` | Split admin-user-create into a wrapped exec error (`%w`, now also carrying the exit code) and a separate non-zero-exit message — no `%v` on an error. |
| errcheck | `gitea.go` `doJSON` | Explicitly ignore `resp.Body.Close()` in the defer. |
| unparam | `gitea.go` `putFile` | Every caller passed the constant `"WORKFLOW.md"`; drop the always-constant `path` param and rename to `putWorkflowFile`. |
| unused | `gitea.go` `commentIssue` | Delete the dead helper. |

**Timing safety of the contextcheck change:** deriving from the parent bounds each poll by `min(parent remaining, timeout)`. Headroom is ample at every call site — HappyPath 90s parent with a mock agent that completes in seconds; reconcile 90s parent / 10s poll; scripted-agent 300s parent / 90s+60s polls. No timeout regression; it is a strict correctness improvement (the poll now honors the test deadline).

## Acceptance criterion 2 — CI decision (implemented)

Per harness principle 3 (tighten the harness on an observed failure), CI now runs the **same blocking set a second time with `--build-tags=e2e` scoped to `./test/e2e/...`**, so build-tagged files can no longer escape lint.

## Verification

- `gofmt -l` clean · `go vet ./...` and `go vet -tags e2e ./test/e2e/...` clean · `go build ./cmd/...` · `go test -race ./...` green · size budget green.
- `go build -tags e2e ./test/e2e/...` compiles.
- **The CI guard is real and analyzes e2e files** (a Codex-family reviewer's degraded sandbox reported "no go files to analyze" — refuted by direct evidence): the exact CI command reports `0 issues` on this head; on `origin/main` it reproduces the 5 findings; and a mutation re-introducing the errcheck finding is **caught** (`gitea.go errcheck, 1 issues`). So the new step both fails on the class and passes once fixed.
- No standalone regression test per fix — these are lint cleanups; the new CI e2e-lint step is the class-level recurrence guard (clean-code rule 10), empirically shown to catch all 5 on the pre-fix tree.
- The full e2e suite runs in CI's required **E2E Gitea mock loop** job (Docker not available locally).
- Pre-push dual review: Claude-family `pr-review-toolkit:code-reviewer` **MERGE-READY** (independently reproduced base=5 / HEAD=0 and verified each fix); Codex-family `codex:codex-rescue` cleared every fix but returned BLOCKED solely on a "no go files to analyze" result from its sandbox — refuted above with a reproduced mutation. The one shared **LOW** (preserve the exit code when there's also an exec error) is applied.

Test-only + CI config; no SPEC-sensitive path, no new key/phase, no behavior change to the orchestrator. `internal/workflow/config.go` untouched; no new `internal/orchestrator`/`internal/worker` file.

### Size gate
- [x] `within budget` — 5 build-tagged test/e2e files + one CI workflow step; no production Go code changed.

Closes #802

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,17 @@ jobs:
           version: v2.12.2
           args: --config=.golangci.yml --enable-only=contextcheck,errcheck,errorlint,funlen,gocognit,gocritic,govet,ineffassign,revive,staticcheck,unparam,unused
 
+      # The e2e suite is build-tagged (`//go:build e2e`), so the default gate
+      # above never type-checks test/e2e/*.go and that class of lint finding
+      # silently accumulated (#802). Run the same blocking set a second time
+      # with the e2e tag, scoped to test/e2e, so build-tagged files can't
+      # escape lint.
+      - name: Run golangci-lint blocking gate (e2e build tag)
+        uses: golangci/golangci-lint-action@82606bf257cbaff209d206a39f5134f0cfbfd2ee # v9.2.1
+        with:
+          version: v2.12.2
+          args: --config=.golangci.yml --enable-only=contextcheck,errcheck,errorlint,funlen,gocognit,gocritic,govet,ineffassign,revive,staticcheck,unparam,unused --build-tags=e2e ./test/e2e/...
+
       - name: Verify Dockerfile Go version matches go.mod
         shell: bash
         run: |
--- a/test/e2e/gitea.go
+++ b/test/e2e/gitea.go
@@ -69,9 +69,13 @@ func startGitea(ctx context.Context) (*giteaEnv, error) {
 		"--email", "aiops-bot@example.invalid",
 		"-c", "/etc/gitea/app.ini",
 	})
-	if execErr != nil || code != 0 {
+	if execErr != nil {
 		_ = c.Terminate(ctx)
-		return nil, fmt.Errorf("create gitea admin user: exit %d err %v", code, execErr)
+		return nil, fmt.Errorf("create gitea admin user (exit %d): %w", code, execErr)
+	}
+	if code != 0 {
+		_ = c.Terminate(ctx)
+		return nil, fmt.Errorf("create gitea admin user: unexpected exit code %d", code)
 	}
 
 	host, err := c.Host(ctx)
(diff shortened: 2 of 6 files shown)


bytevane:
@codex review

bytevane:
@codex review

PR #908, opened 2026-06-16
fix(runner): drain codex app-server stderr before cmd.Wait

### LINKED ISSUE #907: codex app-server: stderr drain joined after cmd.Wait can truncate a final stderr write
Latent defect in `internal/runner/codex_app_server.go` (`RunCodexAppServer`), confirmed against source + the Go stdlib docs (`go doc os/exec.Cmd.StderrPipe`).

stderr is drained by a goroutine (`io.Copy(buf, stderr)`, started at `cmd.Start`) but joined (`<-stderrDone`) **after** `cmd.Wait()`. The pipe is `cmd.StderrPipe()` (`openAppServerPipes`), whose doc states:

> Cmd.Wait will close the pipe after seeing the command exit ... It is thus incorrect to call Wait before all reads from the pipe have completed.

The adjacent **stdout** path already drains-before-Wait correctly (with a comment citing the StdoutPipe doc); stderr was asymmetric. A final stderr write emitted as the child exits (e.g. a bwrap `setting up uid map: Permission denied` denial) can be truncated when `Wait` closes the read fd under the in-flight `io.Copy`. That denial line is the only signal `classifySandboxStartupFailure` uses to re-tag the run as a `*SandboxStartupError` and route it to the #550 cooldown; losing it reverts to hot per-tick retry.

## Fix
Join `<-stderrDone` before `cmd.Wait()`, mirroring stdout. No deadlock: `io.Copy` returns when the child closes stderr on exit (stderr is drained concurrently from `Start`, so the child never blocks on a full pipe).

Behavior-preserving; no SPEC deviation, no new config keys.

---
_A second finding from the same audit — `responseForID` classifying JSON-RPC messages by id shape rather than `method` presence — was split out as a separate design question (see the follow-up issue) because the correct handshake behavior for an interleaved server->client request is a design fork that needs the upstream Elixir `app_server.ex` as tiebreaker._

hey-yvan:
Closes #907

## Summary
`RunCodexAppServer` (`internal/runner/codex_app_server.go`) joined the stderr drain (`<-stderrDone`) *after* `cmd.Wait()`. The pipe is `cmd.StderrPipe()`, whose Go doc states it is *"incorrect to call Wait before all reads from the pipe have completed"* (Wait closes the pipe on process exit). The adjacent **stdout** path already drains-before-Wait correctly (with a comment citing the StdoutPipe doc); stderr was asymmetric.

A final stderr write emitted as the child exits — e.g. a bwrap `setting up uid map: Permission denied` denial that `classifySandboxStartupFailure` depends on for the #550 cooldown — could be truncated when `Wait` closed the read fd under the in-flight `io.Copy`, reverting to hot per-tick retry.

**Fix:** join `<-stderrDone` before `cmd.Wait()`, mirroring stdout. No deadlock — `io.Copy` returns when the child closes stderr on exit, and stderr is drained concurrently from `cmd.Start` so the child never blocks on a full pipe and the join never stalls Wait. Verified against `go doc os/exec.Cmd.StderrPipe` on the pinned Go 1.25.

> A second finding from the same audit — `responseForID` classifying JSON-RPC messages by id shape rather than `method` presence — was **split out to #909** as a design question. The naive guard was prototyped here and rejected in review (it would silently drop an interleaved `method`+`id` server request and could stall `request()` to its read timeout); the correct handshake behavior needs the upstream Elixir `app_server.ex` as tiebreaker, so it is not bundled into this fix.

Behavior-preserving; touches only `internal/runner` (not a SPEC-sensitive path). No SPEC deviation.

- [x] `within budget` — 1 production file / +18/-10 LOC.

## Testing
- [x] `go vet ./...`
- [x] `go test -race -covermode=atomic ./...` (full repo green; `internal/runner` re-run green after the revert)
- [x] `gofmt -l` clean + blocking golangci gate (0 issues on `internal/runner/...`)
- [x] additional: correctness-by-construction per the Go `StderrPipe` doc + the existing stdout drain pattern; backstopped by the `-race` runner suite. Deadlock-freedom on the clean / kill (`terminateProcess` SIGTERM+SIGKILL) / context-cancel exit paths confirmed in dual local review.

## Notes
- GraphQL mutation gate, sandbox arg construction, and all SPEC-sensitive paths untouched.

### DIFF
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -83,19 +83,27 @@ func (CodexAppServerRunner) Run(ctx context.Context, in RunInput) (Result, error
 		terminateProcess(cmd)
 	}
 	_ = stdin.Close()
-	// Drain stdout to EOF BEFORE cmd.Wait — the os/exec idiom (StdoutPipe doc:
-	// "incorrect to call Wait before all reads from the pipe have completed").
-	// The reader keeps scanning as the drain receives, so the child can never
-	// block writing to a full pipe; it finishes, observes the stdin EOF above (or
-	// the kill), exits, and its stdout EOF makes the reader close readCh, ending
-	// the drain. Waiting before draining would deadlock: a reader parked on an
-	// unconsumed send stalls stdout, the child blocks on write and never exits,
-	// and cmd.Wait never returns. Draining also joins the reader so no goroutine
-	// outlives this call (Go Code Review Comments, "Goroutine Lifetimes").
+	// Drain BOTH pipes to EOF BEFORE cmd.Wait — the os/exec idiom (StdoutPipe /
+	// StderrPipe docs: "incorrect to call Wait before all reads from the pipe
+	// have completed", because Wait closes the pipe once it sees the process
+	// exit). The reader keeps scanning as the drain receives, so the child can
+	// never block writing to a full pipe; it finishes, observes the stdin EOF
+	// above (or the kill), exits, and its stdout EOF makes the reader close
+	// readCh, ending the drain. Waiting before draining would deadlock: a reader
+	// parked on an unconsumed send stalls stdout, the child blocks on write and
+	// never exits, and cmd.Wait never returns. Draining also joins the reader so
+	// no goroutine outlives this call (Go Code Review Comments, "Goroutine
+	// Lifetimes").
 	for range client.readCh { //nolint:revive // drain-to-close joins the reader goroutine
 	}
-	waitErr := cmd.Wait()
+	// Join the stderr drain BEFORE Wait for the same reason: Wait closes the
+	// StderrPipe fd on process exit, which would truncate a final stderr write —
+	// e.g. a bwrap sandbox-startup denial that classifySandboxStartupFailure
+	// depends on (#550) — if io.Copy were still reading. io.Copy returns once the
+	// child closes stderr on exit (concurrent draining means the child never
+	// blocks on a full stderr pipe), so this join never stalls Wait.
 	<-stderrDone
+	waitErr := cmd.Wait()
 	elapsed := time.Since(start)
 
 	writeAppServerArtifact(in.Workdir, buf)


bytevane:
@codex review

bytevane:
@codex review — re-review: the diff was reduced to the stderr-drain fix only (the responseForID change was reverted and split to #909). Please review HEAD 072b079.

hey-yvan:
Closing without merge — on full review this was a **misdiagnosis**, not a defect.

The cloud @codex review flagged P1 *"Do not bypass WaitDelay for inherited stderr"*, which is correct. `setupAppServerCommand` sets `cmd.WaitDelay = killGrace` (codex_app_server.go:160). With `WaitDelay` set, the existing `stderr` drain ordering (`<-stderrDone` **after** `cmd.Wait()`) is intentional and correct:

- `cmd.Wait()` gives the stdout/stderr pipes up to `WaitDelay` of grace for the concurrent readers to finish before force-closing them, so a final stderr write (the bwrap-denial concern) is **not** truncated; and
- if a grandchild inherits stderr and lingers, `WaitDelay` bounds the wait and force-closes the pipe, so `io.Copy` returns and there is no hang.

Moving `<-stderrDone` **before** `cmd.Wait()` (this PR) defeats that: on a clean exit (no `terminateProcess`/SIGKILL) where a grandchild holds stderr open, `io.Copy` never returns, `<-stderrDone` blocks forever, and `cmd.Wait()`/`WaitDelay` are never reached → permanent hang. The stdout-drains-before-Wait / stderr-drains-after-Wait asymmetry is deliberate: stdout uses an unbuffered `readCh` (backpressure → must drain first), stderr uses `io.Copy` to a buffer (no backpressure → `WaitDelay` handles the rest). The Go `StderrPipe` "drain before Wait" guidance is the `WaitDelay == 0` case.

No code change needed. The second audit finding (`responseForID` classifying by id shape vs `method`) is a low-priority design question tracked in #909. Lesson: validate a Go pipe-lifecycle "fix" against `cmd.WaitDelay`, not the StderrPipe doc alone.

Issue #941, opened 2026-06-18
Trace L4 M4: add advisory evaluator candidates for recurring harness failures


hey-yvan:
## Summary

Implement Milestone 4 from `docs/design/trace-driven-harness-improvement.md`: when a grouped cluster proves a recurring failure class, add a reviewed advisory evaluator candidate with fixtures and false-positive expectations.

Part of #937. Depends on #938 and #939; it should use evidence from reviewed clusters before proposing an evaluator. Follow-up to #931 / PR #936.

## Problem

Evaluators are useful only when they preserve durable learning without becoming premature gates. The first evaluator milestone should produce report-only signal so the repo can review usefulness, false positives, and missing coverage before any gate decision.

## Required behavior

Add a path for a reviewed cluster to become an advisory evaluator candidate. The evaluator should:

- target one recurring failure class from grouped reports,
- include fixtures or examples derived from bounded, redacted evidence,
- state expected true-positive and false-positive behavior,
- run in report-only/advisory mode,
- publish output that can feed future L4 reports,
- document what evidence would justify later gate promotion.

## Boundary requirements

- Evaluators ship through ordinary code review before affecting future runs.
- First evaluator candidates must not be required CI/runtime gates.
- Promotion to a required gate is a separate later decision after review history and evidence.
- The worker remains scheduler/runner/tracker reader.

## Non-goals

- No immediate required merge gate.
- No natural-language comment parser as a machine contract unless reviewed here.
- No unbounded trace/log retention.
- No automatic prompt/rubric/skill rewrite.

## Acceptance criteria

- [ ] At least one advisory evaluator candidate can be generated or implemented from a reviewed failure cluster.
- [ ] The evaluator includes fixtures/examples and expected signal behavior.
- [ ] The evaluator runs in report-only mode and does not block merge/runtime.
- [ ] Documentation explains false-positive handling and what evidence would justify later gate promotion.
- [ ] Output can be referenced by future grouped reports.

Issue #937, opened 2026-06-18
Trace L4 implementation tracker


hey-yvan:
## Summary

Track the implementation follow-ups from #931 and PR #936 for the trace-driven harness improvement loop.

The design source of truth is `docs/design/trace-driven-harness-improvement.md` on `main`.

## North star

Real run feedback should become durable harness learning. Reports diagnose, issues and PRs transport, and evaluators remember regressions. A confirmed failure class should eventually become one of:

- a reviewed harness improvement,
- a reviewed regression check,
- or an explicit known limitation with evidence.

If repeated failures only produce more observations and never change a harness surface or tracked decision, the loop is not improving.

**Goal, not means:** per the North Star the goal is *durable harness learning that makes future runs fail less* — **not more automation or more reports**. Scheduling/automation is only a means and must be justified by the failure→durable-fix conversion rate, never pursued as the point.

## Milestone issues

**Artifact-production half** (given evidence → reviewable proposal) — delivered:

- [x] #938 — Trace L4 M1: generate grouped harness reports from bounded evidence
- [x] #939 — Trace L4 M2: render actionable issue and draft-PR proposals
- [x] #940 — Trace L4 M3: run agent follow-through from approved harness proposals
- [x] #941 — Trace L4 M4: add advisory evaluator candidates for recurring harness failures

**Closing the ratchet (M5–M6) + optional automation (M7)** — open. M1–M4 produce reports and proposals, but nothing yet guarantees a confirmed failure class becomes a reviewed fix / regression check / documented limitation, and recurrence is not remembered. M5–M6 are the loop-closing core; M7 is deferrable automation on top. Milestone numbers now follow execution order:

- [ ] #952 — Trace L4 M5: durable, redacted evidence input — the precondition for learning from real runs
- [ ] #953 — Trace L4 M6: close the ratchet (evaluator-result feedback + recurrence reopen) — **the goal-bearing milestone** (evaluators remember regressions); does **not** depend on the M7 driver
- [ ] #951 — Trace L4 M7: scheduled outer driver — a **secondary means** to cut operator toil; earns its place only if it raises the failure→durable-fix conversion rate, not by running more reports. Sequenced last; may be deferred.

## Non-goals

- Do not add a worker-owned post-turn verifier.
- Do not add orchestrator-owned PR creation, merge, or tracker write-back.
- Do not silently rewrite `WORKFLOW.md`, rubrics, skills, or `LEARNINGS.md`.
[...]

PR #962, opened 2026-06-20
fix(harness): dedupe byte-identical trace inputs to stop omitted inflation

### LINKED ISSUE #961: Trace harness: omitted affected-id counts double-count across overlapping manifests
Follow-up from PR #960 (#958). Surfaced by `@codex review`.

## Problem

`merge_manifest_affected_omitted` (`scripts/trace-harness-report.py`) **sums** the
per-class `affected_by_class.<class>.omitted` counts across every fold (multiple
`--evidence-manifest` inputs, or multiple runs of the same class within one
manifest). The kept affected-id arrays are deduped by `add_unique`, but `omitted`
is a bounded **count** — the manifest dropped the id *values* to honor the 256 KiB
run-group byte bound — so overlapping inputs cannot be deduped here and inflate
the cluster's omitted vs the raw `--worker-log` path.

Reproduction (same capped manifest passed twice):

```
manifest x1      : kept=4307 omitted=15693
manifest x2 (dup): kept=4303 omitted=31283   <- doubled
raw log x2  (dup): kept=4303 omitted=15697    <- raw dedupes
```

## Scope of impact

- **Advisory only.** The recovered affected-id *lists* stay correct (kept ids are
  deduped). Only the omitted-count-derived `trimmed_affected_occurrence_count`
  (recurrence signal) and the `recovered_affected_ids.omitted` display can
  over-state recurrence. No worker/CI gate consumes it (report-only).
- The PR #960 / #958 acceptance criterion — a **single** capped run's
  manifest round trip matching the raw path — is unaffected and exact.

## Why deferred (inherent to a bounded count summary)

Faithful cross-fold dedupe of omitted ids would require preserving the dropped id
*values*, which defeats the 256 KiB byte bound that the summary exists to honor.
So this is a fidelity limitation of summarizing dropped ids as a count, not a
local bug.

## Options to validate before implementing

1. Dedupe identical manifest/worker-log inputs by their `sha256` (already computed
   in `input_ref`) before folding — cleanly fixes "same input supplied twice", not
   overlapping-but-distinct slices.
2. Take `max` instead of `sum` per (cluster, key) — exact for identical overlap,
   but under-counts genuinely-disjoint same-class multi-run folds (would weaken
   the #941 trimmed-occurrence recurrence signal).
3. Document it as a permanent bounded-summary limitation (count-based omitted
   cannot dedupe across overlapping inputs).

## Acceptance criteria

- [ ] Decide the approach (validate against the raw-path parity goal; do not
      regress the normal single-manifest multi-run case).
- [ ] Overlapping-input round-trip omitted accounting matches the raw
      `--worker-log` path under the chosen approach, or a reviewed
      documented-limitation decision is recorded.
- [ ] Test covers the overlapping-manifest case; mutation-verified if behavioral.

Part of #937.

hey-yvan:
Closes #961

## Summary
- A trace-evidence manifest's per-class `affected_by_class` summary stores a bounded `omitted` **count** whose dropped id *values* are discarded to honor the 256 KiB byte bound. The report's `merge_manifest_affected_omitted` **sums** those counts across folds, so passing **byte-identical** inputs twice (the same manifest, or the same worker log) doubled the recovered `omitted` vs the raw `--worker-log` path (the kept id *lists* stayed correct — only the advisory count diverged). The same-log-twice case also doubled a cluster's **evidence** entries on the raw path.
- **Fix:** collapse byte-identical inputs by their `sha256` (already computed in `input_ref`) **before** folding, via one shared helper `unique_inputs_by_digest(paths, kind)` used by both the report (`generate`, for `--worker-log` and `--evidence-manifest`) and the producer (`build_manifest`, for `--worker-log`). First-seen order is preserved; the `inputs` array lists only the unique inputs.
- **Decision (research to a verdict, not a menu):** keep `sum` — **not** `max` (which would under-count genuinely-disjoint same-class multi-run folds and weaken the #941 trimmed-occurrence recurrence signal). Two *distinct* manifests with **overlapping** content keep different digests and cannot be deduped here (the dropped values are gone), so their `omitted` counts are still summed and may over-state recurrence; documented as an inherent bounded-summary limitation with a recommendation to use **non-overlapping** slices.

Pure `scripts/` analysis tooling over operator-retained logs (the #937/#952 trace-driven harness loop input). No worker/orchestrator change, no new evidence source, restart recovery untouched (#73/#407). The digest is over raw bytes — no opaque/free-text parsing.

- [x] `within budget` — production diff = 2 Python files / small LOC (tests + docs excluded).

## Testing
- [x] `go vet ./...`
- [x] `gofmt -l` clean + blocking golangci gate (0 issues on `./scripts/...`)
- [x] additional validation noted below

### Acceptance criteria (issue #961)
- [x] Approach decided and validated against the raw-path parity goal; the normal single-manifest multi-run case is not regressed (distinct inputs keep distinct digests).
- [x] Identical-input round-trip omitted accounting now matches the raw `--worker-log` path (same manifest twice: omitted `31283`→`15693`, equal to a single manifest and to the raw path); the residual distinct-but-overlapping case is a recorded documented-limitation decision.
- [x] Tests cover the overlapping (byte-identical) case on all three paths — report `--evidence-manifest`, report `--worker-log`, producer `--worker-log`; **mutation-verified** (disabling dedup makes `inputs=2` and doubles omitted/evidence/events → tests fail).

## Notes
- Pre-push local Claude code-reviewer: clean, recommend approval, with its own two mutation passes confirming the tests catch the real omitted/evidence/events double-count (not just the `inputs==1` short-circuit).
- Bonus: also fixes a latent raw-path bug where the same `--worker-log` passed twice doubled a cluster's evidence entries.

### DIFF
--- a/docs/runbooks/trace-evidence-manifest.md
+++ b/docs/runbooks/trace-evidence-manifest.md
@@ -29,6 +29,15 @@ python3 scripts/trace-evidence-manifest.py \
 Multiple `--worker-log` flags are allowed. The command writes compact JSON, so
 for local inspection pipe it through `python3 -m json.tool` or `jq`.
 
+Byte-identical inputs are collapsed by their `sha256` digest before folding, so
+passing the same log (or, for the report, the same manifest) twice does not
+double events or inflate the per-class affected summary (#961). Two *distinct*
+inputs whose contents **overlap** (for example, log-rotation slices that share
+lines) keep different digests and are not deduped: their kept affected-id lists
+still merge correctly, but the per-class `omitted` counts — bounded counts whose
+dropped id values are gone — are summed and may over-state recurrence. Prefer
+**non-overlapping** log slices.
+
 ## What is captured (metadata first)
 
 The output uses schema `trace-evidence-manifest/v2`.
--- a/scripts/trace-evidence-manifest.py
+++ b/scripts/trace-evidence-manifest.py
@@ -79,9 +79,13 @@ def main(argv: list[str]) -> int:
 def build_manifest(paths: list[Path]) -> dict:
     if not paths:
         raise ValueError("at least one --worker-log is required")
-    inputs = [report.input_ref(path, "worker_log") for path in paths]
+    # Collapse byte-identical logs so passing the same file twice does not double
+    # events or inflate the per-class affected summary (#961); shared with the
+    # report so both tools dedupe inputs identically.
+    unique = report.unique_inputs_by_digest(paths, "worker_log")
+    inputs = [ref for _, ref in unique]
     runs: dict[str, dict] = {}
-    for path in paths:
+    for path, _ in unique:
         for finding in report.parse_worker_log(path):
             fold_event(runs, finding)
     return {
--- a/scripts/trace-harness-report.py
+++ b/scripts/trace-harness-report.py
@@ -136,13 +136,14 @@ def generate(worker_logs: list[Path], manifests: list[Path] | None = None, prior
         raise ValueError("at least one --worker-log or --evidence-manifest is required")
     clusters: dict[str, dict] = {}
     run_bytes: dict[str, int] = {}
-    inputs = [input_ref(path) for path in worker_logs]
-    inputs += [input_ref(path, "evidence_manifest") for path in manifests]
+    worker_inputs = unique_inputs_by_digest(worker_logs, "worker_log")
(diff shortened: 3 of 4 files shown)


hey-yvan:
@codex review

bytevane:
@codex review

Issue #977, opened 2026-06-21
feat(dashboard): show active agent model and runtime profile


hey-yvan:
## Summary

The worker dashboard currently shows worker version, state counts, token usage, runtime, rate-limit plan, running issue/session IDs, and process IDs, but it does not show which agent model is being used for the active run.

This became visible during the Docker-backed web-todo E2E when we needed to distinguish a project/harness defect from behavior specific to `GPT-5.3-Codex-Spark`. The dashboard made it easy to see that a reviewer claim was running, but not which model/runtime profile produced the git behavior under investigation.

## Problem

Today the dashboard includes fields like:

```text
aiops v0.1.8
GET /api/v1/state
running/retrying/blocked/delivered counts
input/output/total tokens
runtime total
rate_limits.limit_id = codex
rate_limits.plan_type = prolite
running[].issue_id/session_id/pid/tokens
```

But it does not expose model identity, for example:

```text
model: GPT-5.3-Codex-Spark
provider/runtime: codex-app-server
profile/config source: reviewer vs maker workflow
```

Rate-limit plan/limit fields are not enough: they describe billing/usage windows, not the actual model or runtime profile used by a running claim.

## Recommended UX

Use a layered display so the main dashboard stays scannable while still preserving debugging detail.

1. Header / top summary

Show the worker-level default runtime profile near the existing version and endpoint metadata, for example:

```text
aiops v0.1.8 · codex-app-server · default model GPT-5.3-Codex-Spark
```

If the model is not configured or cannot be detected, render:

```text
model unknown
```

Do not hide the field entirely; an unknown model is itself useful diagnostic data.

2. Running table

Add a compact `MODEL` column to the Running table because concurrent claims may eventually run with different model/runtime settings. Suggested order:

```text
ISSUE | STATE | MODEL | RUNTIME | LATEST ACTIVITY | TOKENS IN / OUT
```

For narrow/mobile layouts, collapse model into the issue detail line under the issue path/session metadata, for example:

```text
model GPT-5.3-Codex-Spark · codex-app-server · pid 127890
```

3. Claim detail / expanded row / JSON state

Expose structured fields in `/api/v1/state` for each running claim when known:

```json
{
  "agent_provider": "codex-app-server",
  "agent_model": "GPT-5.3-Codex-Spark",
  "agent_profile": "reviewer",
  "workflow_path": "/path/to/reviewer-automerge-WORKFLOW.md"
}
```

[...]

PR #984, opened 2026-06-22
feat(dashboard): expose per-run workflow profile on /api/v1/state

### LINKED ISSUE #983: feat(dashboard): expose per-run workflow profile (reviewer/maker) on /api/v1/state
Follow-up from #977 / #982.

#982 surfaced per-claim **agent model** + **runtime/provider** and the worker default provider on `/api/v1/state` and the dashboards. It deliberately did **not** add the issue's illustrative `agent_profile` / `workflow_path` field (reviewer vs maker workflow), because there is no existing per-run workflow-identity concept on the state surface — inventing one was out of scope for that PR.

## What
Expose which WORKFLOW.md/service produced a run (the "profile", e.g. `reviewer` vs `maker`) per active claim on `/api/v1/state`, and render it on the web + TUI dashboards alongside the model/provider already shipped in #982.

## Why
When running maker + reviewer workflows concurrently against one repo, operators want to correlate a failure with the workflow/profile that produced it, not just the model. #977's "tell a project/harness defect apart from model-specific behavior" intent is mostly served by model+provider; this closes the remaining "which workflow profile" gap.

## Notes
- SPEC §13.7.2 permits additional tracked summary fields, so this stays within the observability envelope (no worker phase/gate/tracker write) — same boundary reasoning as #982.
- Requires threading a per-run workflow/service identity into the running state (RunningEntry → RunningView → stateapi.Running); pick existing domain naming (service/workflow) rather than the illustrative `agent_profile`.

hey-yvan:
## What

Surface which WORKFLOW.md (the "profile", e.g. reviewer vs maker) produced each active claim on `/api/v1/state` and the web + TUI dashboards, so an operator running maker + reviewer workflows can correlate a failure with the workflow that produced it — not just the model/provider shipped in #982. Follow-up to #977/#982.

`Closes #983`

## How

Folds the worker's existing `workflow_resolved` task event (`workflow.Resolution` `source`/`path`) into the orchestrator's live session, mirroring #982's session-fold pattern:

- **orchestrator**: `LiveSession.WorkflowSource/WorkflowPath` ← `workflow_resolved`, projected `RunningView → stateapi.Running`. The event is forwarded to `RecordRuntimeEvent` **explicitly** (an extra clause in `runtimeEventForwardingEmitter`), **not** by adding it to `task.RuntimeEvents()` — that set is the SPEC §10.4 runner app-server vocabulary, and `workflow_resolved` is a worker lifecycle event. `recordRuntimeEvent` special-cases it: folds only the profile and returns **before** the general fold, so it never lands in `last_event`/`last_message` or resets the §8.5 stall clock (`last_event_at`).
- **stateapi**: `Running.workflow_source`/`workflow_path` (both `omitempty`; a missing value renders "unknown"). SPEC §13.7.2 permits added summary fields — pure read-only observability, **no** worker phase/gate/tracker write.
- **web dashboard**: per-claim detail line gains `workflow <path|source|unknown>` alongside model/provider.
- **TUI**: a worker-level `Workflow:` header line summarizing the resolved workflow from the running rows. A single TUI views one worker, which resolves one workflow, so a per-row column would be N identical cells stealing width from the EVENT column; the per-claim API field remains the source of truth (and serves cross-worker aggregation).
- **docs/tests**: documents the 4 keys in `runtime-status.md` and pins them in the drift synthetic; also closes a #982 gap where `agent_provider`/`agent_model` were surfaced but never documented or set in the drift synthetic (so the drift detector silently skipped them).

## Acceptance criteria

- [x] `workflow_source`/`workflow_path` exposed per active claim on `/api/v1/state` (RunningEntry → RunningView → stateapi.Running), named from the existing `Resolution` domain (not the illustrative `agent_profile`).
- [x] Rendered on web (per-claim detail line) and TUI (worker-level header), "unknown" when unreported — never blank.
- [x] Default-workflow run (`source=default`, no path) distinguishable from not-yet-resolved.
- [x] No new worker/orchestrator phase, gate, or tracker write (boundary principle 6).

## Verification

```
gofmt -l … (clean) · go vet ./… (clean) · golangci-lint v2.12.2 (0 issues)
go test -race ./internal/orchestrator/… ./cmd/worker/… ./cmd/tui/… (pass)
go build ./cmd/worker ./cmd/tui (ok) · web vitest (24 pass)
```
Full `go test -race ./...` green except `./scripts` `TestTodoSmokeScriptRecordsDraftPRWhenWorkerFails`, which times out under full `-race` load (10s budget) and passes in isolation (5.1s); the diff touches no `scripts/` file (known timing flake, not a regression).

## Mutation verification (against the committed artifact)

Every new production seam was broken and the matching test confirmed to FAIL, then restored: forwarding clause, the `recordRuntimeEvent` early-return (last_event isolation), `recordWorkflowFields` fold, snapshot projection, `apiRunningFromView` mapping, TUI `resolvedWorkflowLabel` (all branches: path-preference, source-fallback, continue-not-break, dedup, join, unknown), and the web `workflowProfile` helper.

## Review

Pre-push dual review (Codex-family + Claude-family subagents) both confirmed the production code correct and the observability boundary clean, and flagged one class: the `WorkflowSource`-only fallback arm + dedup/join branch of `resolvedWorkflowLabel`/`workflowProfile` weren't pinned by render tests. Closed structurally with a comprehensive mutation-tight table test (`TestResolvedWorkflowLabel`) + a web fallback case; production code unchanged (one clarifying comment added). Declined the lifecycle-ordering hardening test: the `run != nil` no-op is a general property of all runtime-event folds (the Claude reviewer separately confirmed nil/ordering safety), so a contrived race test would be a placebo.

### Size gate
- [x] `within budget` — production diff is 8 files / ~107 LOC (mostly doc-comments); tests + docs excluded.

https://claude.ai/code/session_01EkiVcnH5eXSxZfz3wnmEc5

This PR adds only SPEC §13.7.2-permitted per-claim observability summary fields (`workflow_source`/`workflow_path`), folded from an event the worker already emits — no new Config key, phase, gate, artifact, or tracker write.

## `@codex review` status
The GitHub Codex bot returned **"reached your Codex usage limits for code reviews"** on this head (`c5fbd13`) — externally unavailable, not a clean/finding signal. Per the review protocol §4, this is compensated by the equivalent independent **dual subagent review** run pre-push (Codex-family + Claude-family), both of which confirmed the production code correct and the observability boundary clean; their only findings (test-coverage gaps in the profile-label fallback/dedup/join arms) were closed with the mutation-verified `TestResolvedWorkflowLabel` table test + a web fallback case. The current head's production code is unchanged from the reviewed head (`e6e0842`) apart from one clarifying comment; the amend was test-only. No unresolved, non-outdated review threads.

### DIFF
--- a/cmd/tui/layout_test.go
+++ b/cmd/tui/layout_test.go
@@ -145,6 +145,8 @@ func representativeState(now time.Time) *stateapi.StateResponse {
 			CodexAppServerPID: 12345,
 			AgentProvider:     "codex-app-server",
 			AgentModel:        "gpt-5.3-codex-spark",
+			WorkflowSource:    "file",
+			WorkflowPath:      "/srv/reviewer/WORKFLOW.md",
 		}},
 		Retrying: []stateapi.Retry{{
 			Identifier: "ENG-9999",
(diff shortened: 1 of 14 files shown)


bytevane:
@codex review

PR #1044, opened 2026-07-02
fix(orchestrator): drain in-flight agent workers on shutdown

### LINKED ISSUE #1030: [P1][bug] shutdown: main exits without draining in-flight agent workers; spawn fanout's runCtx.Done branch skips finalize
## Summary

On SIGTERM/SIGINT the worker cancels the root context, `orchestrator.RunPollLoopWithRuntime` returns, and `main` returns immediately — the Go runtime then kills the process without waiting for in-flight agent workers. Nothing drains `WorkerTaskDispatcher.Spawn` goroutines, so:

- the codex app-server subprocess kill (via `exec.CommandContext` cancel) races bare process exit;
- `after_run` hooks and workspace ownership release never run;
- run artifacts/events for the in-flight turn are lost.

Additionally, the spawn result fanout's shutdown branch closes `workerDone` without submitting `finalizeRunOp`, so `Running`/`Claimed` state is never cleaned on this path (in-process zombie until exit):

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/internal/orchestrator/actor_dispatch.go#L314-L317

```go
case <-o.runCtx.Done():
    cancel(nil)
    close(workerDone)
    return
```

The comment at `actor_dispatch.go:325` references a "graceful-shutdown drain path", but no such drain exists in `cmd/worker/main.go`:

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/cmd/worker/main.go#L416-L450

`grep -r "WaitGroup\|drain" cmd/worker internal/worker` finds no shutdown drain of spawned workers.

## Why this is a port gap, not a SPEC deviation

Per AGENTS.md cross-cutting checklist item 2 ("Replicate Elixir's implicit runtime guarantees explicitly in Go"): on BEAM, a supervision-tree shutdown terminates children in order and closing a `Port` reliably tears down the OS subprocess. A Go `main` return provides neither — child processes are orphaned unless the `CommandContext` kill goroutine wins the race against process exit. SPEC's restart-recovery posture (no persistent DB) tolerates abrupt state loss, but it does not require leaking live codex subprocesses or skipping `after_run` hooks on an ordinary SIGTERM.

## Failure scenario

1. Operator (or systemd/k8s) sends SIGTERM while N agents are mid-turn.
2. `ctx` cancels; poll loop returns; `main` returns within milliseconds.
3. The `exec.CommandContext` watchdogs may or may not deliver the kill before process exit; on loss, codex app-server processes are orphaned and keep consuming API quota until they finish or hit their own timeouts.
4. Workspace ownership locks are left held by a dead pid; `after_run` hooks never fire.

## Suggested direction

After the poll loop returns, wait (with a bounded grace period, e.g. `WorkerExitTimeout`) for all `Running` entries' `Done` channels before returning from `run()`. The `runCtx.Done()` fanout branch should still ensure worker cancellation, and the bounded wait keeps SIGKILL-after-grace semantics for supervisors.

Related prior work: #197 fixed the dropped-`finalizeRunOp`-submit leak of `workerDone`; this issue covers the remaining shutdown-drain gap.

hey-yvan:
Closes #1030

## Summary
- SIGTERM/SIGINT previously returned from `run()` the instant the poll loop exited; the process died racing the runner's subprocess kill, orphaning codex subprocesses and skipping worker result collection.
- `spawn` now registers every worker with a drain `WaitGroup`; the result fanout goroutine, on `runCtx` cancellation, cancels the worker and **keeps waiting for its result** (per the Dispatcher one-result contract) instead of returning early.
- `cmd/worker` calls the new `Orchestrator.WaitForWorkers(30s)` after the poll loop returns: shutdown now waits for in-flight agents to exit (bounded by the same 30 s the reconcile `WorkerExitTimeout` uses); a second signal still terminates immediately.
- This is the explicit Go replacement for the ordered child shutdown a BEAM supervision tree provides implicitly (AGENTS.md cross-cutting checklist item 2).

No new phase/gate/artifact: this only changes process-exit ordering. Upstream BEAM gets ordered child termination from the supervision tree (`elixir/lib/symphony_elixir/application.ex` children shutdown semantics); a Go `main` return provides no equivalent, which AGENTS.md cross-cutting checklist item 2 classifies as an implicit runtime guarantee the port must replicate explicitly.

Production diff: 3 files (`actor.go`, `actor_dispatch.go`, `cmd/worker/main.go`), ~60 LOC.

## Testing
- [x] `go vet ./...`

Mutation-verified per clean-code rule 6: restoring the old `runCtx.Done()` early-return in the fanout goroutine fails `TestWaitForWorkers_DrainsInFlightWorkerOnShutdown` (`WaitForWorkers returned before the dispatcher delivered the worker result`); tree restored to HEAD afterwards. `TestWaitForWorkers_TimesOutOnStuckWorker` pins the bounded-grace contract.

## Notes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -30,6 +30,13 @@ import (
 // hence the explicit build ARG there).
 var version = "devel"
 
+// workerShutdownDrainGrace bounds how long run() waits for in-flight agent
+// workers after the poll loop exits on SIGTERM/SIGINT. It matches the
+// reconcile WorkerExitTimeout (workflow_runtime.go): workers that ignore
+// cancellation longer than this are abandoned to process exit, and a second
+// signal always terminates immediately.
+const workerShutdownDrainGrace = 30 * time.Second
+
 func resolveVersion() string { return buildinfo.Resolve(version) }
 
 func main() { //nolint:gocognit // baseline (#521)
@@ -447,7 +454,18 @@ func run(ctx context.Context, args []string) error { //nolint:gocognit,funlen //
 			log.Printf("workflow reload loop exited: %v", err)
 		}
 	}()
-	return orchestrator.RunPollLoopWithRuntime(ctx, poller, runtime, orchestrator.PollLoopRuntimeOptions{})
+	pollErr := orchestrator.RunPollLoopWithRuntime(ctx, poller, runtime, orchestrator.PollLoopRuntimeOptions{})
+	// SIGTERM/SIGINT lands here with in-flight agents still running. Drain
+	// them before returning from run(): a bare return exits the process,
+	// racing the runners' subprocess kill and skipping result collection
+	// (issue #1030). Workers see the canceled root context and exit on
+	// their own; the grace period only bounds the wait. A second signal
+	// still kills the process immediately (signal.NotifyContext stops
+	// relaying once canceled).
+	if !orch.WaitForWorkers(workerShutdownDrainGrace) {
+		log.Printf("event=shutdown_drain_timeout grace=%s note=%q", workerShutdownDrainGrace, "exiting with agent workers still running")
+	}
+	return pollErr
 }
 func reconciliationConfigForWorkflow(cfg workflow.Config) orchestrator.ReconciliationConfig {
 	// Single source of truth for the field list lives in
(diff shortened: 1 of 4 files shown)


zjlgdx:
@codex review

zjlgdx:
@codex review

PR #1046, opened 2026-07-02
fix(runner): answer server->client requests interleaved with pending RPCs

### LINKED ISSUE #1031: [P2][bug] runner: request() never answers server->client requests during initialize/thread/start/turn/start
## Summary

The turn loop correctly classifies JSON-RPC messages by method and answers server→client requests (messages carrying an `id`) via `handleServerRequest` → `replyServerRequest` (internal/runner/codex_app_server_turn.go:153-158). The startup/request path does not:

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/internal/runner/codex_app_server_transport.go#L38-L48

```go
for {
    msg, err := c.readMessage(ctx)
    if err != nil {
        return nil, err
    }
    if result, matched, rpcErr := responseForID(msg, id); matched {
        return result, rpcErr
    }
    c.handleNotification(msg)   // server requests (id + method) fall through here, never replied
}
```

`request()` is used for `initialize`, `thread/start`, and `turn/start` (codex_app_server.go:395,403,473). While one of those RPCs is awaiting its response, any server→client request codex emits (approval request, `item/tool/call`, elicitation) is treated as a notification: `handleNotification` only updates `continueRun`/`lastMessage` and never writes a JSON-RPC reply to stdin.

## Failure scenario

Codex interleaves a server request between the harness's `turn/start` send and the matching response (the protocol permits interleaving; the turn loop already defends against it). Codex blocks waiting for the reply; the harness keeps waiting for the `turn/start` response id → the run dies on `read timeout` / stall and is retried, with a misleading classification.

This is the same failure class the runner's own history warns about ("classify RPC messages by method", .trellis task 06-16): the fix landed on the turn loop but not on the sibling `request()` path — AGENTS.md clean-code rule 10 (fix the class, sweep the blast radius).

## Suggested direction

Inside `request()`'s read loop, route messages that carry both `id` and `method` through the same `replyServerRequest`/`handleServerRequest` machinery the turn loop uses (declining per approval policy), keeping only true notifications on `handleNotification`.

hey-yvan:
Closes #1031

## Summary
- `request()` (used by `initialize`, `thread/start`, `turn/start`) treated every non-response message as a notification, so a server->client request codex interleaved while the RPC awaited its response (approval prompt, user-input request, elicitation) was never replied to: codex blocked on the reply, the harness blocked on the response, and the run died on a misleading read-timeout/stall classification.
- The turn loop already classifies messages by method (`id` + `method` => `handleServerRequest`); `request()` now routes interleaved server requests through the same machinery (replying per approval policy, surfacing SPEC §10.4 input-required), and `responseForID` no longer mistakes a method-carrying message with a colliding id for the response.
- Same failure class as the earlier turn-loop fix ("classify RPC messages by method"); this sweeps the sibling `request()` path per clean-code rule 10.

No new phase/gate/artifact: extends the existing server-request reply machinery (SPEC §10.4) to the sibling request path.

Production diff: 2 files (`codex_app_server_transport.go`, `codex_app_server_response.go`), ~40 LOC.

## Testing
- [x] `go vet ./...`

Mutation-verified per clean-code rule 6 against the committed artifact:
- deleting the server-request branch in the request loop fails `TestRequest_AnswersInterleavedServerRequest` (no reply on stdin) and `TestRequest_InputRequiredServerRequestSurfacesInputRequired` (EOF instead of InputRequiredError);
- deleting the method-member check in `responseForID` fails `TestResponseForID/server_request_with_colliding_id_is_not_a_response`;
- tree restored to HEAD after each mutation.

## Notes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/runner/codex_app_server_request_test.go
+++ b/internal/runner/codex_app_server_request_test.go
@@ -126,6 +126,33 @@ func TestRequest_AnswersInterleavedServerRequest(t *testing.T) {
 	}
 }
 
+// TestRequest_RoutesInterleavedToolCallThroughToolHandler pins the review P2
+// on #1046: an id-bearing item/tool/call interleaved while an RPC awaits its
+// response must go through the dynamic-tool handler (here: the unsupported-
+// tool structured failure for an empty tool set), not the approval reply
+// table, which would reject an advertised tool with "Method not found"
+// solely because the call arrived before the pending response.
+func TestRequest_RoutesInterleavedToolCallThroughToolHandler(t *testing.T) {
+	c, stdin := newTurnLoopClient(t, []string{
+		`{"jsonrpc":"2.0","id":9,"method":"item/tool/call","params":{"tool":"nope","arguments":{}}}`,
+		`{"jsonrpc":"2.0","id":0,"result":{"ok":true}}`,
+	})
+	got, err := c.request(context.Background(), "turn/start", nil)
+	if err != nil {
+		t.Fatalf("request() err = %v; want nil", err)
+	}
+	if got["ok"] != true {
+		t.Errorf("request() result = %#v; want result.ok = true after answering the tool call", got)
+	}
+	wire := stdin.String()
+	if strings.Contains(wire, "Method not found") {
+		t.Errorf("stdin = %q; want the tool handler's structured reply, not a Method-not-found rejection", wire)
+	}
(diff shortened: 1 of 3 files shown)


zjlgdx:
@codex review

zjlgdx:
@codex review

Issue #1056, opened 2026-07-03
runner: audit the codex app-server multiAgentMode default


zjlgdx:
## Problem

PR #999 pinned Codex app-server `thread/start.multiAgentMode` to `none` to preserve the existing WORKFLOW-only instruction contract after Codex 0.142 introduced a different default. During #1049's local Codex environment inheritance review, this setting was re-checked and did **not** appear to be the same class of issue as the Apps/connectors disablement: `multiAgentMode: none` controls injected delegation instructions, while Apps/connectors, MCP, skills, and dynamic tools are separate capability surfaces.

That conclusion should still be audited deliberately instead of living only in a PR discussion or local investigation. The project should either keep `none` with evidence, adjust it, or document a tracked deviation if upstream/current Codex behavior proves the setting is suppressing desired unattended capability.

## Scope

- Re-read PR #999, issue #998, the current vendored Codex app-server schema, and the current target Codex app-server behavior.
- Compare with upstream Symphony's Elixir app-server payload, which does not currently send `multiAgentMode`.
- Verify whether `multiAgentMode: none` affects only injected multi-agent instructions or also changes tool availability in practice.
- Specifically check that MCP, skills, Apps/connectors, plugin-provided capability, and explicit runner `dynamicTools` remain available according to their own config/discovery rules when `multiAgentMode` is `none`.
- Decide one of: keep as-is with documented evidence, change the default, or track a SPEC/deviation follow-up if the correct behavior differs from upstream/reference expectations.

## Acceptance criteria

- A live/current Codex schema or app-server behavior check documents what `multiAgentMode: none` actually does.
- Tests or trace evidence cover the behavior that matters for unattended aiops-platform runs.
- If the setting is kept, docs or code comments explain why it is not a host-local capability disablement.
- If the setting is changed, tests catch both the instruction-injection regression from #998 and any tool/capability regression found here.
- #1049 remains focused on Apps/connectors, `CODEX_HOME`, shell environment inheritance, and repo-owned skill dependency support; this issue owns the separate #999 audit.

## References

- Follow-up from #1049's inheritance rollback audit.
- PR #999: `fix(runner): pin codex app-server multi-agent mode`.
- Issue #998: original multi-agent mode regression.

Issue #1049, opened 2026-07-02
workflow: support explicit repo-owned skill dependencies for unattended agents


hey-yvan:
## Problem

A real GitHub maker/reviewer run for `xrf9268-hue/bwu` exposed a workflow portability gap: the reviewer workflow needed the PR follow-through rules from `gh-pr-follow-through`, especially thread-aware `reviewThreads` handling, async Codex review completion, and stale/outdated unresolved bot-thread treatment.

The immediate local fix referenced `/Users/yvan/.agents/skills/gh-pr-follow-through/SKILL.md`, which is machine-specific. That absolute path should not be the reusable unattended workflow contract.

## Decision: use the workflow prompt pattern, not a new worker-side skill dependency mechanism

Do **not** add a structured `skills.required` workflow schema, worker startup/reload skill validation, state/dashboard skill evidence, or `worker --print-config` skill hashing for this issue.

The portable pattern is simpler and matches both Claude Code guidance and upstream Symphony: check the reusable skill into the repo or workflow template, then tell the agent in `WORKFLOW.md` natural language to open/invoke/follow that repo-relative skill file.

Authoritative guidance:

- Boris Cherny (`@bcherny`), Claude Code author, answered a chaining-skills question with: "ask claude to invoke skill 1, then skill 2, then skill 3, in natural language." He also noted parallel subagents can invoke skills in parallel, and the sequence can later be put into a skill if useful. Source: https://x.com/bcherny/status/2006170607092670691
- Upstream Symphony carries repo-owned skills under `.codex/skills/*`: https://github.com/openai/symphony/tree/4cbe3a9699a73b862466c0b157ceca0c1985d6d7/.codex/skills
- Upstream `elixir/WORKFLOW.md` lists related skills in the prompt and directly names `.codex/skills/land/SKILL.md` for the `Merging` flow: https://github.com/openai/symphony/blob/4cbe3a9699a73b862466c0b157ceca0c1985d6d7/elixir/WORKFLOW.md#L98-L104 and https://github.com/openai/symphony/blob/4cbe3a9699a73b862466c0b157ceca0c1985d6d7/elixir/WORKFLOW.md#L127-L128
- Upstream config schema has no `skills.required` equivalent; `codex.command` remains the codex runner schema surface: https://github.com/openai/symphony/blob/4cbe3a9699a73b862466c0b157ceca0c1985d6d7/elixir/lib/symphony_elixir/config/schema.ex#L164-L205
[...]

PR #1058, opened 2026-07-03
fix(runner): preserve inherited codex environment

### LINKED ISSUE #1057: runner: preserve inherited Codex environment for local app-server runs
## Problem

#1049 uncovered a rollback that should land before adding repo-owned skill dependency mechanics: the Codex app-server runner was narrowing the local Codex environment more than upstream Symphony does for same-user local binary deployments.

The rollback scope is narrower than #1049 as a whole. It should remove the default Apps/connectors disablement, preserve `CODEX_HOME`, and update shipped real-Codex templates to use upstream-style shell environment inheritance without closing #1049's larger required-skill feature work.

## Acceptance criteria

- `thread/start` no longer sends runner-owned config that disables Apps/connectors by default.
- `CODEX_HOME` is part of the agent subprocess baseline when set.
- Shipped real-Codex workflow templates use `codex app-server --config shell_environment_policy.inherit=all` explicitly.
- The core SPEC default remains `codex app-server`.
- Tests cover preserving inherited Codex config while keeping explicit runner `dynamicTools`.
- Documentation and AGENTS.md explain that same-user local binary deployments should preserve host-local Codex skills, MCP config, Apps/connectors, and plugin capability unless the operator explicitly restricts them.

## References

- Split from #1049.
- Follow-up #1056 owns the separate `multiAgentMode: none` audit.

zjlgdx:
Closes #1057

Part of #1049. Follow-up #1056 owns the separate `multiAgentMode: none` audit.

## Summary
- Remove the runner-owned `thread/start.config` override that disabled Codex Apps/connectors by default, so same-user local binary app-server runs preserve the operator's effective Codex config.
- Add `CODEX_HOME` to the agent subprocess baseline and update tests/docs so non-default Codex homes keep skills, MCP config, auth, sessions, and plugin/package metadata available.
- Update shipped real-Codex workflow templates to use `codex app-server --config shell_environment_policy.inherit=all` while keeping the core SPEC default as `codex app-server`.

Root cause: the prior unattended App-approval fix disabled an entire Codex capability surface for every local app-server thread instead of only failing closed on approval/input paths the worker cannot satisfy.

- [x] Adds one, justified by an upstream **Elixir reference** — cite the matching file and line from the openai/symphony Elixir tree in the body below (a bare module name does not count; the gate looks for a concrete source path).

Upstream reference checked: `elixir/lib/symphony_elixir/codex/app_server.ex:189` launches `codex.command` from the issue workspace without a runner-injected environment override, and `elixir/lib/symphony_elixir/codex/app_server.ex:280` starts `thread/start` with approval policy, sandbox, cwd, and dynamic tools rather than an Apps/connectors-disabling `config` payload. Upstream's real-Codex workflow also uses `elixir/WORKFLOW.md:33` with `shell_environment_policy.inherit=all`.

## Testing
- [x] `go test -run '^TestProductionGoFilesStayWithinSizeBudget$' -count=1 ./scripts` (covered by `go test -count=1 ./scripts`)
- [ ] `go vet ./...`
- [ ] `go test -race -covermode=atomic ./...`
- [ ] `gofmt -l` clean + blocking golangci gate (`staticcheck,unparam,unused,…`)

Additional validation run:

```bash
go test -count=1 ./internal/runner ./internal/workflow ./scripts ./cmd/worker ./internal/doctor
git diff --check
```

Not run: full `go vet ./...`, full race suite, and golangci gate.

## Notes
- `multiAgentMode: none` remains unchanged in this PR because it is not the Apps/connectors inheritance regression; #1056 tracks the follow-up evidence audit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -187,7 +187,7 @@ govern *how* we evaluate components inside the SPEC-aligned envelope.
 
 ## Cross-cutting checklist when porting from the Elixir reference
 
-The SPEC-alignment rules above govern *protocol contract*. Three
+The SPEC-alignment rules above govern *protocol contract*. Four
 classes of failure have shown up in shipped PRs anyway, because the
 SPEC text does not surface them. Use this checklist in addition to
 reading SPEC and the reference module — every item is tied to a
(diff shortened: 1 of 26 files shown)


zjlgdx:
@codex review

PR #1071, opened 2026-07-03
fix(tui): render codex percent rate windows

### LINKED ISSUE #1042: tui: render Codex percent rate-limit windows instead of n/a reset epochs
## Problem

The TUI rate-limit formatter still assumes bucket-style payloads with `remaining` / `limit`. Current Codex percent-window snapshots can carry `used_percent`, `window_duration_mins`, and `resets_at` instead. For that shape, the TUI loses the useful percent/window information and can render the absolute Unix reset timestamp as a relative seconds value.

That means a live Codex snapshot may show something like `primary n/a reset 1780501875s` rather than `primary 14% used · 5h window · resets in ...`.

## Current evidence

- `cmd/tui/render.go` `formatRateLimitBucket` only builds a useful base from `remaining` and/or `limit`; otherwise it leaves `base := "n/a"`.
- The reset key loop includes `resets_at`, but `formatResetValue` appends `s` to numeric values, so an absolute Unix timestamp is displayed as if it were a relative duration.
- Validation dashboard snapshots under `docs/validation/assets/state/*.json` contain Codex percent windows with `used_percent`, `window_duration_mins`, and `resets_at`.
- The dashboard has a percent-window path for this shape; the TUI does not.

## Expected behavior

The TUI should render Codex percent-shaped rate-limit windows as operator-readable windows, not as `n/a` buckets. For example:

- `used_percent: 14`, `window_duration_mins: 300`, `resets_at: <epoch>` -> `14% used · 5h window · resets in ...`
- A secondary `10080` minute window should be shown as a week/day-scale window, not a raw minute count or epoch seconds.
- Bucket-style `remaining` / `limit` payloads should still render without regression.

## Acceptance criteria

- `formatRateLimitBucket` detects percent-window payloads and displays `used_percent`.
- `window_duration_mins` is displayed in a human-readable unit.
- `resets_at` is treated as an absolute Unix timestamp; `reset_in_seconds` remains a relative duration.
- TUI tests cover both Codex percent windows and legacy bucket windows.
- Sanitization guarantees for untrusted string fields remain covered.

zjlgdx:
Closes #1042

## Summary

- Render Codex percent-window rate limit buckets in the TUI using `used_percent` / `usedPercent`.
- Show `window_duration_mins` / `windowDurationMins` as human-scale window labels.
- Treat `resets_at` / `resetsAt` as absolute Unix timestamps while preserving relative `reset_in_seconds` / `resets_in_seconds` handling for legacy payloads.

This is a TUI display-only fix over the existing state API rate-limit snapshot.

## Acceptance

- [x] Percent-window payloads display percent used instead of `n/a`.
- [x] Window duration minutes render as human-readable windows such as `5h window` and `7d window`.
- [x] Absolute reset epochs render as relative time-to-reset, not raw seconds.
- [x] Relative reset fields remain relative.
- [x] Legacy remaining/limit buckets still render.
- [x] Existing sanitization coverage remains in place.

## TDD and mutation evidence

- RED: `TestFormatRateLimits_RendersCodexPercentWindows` initially failed against the old `n/a reset <epoch>s` rendering.
- RED: after the first fix, the same test was switched to the current Codex camelCase schema shape and failed until camelCase aliases were handled.
- Mutation: breaking `used_percent` detection made the percent-window test fail.
- Mutation: treating `resets_at` as legacy reset seconds made the raw epoch assertion fail.
- Mutation: breaking `usedPercent` alias handling made the camelCase test fail.
- Mutation: removing `resets_in_seconds` alias handling made the relative reset assertion fail.

## Review fallback

- Claude / Claude Code unavailable by environment contract; no `claude` command used.
- Subagent review path is unavailable in this session, so independent review used Codex CLI plus manual diff/schema review.
- Codex CLI first review found a real schema issue: the current vendored Codex app-server schema uses camelCase fields such as `limitName`, `usedPercent`, `windowDurationMins`, and `resetsAt`; the fix was amended to cover both camelCase and the observed snake_case payload names.
- Codex CLI second review on the final diff reported no findings.
- Manual review checked the TUI diff against state API passthrough behavior and the vendored Codex schema.

- [x] `within budget` — <=12 production files / <=300 production LOC (test files and generated code excluded).

Production files changed: 1.

## Testing

- [x] `go vet ./...`
- [x] `gofmt -l` clean + blocking golangci gate (`staticcheck,unparam,unused,...`)
- [x] `go mod tidy && git diff --exit-code -- go.mod go.sum`

### DIFF
--- a/cmd/tui/layout_test.go
+++ b/cmd/tui/layout_test.go
@@ -363,6 +363,62 @@ func TestFormatRateLimits_SanitizesUntrustedValues(t *testing.T) {
 	}
 }
 
+func TestFormatRateLimits_RendersCodexPercentWindows(t *testing.T) {
+	oldNow := rateLimitNow
+	rateLimitNow = func() time.Time { return time.Unix(1_780_000_000, 0) }
+	defer func() { rateLimitNow = oldNow }()
+
+	resetAt := rateLimitNow().Add(90 * time.Minute).Unix()
+	rl := map[string]interface{}{
+		"limitName": "codex",
+		"primary": map[string]interface{}{
+			"usedPercent":        float64(14),
+			"windowDurationMins": float64(300),
+			"resetsAt":           float64(resetAt),
+		},
+		"secondary": map[string]interface{}{
+			"used_percent":         float64(92.5),
+			"window_duration_mins": float64(10080),
+			"resets_in_seconds":    float64(30),
+		},
+	}
+	got := visibleString(formatRateLimits(rl))
+
+	for _, want := range []string{
+		"codex",
+		"primary 14% used · 5h window · resets in 1h 30m",
+		"secondary 92.5% used · 7d window · reset 30s",
+	} {
+		if !strings.Contains(got, want) {
+			t.Errorf("formatRateLimits(percent windows) = %q, want substring %q", got, want)
+		}
+	}
+	if rawEpoch := strconv.FormatInt(resetAt, 10) + "s"; strings.Contains(got, rawEpoch) {
+		t.Errorf("formatRateLimits(percent windows) rendered resets_at as relative seconds: %q", got)
+	}
+}
+
+func TestFormatRateLimits_RendersLegacyBuckets(t *testing.T) {
+	rl := map[string]interface{}{
+		"limit_id":  "legacy",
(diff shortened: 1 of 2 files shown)


zjlgdx:
@codex review

PR #1073, opened 2026-07-03
fix(orchestrator): avoid duplicate stall cancels

### LINKED ISSUE #1039: [P3][bug] orchestrator: stall reconcile re-records a 'stalled' Failed event and re-cancels on every tick until the worker exits
## Summary

`reconcileStalledRunsOp.apply` records a `Failed`/"stalled" runtime event and queues a cancel for every Running entry past the stall budget:

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/internal/orchestrator/actor_reconcile.go#L257-L270

```go
for id, run := range st.Running {
    if run.BudgetExceeded {
        continue
    }
    if !r.isStalled(run) {
        continue
    }
    canceled = append(canceled, run)
    st.RecordEvent(RuntimeEvent{Kind: RuntimeEventFailed, IssueID: id, Identifier: run.Identifier, Message: "stalled"})
}
```

The cancel happens in the off-actor followup, but the entry stays in `st.Running` until `finalizeRunOp` lands — and `LastEventAt` does not advance for a stalled run (that's the definition of stalled). So while the canceled worker is still winding down (up to `WorkerExitTimeout`, or longer if the subprocess ignores the kill), **every subsequent poll tick re-matches the same entry**: another `Failed`/"stalled" event is recorded and `CancelWorker` is invoked again.

Note the `BudgetExceeded` guard above shows the intended pattern — budget reconcile marks the entry so it is not re-processed — but stall reconcile has no equivalent "already canceled for stall" mark.

## Impact

Duplicate `Failed` events pollute `/api/v1/state`, the TUI event feed, and any alerting keyed on Failed counts; one genuinely stuck run with a slow exit produces one spurious failure event per poll tick.

## Suggested direction

Mirror the budget path: set a `StallCanceled` (or reuse-style) flag on the entry when first canceled and skip flagged entries on later ticks, or only record the event when `CancelWorker` is actually about to fire for the first time.

zjlgdx:
Closes #1039

## Summary
- Make stall reconciliation cancel and record the `Failed`/`stalled` event only once for a running entry before `finalizeRunOp` removes it.
- Preserve retry behavior by keeping the latch on the live `RunningEntry`, so a later retry starts with a fresh unstalled entry.

## Acceptance
- A stalled run queues exactly one cancel while it remains in `state.running`.
- A stalled run records exactly one `RuntimeEventFailed` event with message `stalled` before finalize.
- Active/non-stalled runs and existing stall timeout behavior remain unchanged.
- Sibling sweep: budget-exceeded reconciliation already has one-shot event semantics and intentionally retries cancel for marked budget entries via existing tests, so that independent policy was left out of scope.

SPEC/Elixir evidence:
- SPEC §8.5 says stall reconciliation terminates the worker and queues retry when elapsed time exceeds `codex.stall_timeout_ms`.
- SPEC §16.3 runs stall reconciliation as part of `reconcile_running_issues`.
- `elixir/lib/symphony_elixir/orchestrator.ex:601-627` restarts a stalled issue by terminating the running issue and scheduling retry.
- `elixir/lib/symphony_elixir/orchestrator.ex:546-568` removes the issue from `state.running` during termination, so the same stalled run cannot be re-processed on later ticks. The Go latch preserves that one-shot behavior while cancellation/finalize is off-actor.

- [x] `within budget` — ≤12 production files / ≤300 production LOC (test files and generated code excluded). Production files changed: 2.

Do not collapse formatting, merge responsibilities, delete useful tests, or
reduce readability solely to check `within budget`; choose
`size-gated: justified overage` when the smallest readable change plus necessary
tests needs the space.

## TDD and mutation
- RED: added `TestReconcileStalledRunsOpCancelsStalledRunOnceUntilFinalize`; before the fix it failed because the second reconcile queued a duplicate cancel.
- Mutation 1: removed `|| run.StallCanceled`; the new test failed on the duplicate second cancel.
- Mutation 2: removed `run.StallCanceled = true`; the new test failed on the duplicate second cancel.
- Restored both changes, reran focused tests, and confirmed touched-file diff clean against `HEAD` after mutation restore.

## Review fallback
- Claude Code is unavailable in this environment, so no Claude/Claude Code command was run.
- Fallback review used Codex CLI on the PR diff, manual SPEC/Elixir comparison, local diff review, and the full Go gate below.
- Codex CLI independent review result: findings none. It ran the targeted orchestrator stall tests and `gofmt -l` on touched files.

## Testing
- [x] `go vet ./...`

Additional validation:
- `go test ./internal/orchestrator -run 'TestReconcileStalledRuns(Op|Cancels|Leaves|Skips)' -count=1`
- `go test ./internal/orchestrator -count=1`

## Notes

### DIFF
--- a/internal/orchestrator/actor_reconcile.go
+++ b/internal/orchestrator/actor_reconcile.go
@@ -257,12 +257,13 @@ type reconcileStalledRunsOp struct {
 func (r *reconcileStalledRunsOp) apply(st *OrchestratorState) func() {
 	var canceled []*RunningEntry
 	for id, run := range st.Running {
-		if run.BudgetExceeded {
+		if run.BudgetExceeded || run.StallCanceled {
 			continue
 		}
 		if !r.isStalled(run) {
 			continue
 		}
+		run.StallCanceled = true
 		canceled = append(canceled, run)
 		st.RecordEvent(RuntimeEvent{Kind: RuntimeEventFailed, IssueID: id, Identifier: run.Identifier, Message: "stalled"})
 	}
--- a/internal/orchestrator/actor_reconcile_ops_test.go
+++ b/internal/orchestrator/actor_reconcile_ops_test.go
@@ -99,6 +99,49 @@ func TestReconcileStalledRunsOpTimingReference(t *testing.T) {
 	}
 }
 
+func TestReconcileStalledRunsOpCancelsStalledRunOnceUntilFinalize(t *testing.T) {
+	now := time.Now()
+	st := NewOrchestratorState(15000, 10)
+	cancelCount := 0
+	run := &RunningEntry{
+		Identifier:  "S",
+		LastEventAt: now.Add(-time.Hour),
+		StartedAt:   now.Add(-2 * time.Hour),
+		CancelWorker: func(error) {
+			cancelCount++
+		},
+	}
+	st.Running[IssueID("S")] = run
+
+	firstResult := make(chan []*RunningEntry, 1)
+	first := &reconcileStalledRunsOp{timeout: time.Minute, now: now, result: firstResult}
+	first.apply(st)()
+	firstCanceled := <-firstResult
+	if len(firstCanceled) != 1 || firstCanceled[0] != run {
+		t.Fatalf("first stalled reconcile canceled = %+v, want the running entry", firstCanceled)
+	}
+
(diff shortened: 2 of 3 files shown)


zjlgdx:
@codex review

PR #1074, opened 2026-07-03
fix(tracker): cap success JSON decoding

### LINKED ISSUE #1038: [P3][hardening] tracker: success-path JSON decoding is unbounded across Linear/GitHub/Gitea clients
## Summary

`DrainAndClose` caps what gets *drained* on close (8 KiB), but every 2xx success path hands `resp.Body` straight to `json.NewDecoder(...).Decode(...)` with no size ceiling:

- `internal/tracker/linear.go:705` (GraphQL responses)
- `internal/tracker/github_listing.go:219`, `internal/tracker/github_refresh.go:220`, `internal/tracker/github_blockers.go` (REST + GraphQL)
- `internal/gitea/tracker_client_listing.go:253`, `internal/gitea/tracker_client_refresh.go` (REST)

Example:

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/internal/tracker/github_refresh.go#L219-L222

A misbehaving or compromised tracker endpoint (self-hosted Gitea behind a broken proxy, a MITM'd `base_url`) can stream an arbitrarily large body and the worker will buffer/parse it until OOM. The runner's stdio path already treats this class explicitly (`maxAppServerLineBytes` caps codex lines at 10 MB per SPEC §10.1); the tracker HTTP boundary is the sibling input boundary without a cap.

## Suggested direction

Wrap each decode in `io.LimitReader(resp.Body, N)` (a shared constant, e.g. 16–32 MiB, generous for a 250-issue page) and surface an explicit "response exceeded limit" error. One helper in the shared `internal/tracker` plumbing keeps it a single source of truth for all three adapters.

zjlgdx:
Closes #1038

## Summary
- Add a shared bounded JSON decoder for successful tracker/Linear HTTP responses, returning `ErrJSONResponseTooLarge` when a declared or streamed body exceeds 16 MiB.
- Route Linear, GitHub, Gitea, runner workflow-state lookup, and doctor Linear probe success decodes through the shared helper while preserving existing status/rate-limit/drain behavior.
- Add seam tests for every replaced success decode path plus a no-`Content-Length` streaming-body test.

## Acceptance
- Linear/GitHub/Gitea success JSON decoding has a shared size cap and clear typed error.
- Misbehaving endpoints with oversized `Content-Length` fail before decoding, and chunked/unknown-length streams are bounded at `limit+1`.
- Existing response lifecycle remains intact: call sites still `DrainAndClose`/close via their pre-existing deferred cleanup.
- Sibling sweep: Codex review found runner workflow-state lookup and doctor Linear probe shared the same Linear success-response class, so they were fixed in this PR. The remaining `cmd/tui/state_client.go` decoder reads the local state API, not an external tracker/Linear/GitHub/Gitea boundary, so it is out of scope.

SPEC/Elixir evidence:
- SPEC §11.1/§11.2 define tracker read operations, Linear GraphQL transport, pagination, and the 30s network timeout; this PR keeps those operations and transport semantics unchanged.
- SPEC §11.4 says tracker errors are logged/skipped/kept running by callers; this PR only classifies an oversized success body as a decode/payload error at the existing boundary.
- SPEC §17 says implementations should not assume tracker data or other inputs are fully trustworthy; this is Go-side input-boundary hardening, not a new Symphony phase.
- `elixir/lib/symphony_elixir/linear/client.ex:164-172` returns the decoded successful Linear body from `Req.post`, and `elixir/lib/symphony_elixir/linear/client.ex:398-405` issues the JSON request with a 30s connect timeout. The Go port still follows the same request/decode flow, with an explicit response-size cap because Go streams `resp.Body` directly.

- [x] `within budget` — ≤12 production files / ≤300 production LOC (test files and generated code excluded). Production files changed: 9.

Do not collapse formatting, merge responsibilities, delete useful tests, or
reduce readability solely to check `within budget`; choose
`size-gated: justified overage` when the smallest readable change plus necessary
tests needs the space.

## TDD and mutation
- RED: oversized success-body seam tests failed for Linear, GitHub, Gitea, runner workflow-state lookup, and doctor Linear probe before each path was wired to the shared decoder.
- Mutation 1: removed the `ContentLength > maxJSONResponseBytes` guard; all oversized `Content-Length` seam tests failed across tracker/gitea/runner/doctor.
- Mutation 2: changed the streaming read from `maxJSONResponseBytes+1` to `maxJSONResponseBytes`; `TestDecodeJSONResponseBoundsStreamingBodyWithoutContentLength` failed with a normal JSON error instead of `ErrJSONResponseTooLarge`.
- Restored mutations from `HEAD`, reran focused tests, and confirmed `internal/tracker/json_decode.go` was clean against `HEAD`.

## Review fallback
- Claude Code is unavailable in this environment, so no Claude/Claude Code command was run.
- Fallback review used Codex CLI on the PR diff, manual SPEC/Elixir comparison, local diff review, sibling sweep, and the full Go gate below.
- Codex CLI review pass 1 found two P2 gaps: runner workflow-state lookup and doctor Linear probe still decoded successful Linear responses directly. Both were fixed.
- Codex CLI review pass 2 result: PASS, findings none.

## Testing
- [x] `go vet ./...`

Additional validation:
- `go test ./internal/tracker ./internal/gitea ./internal/runner ./internal/doctor -run 'Test(LinearGraphQLRejectsOversizedSuccessBody|GitHub.*RejectsOversizedSuccessBody|DecodeJSONResponseBoundsStreamingBodyWithoutContentLength|TrackerClient.*RejectsOversizedSuccessBody|LookupWorkflowStateIDsRejectsOversizedSuccessBody|DecodeLinearProjectProbeRejectsOversizedSuccessBody)' -count=1`
- `go test ./internal/tracker ./internal/gitea -count=1`
- `go test ./internal/runner ./internal/doctor -run 'Test(LookupWorkflowStateIDsRejectsOversizedSuccessBody|LookupWorkflowStateIDsDrainsResponseBodies|DecodeLinearProjectProbeRejectsOversizedSuccessBody|DecodeLinearProjectProbeDrainsResponseBodies|ProbeLinearProjectSlugMasksEndpointUserinfoInErrors)' -count=1`

## Notes

### DIFF
--- a/internal/doctor/doctor_tracker.go
+++ b/internal/doctor/doctor_tracker.go
@@ -297,8 +297,8 @@ func decodeLinearProjectProbe(resp *http.Response, out any) error {
 	// Preflight mirrors the consumer: the poller decodes this response
 	// immediately, so name the non-JSON body (a login proxy answering 200
 	// HTML) instead of surfacing a bare json.SyntaxError.
-	if err := json.NewDecoder(resp.Body).Decode(out); err != nil {
-		return fmt.Errorf("linear returned %s with a body that is not JSON: %w", resp.Status, err)
+	if err := tracker.DecodeJSONResponse(resp, out); err != nil {
+		return fmt.Errorf("linear returned %s with a body that is not JSON or exceeded maximum size: %w", resp.Status, err)
 	}
 	return nil
 }
--- a/internal/doctor/doctor_tracker_drain_test.go
+++ b/internal/doctor/doctor_tracker_drain_test.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"strings"
 	"testing"
+
+	"github.com/xrf9268-hue/aiops-platform/internal/tracker"
 )
 
 // drainRecordingBody records whether the response body was read to EOF (the
@@ -86,6 +88,31 @@ func TestDecodeLinearProjectProbeDrainsResponseBodies(t *testing.T) {
 				t.Fatalf("status %d body sawEOF = %t, closed = %t; want true, true (undrained bodies break connection reuse)", tc.status, body.sawEOF, body.closed)
 			}
 		})
+	}
+}
+
+func TestDecodeLinearProjectProbeRejectsOversizedSuccessBody(t *testing.T) {
+	body := &drainRecordingBody{reader: strings.NewReader(`{"data":{"projects":{"nodes":[]}}}`)}
+	resp := &http.Response{
+		Status:        "200 OK",
+		StatusCode:    http.StatusOK,
+		Body:          body,
+		ContentLength: 1 << 62,
+	}
+	var out struct {
+		Data struct {
+			Projects struct {
+				Nodes []struct {
+					ID string `json:"id"`
+				} `json:"nodes"`
+			} `json:"projects"`
+		} `json:"data"`
+	}
+
+	err := decodeLinearProjectProbe(resp, &out)
+
+	if !errors.Is(err, tracker.ErrJSONResponseTooLarge) {
+		t.Fatalf("decodeLinearProjectProbe() oversized success error = %v; want errors.Is(err, tracker.ErrJSONResponseTooLarge)", err)
 	}
 }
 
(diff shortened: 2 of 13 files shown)


zjlgdx:
@codex review

PR #1081, opened 2026-07-03
fix(orchestrator): recover dispatch spawn panics

### LINKED ISSUE #1037: [P3][bug] orchestrator: sleepContext leaks a timer per interrupted sleep; dispatchOp followup panic strands RequestDispatch callers
Two small, related actor-plumbing defects found in the same review; filing together because both are single-function fixes in the orchestrator package.

## 1. `sleepContext` uses `time.After`, leaking the timer on early wake

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/internal/orchestrator/workflow_runtime.go#L367-L374

```go
func sleepContext(ctx context.Context, d time.Duration) error {
    select {
    case <-ctx.Done():
        return ctx.Err()
    case <-time.After(d):
        return nil
    }
}
```

When `ctx` wins the race, the `time.After` timer lives until `d` expires and cannot be stopped. This helper backs every poll-loop and workflow-reload-loop sleep, and `sleepOrRetryWake` interrupts it on every retry wake — a long-running worker with frequent wakes accumulates pending timers. The repo already uses the correct pattern elsewhere (`retryReconciledWorkspaceCleanup` uses `time.NewTimer` + `defer Stop()`); AGENTS.md's timer rule ("store the timer and Stop() it") applies. Fix: `time.NewTimer(d)` + `defer t.Stop()`.

## 2. `dispatchOp` followup sends the reply only after `spawn` returns

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/internal/orchestrator/actor_dispatch.go#L135-L138

```go
return func() {
    o.spawn(id, issue, attempt, continuationAttempt, continuationTurnCount, cleanTurnBudget)
    result <- nil
}
```

Followups run under `safeGo`, whose `recoverPanic` swallows a panic in `spawn` — and then `result <- nil` never executes. `RequestDispatch` blocks on `<-reply` with only `ctx.Done()` as the escape (actor_dispatch.go:41-46); the poller calls it with the long-lived poll context, so a single panic inside `spawn` silently wedges that dispatch call for the process lifetime while the issue stays `Claimed` forever (the claim was reserved in `apply`). Fix: send the reply via `defer` (or before spawning, since the claim decision is already made), so panic containment cannot strand the caller.

zjlgdx:
Closes #1037

## Summary
- Replace `sleepContext`'s `time.After` wait with a stoppable `time.NewTimer` so interrupted poll/reload sleeps do not leave timers alive until the full interval elapses.
- Harden dispatch spawn panics so `RequestDispatch` always receives its reply, `WaitForWorkers` is not stranded, and recovered spawn failures flow into the existing failure-retry state path while preserving retry workspaces.

SPEC / upstream notes:
- This is Go runtime-guarantee compensation, not a new Symphony phase or deviation.
- SPEC describes retry timers as cancelable handles and dispatch failures as retryable; upstream Elixir cancels existing timers in `elixir/lib/symphony_elixir/orchestrator.ex` (`schedule_tick`, `schedule_issue_retry`) and routes spawn child errors back into retry handling.

Production diff: 3 files, ~129 added / 9 removed production LOC.

## Testing
- [x] `go vet ./...`

Additional validation:
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --config=.golangci.yml --issues-exit-code=0` → `0 issues.`
- `go test ./internal/orchestrator -count=1`

Mutation checks:
- Disabled dispatch followup reply while keeping the code compiling; `TestRequestDispatchReturnsWhenSpawnPanics` failed with `context deadline exceeded`.
- Disabled `spawnPanicRecovery`; `TestRequestDispatchReturnsWhenSpawnPanics` failed because `WaitForWorkers` did not drain.
- Disabled claim-level spawn-panic retry scheduling; `TestRecoverSpawnPanicSchedulesFailureRetryWithoutRunning` failed waiting for the failure retry.
- Dropped retry workspace threading on retry redispatch; `TestRetryDispatchSpawnPanicPreservesWorkspace` failed because the recovered retry lost its workspace path.
- Reverted `sleepContext` to `time.After`; `TestSleepContextUsesStoppableTimer` failed.

Review fallback:
- Claude Code is unavailable in this environment, so the double-review fallback used Codex CLI diff review, manual local diff review, SPEC/Elixir comparison, GitHub `@codex review`, and the Go gates above.
- Local Codex CLI diff review final pass on head `2226a24d`: `PASS` with no source-confirmed correctness, race, leak, or test-quality findings.
- GitHub `@codex review` on the previous head found a retry-workspace preservation gap; the final head threads the workspace through dispatch spawn recovery and retry redispatch. GitHub `@codex review` on head `2226a24d28` then reported no major issues.

## Notes

### DIFF
--- a/internal/orchestrator/actor_dispatch.go
+++ b/internal/orchestrator/actor_dispatch.go
@@ -7,6 +7,7 @@ package orchestrator
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/xrf9268-hue/aiops-platform/internal/tracker"
@@ -116,6 +117,10 @@ func (d *dispatchOp) apply(st *OrchestratorState) func() {
 		d.result <- ErrCapacityFull
 		return nil
 	}
+	workspace := Workspace{}
+	if consumedContinuation != nil {
+		workspace = consumedContinuation.Workspace
+	}
 	consumeContinuationRetry(st, id, consumedContinuation)
 	st.RecordEvent(RuntimeEvent{
 		Kind:       RuntimeEventCandidate,
@@ -133,8 +138,8 @@ func (d *dispatchOp) apply(st *OrchestratorState) func() {
 	attempt := d.attempt
 	result := d.result
 	return func() {
-		o.spawn(id, issue, attempt, continuationAttempt, continuationTurnCount, cleanTurnBudget)
-		result <- nil
+		defer func() { result <- nil }()
+		o.spawn(id, issue, attempt, continuationAttempt, continuationTurnCount, cleanTurnBudget, workspace)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


zjlgdx:
@codex review

zjlgdx:
@codex review

PR #1082, opened 2026-07-03
fix(gitea): reuse default tracker HTTP client

### LINKED ISSUE #1036: [P3][bug] gitea: default httpClient() allocates a new http.Client per request, defeating connection reuse
## Summary

When `TrackerClient.HTTP` is nil (the production default — `NewTrackerClient` never sets it), every API call constructs a fresh `http.Client`:

https://github.com/xrf9268-hue/aiops-platform/blob/b589dca5cc47137ec72ae6146b79968f14e1e4b8/internal/gitea/tracker_client.go#L75-L80

```go
func (c *TrackerClient) httpClient() *http.Client {
    if c != nil && c.HTTP != nil {
        return c.HTTP
    }
    return &http.Client{Timeout: c.requestTimeout()}
}
```

Each client uses `http.DefaultTransport` under the hood, so idle-connection pooling still nominally works — but a per-call `http.Client` prevents any client-level tuning, and the pattern diverges from the Linear/GitHub adapters which default to a shared `http.DefaultClient`. On a Gitea instance behind TLS with keepalive quirks, paginated listings plus per-issue `Depends on #N` blocker lookups can open many short-lived connections per tick.

## Suggested direction

Initialize a single `http.Client{Timeout: ...}` in `NewTrackerClient` (or lazily with `sync.Once`) so all requests share one client, matching the sibling adapters. One-line class fix plus a test asserting the same `*http.Client` is returned across calls.

zjlgdx:
Closes #1036

## Summary
- Cache the default Gitea tracker `*http.Client` behind `sync.Once` so production requests reuse one client instead of allocating a fresh client per API call.
- Keep explicit `TrackerClient.HTTP` injection as the highest-priority path, and keep the default client timeout wired through `requestTimeout()`.

## Technical conclusion
- Problem: `TrackerClient.HTTP == nil` caused `httpClient()` to allocate a new `http.Client` for each Gitea request.
- Boundary: this is tracker adapter transport hygiene, not a new tracker operation, worker/orchestrator phase, cache, state machine, or tracker write.
- Minimum fix: reuse one default client per `TrackerClient`; keep explicit test/caller injection unchanged.
- Non-goals: no changes to Gitea pagination, blocker lookup semantics, tracker API shape, or request timeout policy.

SPEC / upstream notes:
- SPEC §11 requires tracker reader operations and normalized outputs; it does not require per-request HTTP client allocation.
- Upstream Elixir's Linear client issues requests through the shared `Req` stack (`elixir/lib/symphony_elixir/linear/client.ex`), so this change keeps Gitea adapter transport behavior inside the existing tracker boundary.

Production diff: 1 file, ~10 added / 1 removed production LOC.

## Testing
- [x] `go vet ./...`

Additional validation:
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --config=.golangci.yml --issues-exit-code=0` → `0 issues.`
- `go test ./internal/gitea -count=1`
- `go test ./internal/gitea -run 'TestTrackerClient(DefaultHTTPClientIsReused|InjectedHTTPClientWins|FallbackHTTPClientUsesRequestTimeout)' -count=1`

Mutation checks:
- Replaced the cached default client return with a fresh `&http.Client{Timeout: c.requestTimeout()}`; `TestTrackerClientDefaultHTTPClientIsReused` failed on pointer mismatch.
- Disabled the explicit `HTTP` injection branch; `TestTrackerClientInjectedHTTPClientWins` failed because `httpClient()` did not return the injected client.

Review fallback:
- Claude Code is unavailable in this environment, so the double-review fallback used Codex CLI diff review, manual local diff review, SPEC/Elixir comparison, sibling sweep, GitHub `@codex review`, and the Go gates above.
- Local Codex CLI diff review on head `7c20c22`: `PASS` with no source-confirmed correctness, race, leak, or test-quality findings. Its read-only sandbox could not run live `gh`/Go cache checks, so the main session performed those gates directly.
- GitHub `@codex review` on head `7c20c22372` reported no major issues.

## Notes

### DIFF
--- a/internal/gitea/tracker_client.go
+++ b/internal/gitea/tracker_client.go
@@ -52,6 +52,9 @@ type TrackerClient struct {
 
 	issueNumbers sync.Map
 
+	defaultHTTPOnce sync.Once
+	defaultHTTP     *http.Client
+
 	paginationCapHits atomic.Int64
 }
 
@@ -76,7 +79,13 @@ func (c *TrackerClient) httpClient() *http.Client {
 	if c != nil && c.HTTP != nil {
 		return c.HTTP
 	}
-	return &http.Client{Timeout: c.requestTimeout()}
+	if c == nil {
+		return &http.Client{Timeout: defaultGiteaRequestTimeout}
+	}
+	c.defaultHTTPOnce.Do(func() {
+		c.defaultHTTP = &http.Client{Timeout: c.requestTimeout()}
+	})
+	return c.defaultHTTP
 }
 
 // PaginationCapHits returns how often this client observed more than
--- a/internal/gitea/tracker_client_timeout_test.go
+++ b/internal/gitea/tracker_client_timeout_test.go
@@ -38,6 +38,33 @@ func TestTrackerClientFallbackHTTPClientUsesRequestTimeout(t *testing.T) {
 	}
 }
 
+func TestTrackerClientDefaultHTTPClientIsReused(t *testing.T) {
+	c := NewTrackerClient(workflow.TrackerConfig{APIKey: "secret"}, "https://gitea.example", "owner", "repo")
+
+	first := c.httpClient()
+	second := c.httpClient()
+
+	if first == nil || second == nil {
+		t.Fatalf("default HTTP clients = %p, %p; want non-nil", first, second)
+	}
+	if first != second {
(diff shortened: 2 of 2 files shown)


zjlgdx:
@codex review

PR #1103, opened 2026-07-15
fix(orchestrator): make token usage folding exact

### LINKED ISSUE #1098: fix(orchestrator): make Codex token folding obey absolute-total precedence
## Problem

SPEC §13.5 requires absolute thread totals to take precedence, with increments computed against the last reported absolute totals. The current fold violates that contract in two ways:

- internal/orchestrator/runtime_events.go accepts a generic usage map on runtime events that do not define it as a per-event delta.
- applyTokenUsage stores LastReported*Tokens, but subtracts accumulated Codex*Tokens instead.

A concrete sequence is therefore wrong today:

1. absolute 100 / 20 / 120;
2. unrelated notification carrying generic usage 5 / 1 / 6;
3. the same absolute 100 / 20 / 120.

The run remains at 105 / 21 / 126 instead of converging to 100 / 20 / 120. The #1089 benchmark also observed public input + output totals exceeding provider-reported total, so token comparisons and budget enforcement cannot be treated as exact until this seam is fixed.

Follow-up to #1027 and evidence from #1089; neither issue fixed this folding defect.

## Acceptance criteria

- [ ] Selected absolute shapes (thread/tokenUsage/updated, token_usage.total, and their supported aliases) are authoritative for the session.
- [ ] Repeating an absolute snapshot contributes zero; a larger snapshot contributes only the increase from the prior absolute report.
- [ ] last_token_usage / token_usage.last is never added to running, process, or retained-session totals.
- [ ] A generic usage map on notification or other message events is ignored; only event types explicitly defined as per-event deltas may use it.
- [ ] Once an absolute total has been observed, later generic/delta payloads cannot double-count the same tokens.
- [ ] LastReported*Tokens is the actual absolute baseline, or is removed in favor of one equivalent source of truth.
- [ ] Running rows, process totals, retained session usage, and token-budget checks all consume the same corrected counters.
- [ ] Provider-reported total_tokens remains authoritative; do not force total = input + output or rewrite historical evidence.
- [ ] Add regression tests for the concrete failing sequence, repeated/increasing absolute snapshots, and token_usage.total versus token_usage.last.
- [ ] Mutation proof: removing the absolute-baseline guard makes a behavior assertion fail.

## Non-goals

- No cached-input or reasoning-output API fields.
- No nested-subagent or external @codex review accounting.
- No billing estimator, dashboard redesign, Trace L4, or new metrics subsystem.

## Local verification

    go test ./internal/orchestrator -run 'TestRecordRuntimeEvent.*(Token|Usage)' -count=1
    go test -race ./internal/orchestrator
    go mod tidy && git diff --exit-code -- go.mod go.sum
    go vet ./...

This issue is fully locally verifiable; it requires no live tracker or forge mutation.

zjlgdx:
Closes #1098

## Summary
- Treat supported Codex thread token totals as authoritative absolute snapshots and fold only monotonic increases.
- Ignore generic usage carried by unrelated runtime events and never count `last_token_usage`.
- Preserve an explicitly reported provider `total_tokens`; input/output-only snapshots no longer invent a total.

## Testing
- [x] `go vet ./...`
- [x] `go test ./internal/orchestrator -run 'TestRecordRuntimeEvent.*(Token|Usage)|TestApplyTokenUsage' -count=1`

## Notes
- The new missing-total regression failed before the fix with an invented total of 140 instead of the provider baseline 120.
- Independent Standards and Spec review passed after the provider-total correction.

### DIFF
--- a/internal/orchestrator/actor_test.go
+++ b/internal/orchestrator/actor_test.go
@@ -1326,7 +1326,7 @@ func TestRecordRuntimeEventTracksLastEventAndMessage(t *testing.T) {
 	}
 }
 
-func TestRecordRuntimeEventTreatsGenericUsageAsEventDelta(t *testing.T) {
+func TestRecordRuntimeEventTreatsTurnCompletedUsageAsAbsolute(t *testing.T) {
 	o, issueID, cancel := startRuntimeEventActor(t, "ENG-USAGE")
 	defer cancel()
 
@@ -1342,8 +1342,8 @@ func TestRecordRuntimeEventTreatsGenericUsageAsEventDelta(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Snapshot: %v", err)
 	}
-	if view.CodexTotals.TotalTokens != 10 {
-		t.Fatalf("CodexTotals.TotalTokens = %d, want generic usage counted as per-event delta", view.CodexTotals.TotalTokens)
+	if view.CodexTotals.TotalTokens != 5 {
+		t.Fatalf("CodexTotals.TotalTokens = %d, want repeated absolute turn usage folded once", view.CodexTotals.TotalTokens)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


zjlgdx:
@codex review

PR #1132, opened 2026-07-17
fix(runner): restore schema-aligned continuation

### LINKED ISSUE #1129: fix(runner): restore schema-aligned in-session continuation
## Problem Statement

The Codex app-server runner does not perform the in-session continuation loop required by Symphony SPEC §7.1/§16.5 with Codex 0.144.4.

The production path refreshes the issue after each turn, but then still returns the agent-derived `turn/completed.params.continue` value. Codex 0.144.4's vendored `TurnCompletedNotification` schema contains only `threadId` and `turn`; the field `continue` does not exist. The live #1128 container run received that schema-shaped notification without `continue`, so every normal turn exited the worker session and consumed a new outer continuation claim.

This is the remaining gap from #218 / PR #323: tracker refresh was added, but active tracker state did not become the authoritative positive continuation decision.

## Solution

When the configured tracker refresh succeeds:

- continue on the same app-server thread/session while the refreshed issue remains active and the in-session turn budget remains;
- stop cleanly when the issue is no longer active;
- stop cleanly, rather than fail, when `agent.max_turns` is exhausted so the existing orchestrator continuation retry can take over.

Remove the production dependency on the schema-invalid `params.continue` signal and replace synthetic continuation fixtures with current-schema notifications.

Do not change the orchestrator's normal one-second continuation retry.

## User Stories

1. As an operator, I want an active issue to receive multiple turns in one worker session, so that `agent.max_turns` works as documented.
2. As an operator, I want a canceled or otherwise inactive issue to stop after the current turn, so that tracker state remains authoritative.
3. As an operator, I want max-turn exhaustion to return control normally, so that the existing continuation retry remains the outer-loop fallback.
4. As a maintainer, I want tests to use the vendored Codex protocol shape, so that removed wire fields cannot silently keep production behavior dead.

## Implementation Decisions

- Keep the fix inside the Codex app-server runner and its tests.
- With a tracker refresher, refreshed active state is sufficient to start the next turn; refreshed inactive state stops cleanly.
- Without a tracker refresher, retain a single-turn fallback for non-production/test callers rather than inventing tracker state.
- Remove parsing and tests for `turn/completed.params.continue`; this project is pre-release and does not carry a compatibility shim for a field absent from the pinned schema.
- Reuse the existing app-server process, thread ID, continuation prompt, tracker refresher, turn limits, and outer retry.
- Reaching `agent.max_turns` is a normal handback to the orchestrator, matching the Elixir reference.
- Do not add a new state, queue, retry policy, protocol adapter, or worker phase.

## Testing Decisions

- Add a high-seam fake app-server test whose `turn/completed` notifications contain only `threadId` and `turn`; while tracker state stays active, assert multiple `turn/start` requests use the same thread/session.
- Assert an inactive refresh stops after the current turn with no extra `turn/start`.
- Assert an active issue at `agent.max_turns` exits cleanly after exactly the configured count.
- Replace synthetic `continue=true/false` fixtures that contradict the pinned schema.
- Mutation-check the active-state continuation seam so returning false after an active refresh fails behavior assertions.
- Run targeted runner tests, the schema contract test, race tests for the package, and the repository's normal local gates.

## Out of Scope

- No orchestrator, retry-delay, state API, tracker adapter, GitHub workflow, validation controller, or handoff-detection changes.
- No increase to #1128's frozen four-claim limit and no repair/resume of its stopped repository.
- No parsing of PR comments/reviews inside the orchestrator.
- No new compatibility layer for historical Codex notification fields.

## Further Notes

Primary evidence:

- #1128 failed closed at four completed claims plus a fifth pending continuation.
- Symphony SPEC §7.1 and §16.5 require active issues to continue on the same live thread up to `agent.max_turns`.
- The upstream Elixir `AgentRunner` continues solely from refreshed tracker state and returns `:ok` at max turns.
- The pinned Codex 0.144.4 schema has no `continue` field in `TurnCompletedNotification`.
- The late external review in the stopped #1128 run cannot repair or invalidate the terminal result.

Refs #1128

zjlgdx:
Closes #1129

## Summary
- Make refreshed tracker state the sole in-session continuation authority: active continues on the existing app-server thread; inactive stops cleanly and preserves the exit snapshot.
- Remove the schema-invalid `turn/completed.params.continue` shim and replace related fixtures with the Codex 0.144.4 `threadId` + `turn` notification shape.
- Treat `agent.max_turns` and the smaller `CleanTurnBudget` as clean handback limits; callers without a refresher keep a single-turn fallback.

This restores the loop in SPEC §7.1/§16.5 and matches `openai/symphony/elixir/lib/symphony_elixir/agent_runner.ex`, where tracker refresh drives continuation and max-turn exhaustion returns `:ok`.

## Testing
- [x] `go test -run ^TestProductionGoFilesStayWithinSizeBudget -count=1 ./scripts`
- [x] `go vet ./...`

Additional validation:
- `go test -race ./internal/runner -count=1`
- `go test -tags e2e -race -timeout 15m ./test/e2e/...`
- committed-artifact mutation: active refresh changed to stop → high-seam test failed at one refresh; restored commit → test passed
- fixed-base Standards review: CLEAN; fixed-base Spec review: CLEAN
- `codex review --commit a66320f77d964d5850a14903033a7060852d3c6e`: no actionable defects

## Notes
- Production behavior changes remain in `internal/runner`; worker/orchestrator hunks only correct stale comments about the removed fallback.
- No new state, retry policy, adapter, queue, phase, or outer-loop change.

### DIFF
--- a/internal/orchestrator/runtime_poller_test.go
+++ b/internal/orchestrator/runtime_poller_test.go
@@ -336,8 +336,8 @@ func TestRuntimeDispatcherIssueStateRefresherPassesIssueIdentifierFallback(t *te
 
 // TestRuntimeDispatcherConfigForSnapshotReturnsNilWithoutRefresher pins the
 // fallback: without SetIssueStateRefresher the dispatcher leaves
-// IssueStateRefresher unset so RunTask keeps the legacy continueRun-only
-// path (e.g. tests, mock runners).
+// IssueStateRefresher unset so RunTask uses the runner's single-turn fallback
+// (e.g. tests, mock runners).
 func TestRuntimeDispatcherConfigForSnapshotReturnsNilWithoutRefresher(t *testing.T) {
 	d := &RuntimeDispatcher{baseConfig: worker.Config{}}
 	snap := WorkflowSnapshot{Workflow: &workflow.Workflow{Config: workflow.Config{
@@ -470,7 +470,7 @@ func TestRuntimePollerWiresRefresherOntoProvidedDispatcher(t *testing.T) {
 // TestRuntimeDispatcherConfigForSnapshotEmptyActiveStatesDisablesFactory
 // guards a foot-gun: if active_states is empty the refresher would mark
 // every state as inactive and kill workers after the first turn. The
-// factory must return nil so the runner falls back to continueRun.
+// factory must return nil so the runner uses its single-turn fallback.
 func TestRuntimeDispatcherConfigForSnapshotEmptyActiveStatesDisablesFactory(t *testing.T) {
 	d := &RuntimeDispatcher{baseConfig: worker.Config{}}
 	d.SetIssueStateRefresher(&stubRefresher{states: map[string]string{"issue-1": "In Progress"}})
--- a/internal/runner/codex_app_server.go
+++ b/internal/runner/codex_app_server.go
@@ -300,24 +300,15 @@ type appServerClient struct {
 	runtimeEvents       []task.RuntimeEvent
 	runtimeEventSink    func(task.RuntimeEvent)
 	phaseTransitionSink func(from, to task.RunAttemptPhase)
-	// continueRun is the agent-emitted "should we keep going?" signal
-	// from turn/completed notifications (`params.continue`). It only
-	// gates the legacy path: when refreshIssueState is wired, SPEC §16.5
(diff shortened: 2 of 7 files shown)


zjlgdx:
@codex review

Issue #1144, opened 2026-07-24
feat(tracker): adopt adapter-owned tracker.provider config


hey-yvan:
## Context

The current workflow schema prescribes one flat cross-provider tracker shape: `api_key`, `endpoint`, `team_key`, `project_slug`, and `pagination_max_pages`. Linear, Gitea, GitHub, doctor, tools, env filtering, and print-config all reach into those shared fields.

Symphony 0.0.2 replaces that contract with `tracker.provider`, an adapter-owned object whose unknown keys are preserved and whose defaults, scope, credentials, `$VAR` support, validation errors, and secret environment names belong to the selected adapter. This is a mandatory config-boundary change, not a request to add upstream's new Jira/Asana/GitLab adapters.

Tracked by #1137.

## Required behavior

- Add `tracker.provider` as a default-empty object that preserves unknown YAML keys without core reinterpretation.
- Move every supported Linear, Gitea, and GitHub endpoint/scope/auth/pagination setting behind adapter-specific profile parsing and validation.
- Remove the old flat provider fields and update all in-repo workflow examples/consumers atomically; this pre-release repo must not accept both old and new names.
- Keep generic `tracker.kind`, `required_labels`, `active_states`, and `terminal_states` in core config; retain documented aiops extensions only where they are genuinely scheduler-owned.
- Resolve `$VAR` only for adapter-documented provider keys, treat an empty resolved secret as missing, and retain the exact referenced secret env names for child-process stripping.
- Mask every adapter-declared secret in `--print-config` without hiding non-secret provider scope/endpoint fields.
- Preserve each current adapter's effective endpoint/scope/token fallback behavior unless the new profile explicitly replaces it and the examples/docs change in the same PR.

## Negative constraints

- No compatibility aliases, dual emission, provider-field copies, or generic `api_key` accessor retained after the cutover.
- Core workflow/orchestrator code must not prescribe one credential or project schema across providers.
- Do not add Jira, Asana, GitLab, SSH workspace, tracker writes, or generic CRUD.
- Do not move provider-native tool execution in this issue; tool ownership/session context is a separate child.
- Do not add adapter semantic validation to per-tick orchestrator logic.
\n\n## Acceptance criteria

- [ ] `tracker.provider` round-trips unknown non-secret keys unchanged through load/reload/print-config.
- [ ] Linear, Gitea, and GitHub each have one adapter-owned effective config path with documented defaults and stable validation errors.
[...]

PR #1172, opened 2026-08-11
fix(workflow): reject semantically invalid snapshots before scheduling

### LINKED ISSUE #1145: fix(workflow): reject semantically invalid snapshots before scheduling
## Context

The workflow runtime already preserves its last-good snapshot when typed loading or an existing validator fails. The admission set is incomplete: a selected tracker can still reach runtime without its required provider secret, and a whitespace-only `codex.command` is accepted then silently replaced by a runtime default. On reload, either semantic-invalid snapshot can replace the previous good configuration before a later preflight fails.

Upstream `d476215c23f33ff27d222ea2d2c44e631de4a922` makes semantic validation part of workflow-store admission: invalid cold start fails before orchestration side effects; invalid reload retains the complete previous typed snapshot and prompt.

Tracked by #1137. This issue is intentionally blocked by #1144 so validation targets the final `tracker.provider` contract rather than adding a short-lived core check for the removed flat `tracker.api_key` field.

## Required behavior

- Define one snapshot-admission path used by initial load and reload before the workflow becomes effective.
- Invoke the selected tracker adapter's provider validator there, including required secret/scope checks after documented defaults and `$VAR` resolution.
- Reject a whitespace-only explicitly configured `codex.command` (and the equivalent Claude command if the same invariant applies) instead of silently substituting the default binary.
- On reload failure, keep the exact prior config, prompt, fingerprint, tracker binding, and runner settings; emit the existing reload-failed signal and allow a later valid edit to recover.
- On cold-start failure, return before startup cleanup, tracker polling, workspace hooks, or runner launch.
- Keep per-tick operational preflight as defense in depth without making it the semantic source of truth.

## Negative constraints

- Do not duplicate provider-specific required fields in orchestrator or worker switches.
- Do not partially merge a new prompt with an old config (or vice versa).
- Do not mutate the last-good fingerprint on failure.
- Do not fall back to a default command when the operator explicitly supplied only whitespace.
- No compatibility keys or new runtime phase.

## Acceptance criteria

- [ ] Valid snapshot → missing selected-provider secret on disk → reload error, unchanged effective snapshot/fingerprint/binding, and no reload-success event.
- [ ] Valid snapshot → whitespace-only explicit command → the same last-good behavior.
- [ ] Fixing the file after either failure successfully installs the new complete snapshot.
- [ ] An equivalent invalid cold start performs zero cleanup/poll/hook/runner side effects.
- [ ] Existing typed-invalid YAML and numeric-field last-good tests remain green.
- [ ] Dropping adapter admission or command validation makes the real reload wiring regression fail.
- [ ] Targeted runtime/worker tests and all repository gates pass.

## Dependencies

Blocked by #1144.

zjlgdx:
Closes #1145

Head: `89cc9b32080bed19c9dfee96fdcf6a0ea7140be8`
Base: `d3cdd23cff47cb00559b034670e2c37a7604c99b`

## Summary
- make workflow load the single complete-snapshot admission seam shared by cold start and hot reload
- bind #1144's selected-adapter provider admission into that named seam after defaults and `$VAR` resolution
- reject explicit blank/whitespace `codex.command` and `claude.command` before a complete snapshot can become effective
- preserve the complete last-good config, prompt, fingerprint, tracker profile binding, and runner settings on failure, then recover on the next valid edit
- close D44 cumulatively across #1144/#1145 while retaining per-tick preflight as defense in depth

## Acceptance ledger
- [x] Missing selected-provider secret rejects reload with no reload-success event and an exact last-good snapshot/binding/fingerprint
- [x] Explicit whitespace-only Codex and Claude commands, including resolved `$VAR` values, have the same last-good behavior
- [x] A later valid edit installs one complete new snapshot and tracker binding
- [x] Equivalent invalid cold starts return before the first tracker request; cleanup/dispatch/hooks/runners cannot follow
- [x] Omitted runner commands and non-command-only sections retain the documented defaults
- [x] Existing affected-package typed-invalid and numeric-field last-good coverage remains green
- [x] Removing adapter admission or command validation makes the real reload/cold-start wiring tests fail

The pinned local SPEC controls: §6.1 orders defaults/`$VAR` before typed validation; §6.2 requires invalid reloads to retain last-good; §6.3 requires selected-adapter provider admission and non-empty commands before startup/dispatch. The audited Elixir reference installs only a fully parsed and semantically validated state in [`workflow_store.ex`](https://github.com/openai/symphony/blob/653f8b3cc476db03420479ba6f95b2ed7281c401/elixir/lib/symphony_elixir/workflow_store.ex#L155-L166), delegates provider validation through [`config.ex`](https://github.com/openai/symphony/blob/653f8b3cc476db03420479ba6f95b2ed7281c401/elixir/lib/symphony_elixir/config.ex#L105-L114), rejects whitespace-only Codex commands in [`config/schema.ex`](https://github.com/openai/symphony/blob/653f8b3cc476db03420479ba6f95b2ed7281c401/elixir/lib/symphony_elixir/config/schema.ex#L197-L211), and reaches cleanup/tick only after admitted settings in [`orchestrator.ex`](https://github.com/openai/symphony/blob/653f8b3cc476db03420479ba6f95b2ed7281c401/elixir/lib/symphony_elixir/orchestrator.ex#L54-L79). Upstream semantic-admission fix: [`d476215c`](https://github.com/openai/symphony/commit/d476215c23f33ff27d222ea2d2c44e631de4a922).

## Testing
- [x] `go vet ./...`
- [x] `gofmt -l` clean + blocking golangci gate (`0 issues`)

Additional committed-head validation:
- `go test -race ./internal/workflow ./internal/orchestrator ./cmd/worker -count=1`
- `git diff --exit-code HEAD`
- Independent final diff-only review: `LGTM` after addressing its actionable findings (explicit seam placement, env-resolved whitespace coverage, alias-proof last-good assertions, default-command positive guard, and removal of a non-discriminating hook marker).

## Mutation evidence
- On final committed head, removed `admitTrackerProvider(path, cfg)` from `admitSnapshot`. `TestWorkflowRuntimeSemanticAdmissionKeepsCompleteLastGoodSnapshot` accepted the missing secret and `TestRunRejectsSemanticInvalidColdStartBeforeTrackerSideEffects` reached runtime; exit 1.
- On final committed head, removed both non-blank command checks. Loader (literal and env-resolved), real reload, and real cold-start tests failed for Codex and Claude; exit 1.
- Restored each mutation with `git checkout HEAD -- <exact-file>`; `git diff --exit-code HEAD` passed after both.

## Risks / deferrals
- Provider behavior itself landed in dependency #1144; this PR owns the explicit complete-snapshot seam, blank-command admission, and wiring regressions that keep #1144's adapter boundary effective across cold start/reload.
- No provider-specific fields moved into core/worker/orchestrator, no compatibility aliases, no partial snapshot merge, no failed-fingerprint publication, and no new runtime phase.
- The existing per-tick preflight remains unchanged as a defensive check for manually constructed configs; it is not the semantic source of truth.
- Prompt-only/default workflow behavior and reload-failure dedupe are preserved.
- No blocker.


zjlgdx:
@codex review

